PR #325, opened 2014-12-22
fix(po.English): add english po file

vomikan:


guanlisheng:
So how to deal with this pull request?

stef145g:
I had not looked at this in depth before, but I agree it should be added. I assumed it was already.
Without adding **english.mo** to the distribution will cause the language request dialog to appear every time MMEX starts. I have now merged this.

PR #334, opened 2015-01-01
fix(feat.request.353): remember column width

stef145g:


gabriele-v:
Hi Stefan, have you tested that works fine without event? GetColumnWidth() function doesn't work correctly inside event under Windows, maybe outside event it works fine?

guanlisheng:
Better to leave this kind of code in the checking panel instead of the Model

stef145g:
I have now tested it on win7 and win8.1, and fixed version works on all conditions.
In non patched version, the columns sometimes revert to previous setting, causing the problem.

As for using the model, it keeps the parameters centralised, and the model will be used in any case because it needs to access the database.

guanlisheng:
Considering the case if we want to store other panels' column width settings.

stef145g:
Appears that this problem also occurs in **Repeating Transactions**, **Assets** and **Stocks**
Condition:

1. Open panel, resizes a column using auto size
2. open another panel
3. return to original panel and check size of resized column.

stef145g:
@guanlisheng appears there is another Travis problem

guanlisheng:
Don't you get the reason why we can't put such code in Model ? Basically, it's the logic of the panels.

stef145g:
I have relocated the positioning of the parameter setting to the panel, which calls Model_Setting.

From a standards viewpoint, the Model_Setting class should contain all parameters belonging to the database, since it is responsible for maintaining the database. It could also be argued, as you have, that the parameter is a panel responsibility, since it only belongs to the panel.

Trying to match the database parameters to the Model Class, how do we determine what should be stored in the database? and how do we determine who is storing what in the database?

guanlisheng:
I just saw the changes and restarted the CI building.
There is no a strict standard. 
In general, a proper approach will make the code easy readable, reasonable and maintainable. to this particular case, one getter/setter interface from Model is enough since most of the logic is inside the panel.

PR #346, opened 2015-01-08
feat(Repeating.Transactions): replace next occurance with payment and due dates

stef145g:
On existing databases, all Repeating Transactions need to be edited to reset the due date.

### DIFF
--- a/doc/version.txt
+++ b/doc/version.txt
@@ -7,6 +7,8 @@ Version History of Money Manager Ex
 - New: Stock History Monitoring
 - New: My Usage Reports
 - New: MMEX News display on Home page
+- Change: Repeating transaction - 'Next Occur Date'
+          redefined as 'Payment Date' and 'Due Date'
 - Fix: Many bug fixes
 
 1.1.2 (November 16, 2014)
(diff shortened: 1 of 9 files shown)


GitCop:

 * Commit: e8352b21c621e43099654e345577cbdf16a306a0
  * Your commit message body contains a line that is longer than 72 characters

---

GitCop:

 * Commit: e8352b21c621e43099654e345577cbdf16a306a0
  * Your commit message body contains a line that is longer than 72 characters

---

guanlisheng:
seems we need a manual data migration for end user ?

stef145g:
Yes to some degree.
The home page will display all pending transactions as being overdue.

**Requirement:** All repeating transactions should be opened in edit mode, and saved. If nothing is changed, the due date will be automatically set to the payment date. This is because the Due date cannot be set to a date before the payment date. This will correct the home page display.

The data change is backward compatible.

gabriele-v:
We should make a DB patch at start-up to fix it. 
We will have lots of users that notice this like a bug (and they are right..)

stef145g:
Agreed.
I have created a patch to fix this, but will also upgrade the database version.
https://github.com/moneymanagerex/moneymanagerex/pull/347

The database version is tested and any database at version 2, will have the fix applied, and updated to version 3. Subsequent actions will not affect the database.

Issue #336, opened 2015-01-04
Support buying & selling for stock account


guanlisheng:

guanlisheng:
@vomikan , would you mind to own this one since i heard you are using stock often.

stef145g:
I think I have worked out a way of achieving this by introducing a new table: **TRANSFER_TRANS_V1**

This table would tie the CHECKINGACCOUNT_V1 table to any other account type. This would extend the transfer functionality of the CHECKINGACCOUNT_V1 table, allowing transfer transaction to other account types such as stocks or assets. This would also allow purchase or sale from these account types by transferring money from existing accounts.

This table could also maintain a link to the currency, to allow a transaction to change its currency.

I am still working out the details before I put it into practice.

guanlisheng:
why a brand new table here ?

stef145g:
Continuing from issues:
**Support transfer operation for stock account** #326, 
**Assets should be closable or have an end date** #361, and 
**feat(Stocks.And.Assets): expand functionality** #367 
---------------------
As stated, Assets and Stocks require similar behaviour, even though they appear to be different.
I will explain in next comment. This is the reason fro closing the other issues.

stef145g:
**Assets**
Assets are created in the asset register. We can think if each record of an asset register as an account. As an account, we can buy, or sell this item, or even generate income from this item or close it.
**In order to do this we need to maintain transactions in a transaction table.**

**Stocks**
Stocks are generated as an Account. In each Stock account we add Shares to maintain a Stocks portfolio. Similar to assets, we can think of these shares as Share accounts belonging to a Stock account, where we buy **shares**, sell **shares** or generate income from **shares** in the form of dividends. We can even close one share account within the stock account.
**In order to do this we need to maintain transactions in a transaction table.**

From both Stocks and Assets we come to the same conclusion.  **We need to maintain transactions in a transaction table**. This is why all the requests are similar, and can all be achieved with one solution.

We already have a transaction table which we call "CheckingAccount_V1". To maintain backward compatibility in the database, we require minimum disruption to the existing tables. The simplest and clean way is the introduction of a new table, as suggested in the feature (Stocks And Assets): expand functionality.

This table unites transactions with existing accounts, and allows transfers or straight transactions to occur for stocks, assets or anything else in the future.

stef145g:
If this proposed solution is unacceptable, then someone else can put forward an acceptable solution and continue to implement all the proposed requests.

[...]

PR #367, opened 2015-01-23
feat(Stocks.And.Assets): expand functionality  

stef145g:
Allow Stocks and Assets to use the Checking Account transaction table for buying and selling
https://github.com/moneymanagerex/moneymanagerex/issues/336#issuecomment-70448615

This is only added to tests to test out functionality and feasibility at this stage.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -14,8 +14,7 @@
 *.html text
 
 # Declare files that will always have LF line endings on checkout.
-sql/tables_v1.sql text eol=lf
-*.sh              text eol=lf
+*.sh      text eol=lf
 
 # Declare files that will always have CRLF line endings on checkout.
 *.bat     text eol=crlf
--- a/build/msw_tests-vc-2013e/mmex_tests_mmex_tests.vcxproj
+++ b/build/msw_tests-vc-2013e/mmex_tests_mmex_tests.vcxproj
@@ -279,6 +279,7 @@
     <ClCompile Include="..\..\src\model\Model_Stock.cpp" />
     <ClCompile Include="..\..\src\model\Model_StockHistory.cpp" />
     <ClCompile Include="..\..\src\model\Model_Subcategory.cpp" />
+    <ClCompile Include="..\..\src\model\Model_TransferTrans.cpp" />
     <ClCompile Include="..\..\src\model\Model_Usage.cpp" />
     <ClCompile Include="..\..\src\aboutdialog.cpp" />
     <ClCompile Include="..\..\src\accountdialog.cpp" />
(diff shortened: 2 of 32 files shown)


guanlisheng:
Would we have a brief talk about this approach?

stef145g:
Adding external transfers to the **CHECKINGACCOUNT_V1** table is a problem.
The table contains the fields **ACCOUNTID** , **TOACCOUNTID** and a **TRANSCODE**
When **TRANSCODE** == Withdrawal or Deposit, the **TOACCOUNTID** is set to -1
When **TRANSCODE** == Transfer the **TOACCOUNTID** is set to ACCOUNTID of the **to account**.

Transactions using Account --> Transaction  for Withdrawals and deposits is good.
Transactions using Account --> Transaction  --> Account  for Transfers is ok but can be problematic.
Identifying another account type from here is **NOT** ok

Problem occurs in checking account, that we cannot distinguish between an account, an asset or a stock. All transfers are automatically assumed as another account in ACCOUNTLIST_V1

**Solution:**
Provide the new table **TRANSFERTRANS_V1**
This provides a link to the **CHECKINGACCOUNT_V1** table, a link to another table and a table type
Now if we want to link an Asset to an account, for a purchase or a sale, this now becomes easy.
The new table can link the asset to a transaction or not.
- As a transaction, the **CHECKINGACCOUNT_V1** TRANSCODE displays the transaction as a deposit or withdrawal, with the **TOACCOUNTID** pointing to a record in the **TRANSFERTRANS_V1** table.

The new table allows the linking of an Asset or a Stock to one or many transactions that are associated with an account. Any transaction can be easily identified from the Checking Account table. As an added bonus, we can now have an Asset or a Stock in a different currency to the account that the money is being transferred to or from.

In writing this, I have just thought of adding an extra an extra field to the **TRANSFERTRANS_V1**
 table to determine if this is to be accounted as a deposit / withdrawal or not in the checking account.

guanlisheng:
@moneymanagerex/desktop-mmex-team as FYI.
Thinking that we already have stock account, the simple approach might be to add asset account type.

stef145g:
The addition of an Asset Account type would not solve the initial problem which is the way transfers have been originally implemented. This was fine in a simple system, transferring money from account A to Account B. This approach was and is still a problem when it comes to certain report types. Also the introduction of an Asset Account type would introduce other major problems.

Stock accounts represent Stock Portfolios which hold Share accounts. Share accounts are not well managed when it comes to the buying and selling of shares, or receiving dividends from shares. Setting up this structure will correct this issue.

vomikan:
I dont like this idea.
Better to add one more account type ASSET.
In that case 2 feature request will sutisfied.
1 other currency for assets
2 transfers to assets

N


суббота, 24 января 2015 г. пользователь Stefano написал:

stef145g:
If this concept is unacceptable, this pull request can be closed.

omalleypat:
I think being able to transfer from a checking account to a stock account
would be a great feature. For example if I make a contribution to a
retirement account but don't purchase stocks right away, that money is
"lost" to mmex as it is set up now. This has been a request from several
users for many years.

stef145g:
@omalleypat, Thankyou for the positive response.
At this point, I will close this pull request, because I will readjust this table and add the new field.

If I have more positive responses, I will resubmit a modified pull request.

PR #371, opened 2015-01-27
fix(Bug.593): category totals fix

vomikan:


guanlisheng:
Seems we use some magic here and would we remove that ?

vomikan:
Mark it as TODO: I'll think how to avoid magic in future.
I wish to have this fix for 1.2.0.

PR #372, opened 2015-01-27
revert(Bug): After last commit, no longer calls function if id<0

dwrj87:


guanlisheng:
Thanks for this pull request

dwrj87:
@guanlisheng the least I could do considering it was me that broke it....

guanlisheng:
Would you mind to join the team as a regular contributor?

Issue #387, opened 2015-02-15
QIF import ability to chose account number instead of account name to match account


loki36:
it could bee good to have ability to chose account number instead of account name to match account when you import data from qif file.

this beacause my bank use the BankID + account ID as account name, but i prefer display an other name on mmex.

guanlisheng:
Would we have another approach for this requirement.

PR #393, opened 2015-02-19
feat(Import.qif.account): QIF import 

loki36:
ability to chose account number instead of account name to match account #387

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -258,8 +258,6 @@ install: install_mmex
 	(cd $(srcdir)/doc ; $(INSTALL_DATA)  *.txt $(DESTDIR)$(datadir)/doc/mmex)
 	$(INSTALL_DIR) $(DESTDIR)$(helppath)
 	(cd $(srcdir)/doc/help ; $(INSTALL_DATA)  *.html *.png $(DESTDIR)$(helppath))
-	$(INSTALL_DIR) $(DESTDIR)$(helppath)/german
-	(cd $(srcdir)/doc/help/german ; $(INSTALL_DATA)  * $(DESTDIR)$(helppath)/german)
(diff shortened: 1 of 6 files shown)


GitCop:

 * Commit: f314ff2056a38d8bb449ae61eddfdb7278ec398c
  * Your subject line is longer than 72 characters

---

GitCop:

 * Commit: f314ff2056a38d8bb449ae61eddfdb7278ec398c
  * Your subject line is longer than 72 characters

---

loki36:
we could create a unique index on table based on accountnum or based on  heldat(accountbank) + accountnum ?

guanlisheng:
Technically, we could.
I don't think it's correct base on existing usage about account number.
So do we plan to use it to identify an account, treat it same as account name and id?

loki36:
an other approach would be to have the "accoutname" and the "accountdysplayname" and set accountdysplayname equal to accountname if it not set . But we need after that to change everywhere in gui the accountname by accountdysplayname.

vomikan:
I hope will have time to discover this tonight.

loki36:
not really sure to understand.

vomikan:
@loki36 Could you provide QIF file with account number?
There is only "Number of the check" in the QIF specification.

loki36:
yes, it's because the account number is in the field NAccount Name

an qif example file : http://owncloud.far.bongrand.fr/public.php?service=files&t=291345253e6918a6658cd882d9049407
link will expire by the end of march

vomikan:
I see. Other words, in mmex U prefer to have aliace for account names. At the same time a bank provide account number for QIF file.
In that case U have provided nice feature.

Issue #422, opened 2015-03-24
Reccuring transactions layout grid


AlessandroLazzari:
In the window of reccuring transactions is not saved the layout of columns and their width. When thes user close this windows and re-open the layout will be reset

Size of columns that were modified manualy:

![image](https://cloud.githubusercontent.com/assets/7428710/6800893/3a8c8d8e-d223-11e4-956b-4dbddc4d3207.png)

After closing the window and re-opened

![image](https://cloud.githubusercontent.com/assets/7428710/6800914/68113e30-d223-11e4-9357-670d309ffc87.png)

Works fine for normal transactions

Version: 1.2.0-Beta.2 Portable

guanlisheng:
Thanks @AlessandroLazzari for the feedback.
@stef145g  , would you mind to take a look here?

stef145g:
Issued pull request could have cured the problem, but was closed due to it being found to be unacceptable. Someone else can investigate further.

stef145g:
I discovered a merge conflict with the pull request as a result of other commits to master. I have now fixed the resulting conflicts and successfully merged **stef145g/master** with **moneymanagerex/master**. This now resides in my repo.

vomikan:
Closed?

stef145g:
This issue is now fixed in both master and v1.2.x branch

PR #425, opened 2015-03-27
fix(GitHub.Issue.422): Recurring transactions layout grid

stef145g:
Ref: https://github.com/moneymanagerex/moneymanagerex/issues/422
This fix has been provided, but on my system the SetColumnWidthSetting(...) command in the destructor does not appear to get executed.

Please test to see if this corrects the problem.

### DIFF
--- a/src/mmpanelbase.cpp
+++ b/src/mmpanelbase.cpp
@@ -199,6 +199,16 @@ void mmListCtrl::OnHeaderColumn(wxCommandEvent& event)
     }
 }
 
+int mmListCtrl::GetColumnWidthSetting(const int& column_number, int default_size)
+{
+    return Model_Setting::instance().GetIntSetting(wxString::Format(m_col_width, column_number), default_size);
+}
+
+void mmListCtrl::SetColumnWidthSetting(const int& column_number, int column_width)
+{
+    Model_Setting::instance().Set(wxString::Format(m_col_width, column_number), column_width);
+}
+
 mmPanelBase::mmPanelBase()
 {
 }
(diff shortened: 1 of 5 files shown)


src/mmpanelbase.cpp
```diff
@@ -199,6 +199,16 @@ void mmListCtrl::OnHeaderColumn(wxCommandEvent& event)
     }
 }
 
+int mmListCtrl::GetColumnWidthSetting(const int& column_number, int default_size)
```
guanlisheng: would we just pass the value here ?

stef145g:
Your question is invalid.
The 2nd parameter is required for the underlying code call to the stored database setting. The setting has 2 parameters. One is the key, and the other is the expected default.
In the .h file, where the function interface is, the 2nd parameter becomes optional with a pre-set default value for this particular case.

stef145g:
I don't quite understand why in some instances we are using
 **const& int value** and other instances we use **int value** to do the same thing.

From my understanding, passing as value creates a copy of the value within the function, and can be changed with no affect to the caller, where as passing as a reference, passes the address and therefore any change is reflected back to the caller. Declaring it **const**, should prevent changes and possible overheads thus increasing speed. Yes?

guanlisheng:
I have sent a dedicate email talking about the issue.
For other instances w/ the similar issue, I think only auto generated code is acceptable while we should not write the same code manually.

LiSheng

stef145g:
I don't quite understand your response, but whatever...
Is this an acceptable and workable solution or what?

guanlisheng:
Not acceptable.

PR #437, opened 2015-04-02
refactor(Code): cleanup

vomikan:


src/mmframereport.cpp
```diff
@@ -91,7 +91,7 @@ class mmGeneralGroupReport : public mmPrintableBase
             contents += report.to_row_t();
 
         mm_html_template report(group_report_template);
-        report(L"REPORTNAME") = this->local_title() + " For " + this->m_group_name;
+        report(L"REPORTNAME") = wxString::Format(_("%s For %s"), this->local_title(), this->m_group_name);
```
guanlisheng: why we need this?

guanlisheng:
would we just change one category things in one topic

vomikan:
I don't undestand your policy - closed a lot of codes if one questionable
issue is present.
May be betted to discass then fix it with additional commit. In other case
I lost a lot of time.

2015-04-03 5:08 GMT+03:00 Lisheng <notifications@github.com>:

guanlisheng:
when you say it's cleanup, it should be cleanup only. 
To this particular pull request,  there are many other changes and this is the root reason i can not merge it.

PR #443, opened 2015-04-07
fix(ColumnWidthSettings): saving settings in list destructor

stef145g:
This is common to all panels

### DIFF
--- a/src/mmpanelbase.cpp
+++ b/src/mmpanelbase.cpp
@@ -47,7 +47,11 @@ mmListCtrl::~mmListCtrl()
       Save the column widths of the list control. This will ensure that the
       column widths get set incase the OnItemResize does not work on some systems.
     */
+#ifdef __WXMSW__ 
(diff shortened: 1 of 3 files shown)


src/mmpanelbase.cpp
```diff
@@ -43,6 +43,18 @@ mmListCtrl::~mmListCtrl()
 {
     if (attr1_) delete attr1_;
     if (attr2_) delete attr2_;
+    /*
```
vomikan: Here should be m_col_width validation. If the variable isemptyor not like *%d* skip save settings.

stef145g:
@vomikan Your comment is invalid.
The variable **m_colCount**  is derived from the protected variables in the base class of **mmListCtrl** being **wxListCtrl**. This determines the amount of columns to be saved. Nothing to do with column width validations.

Travis is reporting an undeclared variable. Works in Windows so the only explanation is that the Mac port of wxWidgets does not support  the direct variable approach but uses the method:  **GetColumnCount()** instead of the direct variable access.

PR #456, opened 2015-04-16
fix(aboutdialog): Show privacy tab at startup

gabriele-v:
@loki36 please check where a function\code is used befor change it (#454).

Here we need to show privacy at startup to inform users that we will collect usage statistics.

Thanks


dwrj87:
@gabriele-v my apologies, that was me, not @loki36

gabriele-v:
Sorry for the nick error, anyway no problem with this, it's just for your information

PR #458, opened 2015-04-17
fix(GRM): exception constraint failed for the same report name

vomikan:
Also hide reports in main navigation tree if report group name starts with dot.

### DIFF
--- a/src/general_report_manager.cpp
+++ b/src/general_report_manager.cpp
@@ -34,8 +34,8 @@
 #include <wx/sstream.h>
 #include <wx/zipstrm.h>
 
-static const wxString SAMPLE_ASSETS_LUA = R"(
-local total_balance = 0
+static const wxString SAMPLE_ASSETS_LUA = 
+R"(local total_balance = 0
 function handle_record(record)
     total_balance = total_balance + record:get('VALUE');
 end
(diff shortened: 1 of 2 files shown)


guanlisheng:
I don't know why we need this hiding functionality.

vomikan:
1. My goal is to have possibility to run some grm reports from another
report.
2. Some temporary reports better to hide from every day usage.

воскресенье, 19 апреля 2015 г. пользователь Lisheng написал:

guanlisheng:
Would you explain more on the both purpose?

vomikan:
For example, I'll create '.homepage' group with some reports included. All
reports of this group will provide a small peace of info (just table, for
example) . Then these reports will used for build home page.
 20 апр. 2015 г. 1:54 пользователь "Lisheng" <notifications@github.com>
написал:

guanlisheng:
I recommend to use separate table to store the pieces

vomikan:
Your suggestion is yo complex for now.
My suggestion is simple and ma be removed if useless at any time without
any problems.

понедельник, 20 апреля 2015 г. пользователь Lisheng написал:

guanlisheng:
I don't see the complications while your solution make it complex. 

We can't remove your code once there are some pieces report there.

For this standalone table, I would like to create new one instead of dirty reuse 

LiSheng

vomikan:
OK 
Hide group is only for hide not complited reports anly.

guanlisheng:
So let's close this PR .

For GRM, I think we should focus on the UX, for instance, how to build a report easily 

LiSheng

PR #467, opened 2015-04-24
chore(AccountDialog): cleanup, format, simplification

vomikan:


vomikan:
I've merged this because nobody care about it.

PR #472, opened 2015-05-11
Bug672

vomikan:
https://sourceforge.net/p/moneymanagerex/bugs/672/


vomikan:
a129eb8 is for 1.2.1 as well

PR #474, opened 2015-05-14
refactor(git): fixed linux files not having lf

dwrj87:
At the moment, the following files will have crlf instead of lf, so file when run, unless line endings are converted manually. This converts the line endings to lf, and hopefully keeps them that way


setup/linux/build_deb.sh
```diff
@@ -24,7 +24,7 @@ PACKAGE_NAME="mmex-$MMEX_VERSION-$ARCHITECTURE"
 cd ../..
 MMEX_DIR=`pwd`
 
-./bootstrap
+./bootstrap.sh
```
vomikan: I am confusing. There is no "bootstrap.sh" file in the repository.

dwrj87: I've renamed bootstrap to sh since it is a sh file and the .gitattributes file will cover it then without adding a new rule

vomikan: cool

2015-05-14 17:23 GMT+03:00 David Jones <notifications@github.com>:

PR #486, opened 2015-05-19
refactor(Icons): resizing

vomikan:


gabriele-v:
Nikolay, toolbar icons are poorly scaled and they are blurred.
Please add the same logic based on HTML factor also on toolbar.

PS: We need to refactor all icons befor make this change in stable version..

PR #498, opened 2015-06-03
refactor(Checking.Transaction): separate function for new transfer

vomikan:
Preparations for the separation of functions for create new transaction as Transfer or Withdrawal/Deposit.


guanlisheng:
Better to provide separate dialog as well

PR #500, opened 2015-06-07
refactor(News): no news on home page

vomikan:
Stage #1, Now any new on Home Page. 
At startup last year news will store in "g_WebsiteNewsList" variable.
Then icon in Navigation tree or in Panel bar may be provided to open news list.


omalleypat:
Just curious - why do we want to remove the news from the home page?

vomikan:
Now there is an icon with news anailable in a toolbar
screenshot 46

PR #505, opened 2015-06-16
refactor(PanelBase): template for html

vomikan:


guanlisheng:
Too aggressive to use html template here

PR #506, opened 2015-06-16
fix(Bug.690): export to html

vomikan:


vomikan:
how to fix?

vomikan:
Need more details here why the PR is incorrect.

Issue #306, opened 2014-12-15
History exchange rate support


guanlisheng:
Both @vomikan and @gabriele-v think this is a cool feature to stick more audiences here.
anyway, would you guys help to compose some design before coding?

guanlisheng:
Any update?

gabriele-v:
Working on it in next weeks

vomikan:
https://github.com/moneymanagerex/moneymanagerex/issues/647

PR #508, opened 2015-06-20
feat(currency): Currency history interface

gabriele-v:
First step to complete #306 

UI is complete, better to merge it now because all logic will require more time maybe creating conflicts difficoult to solve.


src/currencydialog.cpp
```diff
@@ -231,59 +206,65 @@ void mmCurrencyDialog::CreateControls()
     wxBoxSizer* itemBoxSizer22 = new wxBoxSizer(wxHORIZONTAL);
     itemBoxSizer2->Add(itemBoxSizer22, wxSizerFlags(g_flags).Centre());
 
-    wxButton* itemButton24 = new wxButton( this, wxID_REFRESH, _("&Update"));
+    wxButton* itemButton24 = new wxButton(this, wxID_OK, _("&OK"));
```
vomikan: _("&OK") should be _("&OK ")

gabriele-v: Fixed in my trunk (cannot commit because unstable)

vomikan:
Don't forget put into changed files
Copyright (C) 2015 Gabriele-V

vomikan:
Suggesions:
1. After "Online Update" the currency history list should be refreshed.
2. Add "Delete history for unused currencies" function.

PR #512, opened 2015-07-02
chore(File.Path): more const ref

vomikan:


guanlisheng:
agreed to add/fix something really useful

vomikan:
Your message is too short to be useful. Could you please comment exectly incorrect line of the code but not all PR.

vomikan:
Is it OK now?

PR #514, opened 2015-07-02
fix(bug.sf707): properly handle from field for transfers

omalleypat:


vomikan:
I've already cherry-picked it from master.

PR #524, opened 2015-07-12
feat (ui.incexpenses): Add chart to current/last year 

mek-x:
Please look at my proposition of additional chart to "Income vs Expenses -> Current/Last Year" report


guanlisheng:
hi @mek-x , i would like we can build this feature with HTML5 instead of the CPP. thought?

mek-x:
hi @guanlisheng ! Yes, that's a good idea. You mean something similar to the homepage from #511 ?

guanlisheng:
similar

vomikan:
It should be used other style of coding: html template and js code.

PR #541, opened 2015-08-05
refactor(UI): png icons for toolbar

vomikan:


dwrj87:
It looks *much* better with these

vomikan:
Thank you for your fidback. I will continue to change xpm icons to png.

PR #555, opened 2015-08-28
feat(build): fix warnings on VS2015

gabriele-v:


src/import_export/export.cpp
```diff
@@ -72,10 +72,10 @@ const wxString mmExportTransaction::getTransactionQIF(const Model_Checking::Full
 
     for (const auto &split_entry : full_tran.m_splits)
     {
-        double value = split_entry.SPLITTRANSAMOUNT;
+        double valueSplit = split_entry.SPLITTRANSAMOUNT;
         if (Model_Checking::type(full_tran) == Model_Checking::WITHDRAWAL)
-            value = -value;
-        const wxString split_amount = wxNumberFormatter::ToString(value, 2, style);
+			valueSplit = -valueSplit;
```
vomikan: Please use break not tab

gabriele-v: Fixed, every time that I update VS I don't remember to re-set this setting

PR #581, opened 2015-10-27
feat(CurrencyHistory): converted IncExpenses

gabriele-v:
I've converted IncExpenses with new CurrencyHistory.
Please test and, if working fine, I'll update also other reports.

We should think about a way to obtain nearest currency value if not present for a specific date.
I've tried with the commented code in Model_CurrencyHistory.cpp but it's quite slow so I preferred to use BASECONVRATE for now.

### DIFF
--- a/src/maincurrencydialog.cpp
+++ b/src/maincurrencydialog.cpp
@@ -116,7 +116,7 @@ void mmMainCurrencyDialog::fillControls()
         data.push_back(wxVariant(baseCurrencyID == currencyID));
         data.push_back(wxVariant(currency.CURRENCY_SYMBOL));
         data.push_back(wxVariant(currency.CURRENCYNAME));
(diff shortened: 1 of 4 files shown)


gabriele-v:
Where are them? I've tested all that ones and they work fine
incomevsexpenses

If you mean HomePage, it hasn't been changed yet.

vomikan:
I mean .find() may not found record if the history has weekly or montly values
Model_CurrencyHistory::Data_Set Data = Model_CurrencyHistory::instance().find(Model_CurrencyHistory::CURRENCYID(currencyID),Model_CurrencyHistory::CURRDATE(Date));

gabriele-v:
With currency history you cannot download monthly or weekly. This is possible only with stock history. 

Anyway the problem that you have noticed is what I was saying in pull request open: we should think about a way to get the nearest exchange rate if day doesn't have it. I cannot find any way to do it with good performance..

PR #583, opened 2015-10-29
feat(Import.CSV): Ignore lines from the start and/or end of csv file

gitys:
- Add row range
  - Add "ignore from start" and "ignore from end" num picker.
  - Rows not in import range will be grayed out in list.
  - Add row to ignore to template.
- Change "search" button to "browse"
- Remove limit of 10 line on import in preview table.
- Remove bold font
- Set max rows in preview to 20.
- Modified resizer proportions to prefer enlarging the list.
- Make log file read only.

PR #590, opened 2015-11-09
fix(Import.CSV): Fixed problems caused by #583

dwrj87:
A problem with pull request #583 causes MMEX not to compile for me. First it complains about missing header, and then that SetMax is not a member.

This fixes that


crayxt:
+1

PR #606, opened 2015-12-10
fix(IE): fix IE compatibility when build with VS2015

gabriele-v:
NB: should be checked if works with VS2013 and a clean Windows machine

First version of dbCheck to check DB consistency


gabriele-v:
Yes, you are right, I haven't noticed tooltip problem.

I don't know how to fix, but this is the only way to make it working (at least a bit) when built with VS2015

PR #614, opened 2015-12-29
feat(Import.XML): import XML files in excel format

gitys:
Can now import transactions for a file saved by excel in XML format (Save as->Other formats-> XML Spreadsheet 2003).

* I debated between creating a new dialog or adding it to the CSV dialog. I chose to use the CSV dialog and simply cut out the parser sections. This actually removes duplicate code that was used for both preview and import.
* Only import is currently supported, not export.
* Data is imported from first spreadsheet in workbook.
* The concept of printing a few lines of the source file to an edit box to allow the user to choose fields doesn't work with XML. Instead of having the fields list determine the number of columns, the parser determines the number of columns. Any column that does not have a corresponding field we be set to "Don't care". This will now be the case for CSV as well.
* Templates for CSV and XML are separate.
* Only field value are extracted from XML. It can be considered in the future to parse field formatting as well.
* Everything else is same as CSV is today.

### DIFF
--- a/build/msw-vc-2015/mmex_mmex.vcxproj
+++ b/build/msw-vc-2015/mmex_mmex.vcxproj
@@ -266,6 +266,7 @@
   </ItemDefinitionGroup>
   <ItemGroup>
     <ClCompile Include="..\..\src\dbcheck.cpp" />
+    <ClCompile Include="..\..\src\import_export\parsers.cpp" />
     <ClCompile Include="..\..\src\mmOptionAttachmentSettings.cpp" />
     <ClCompile Include="..\..\src\mmOptionBaseSettings.cpp" />
     <ClCompile Include="..\..\src\mmOptionGeneralSettings.cpp" />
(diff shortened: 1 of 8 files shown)


gitys:
I would like to elaborate about two points that Nikolay questioned: export, and sharing the CSV dialog

**Export**
Of course, it would be nice also to have export. I think it is less important because it is necessary to import transaction from different financial sources at minimum effort. If my bank or credit card company gives XMLs I need to import XML. It is unreasonable to open in excel an convert to CSV (this doesn't support unicode either). Export would be nice but I think it is very reasonable to start of with only import, also considering the fact that if I want to export to excel I can still use CSV format.

**Joint dialog**
I really did debate about this but eventually my conclusion was clear. What I implemented was similar to view-model pattern separating GUI from the file format. I think this came out to be a very reasonable design. If I had separated to a new dialog there would be many duplicate components such as: file selection, templates, field selection, formatting of fields, preview grid, and logic that creates the DB entry including user confirmation. This design also unifies the parsing done for the preview and the parsing done for the actual import.

guanlisheng:
i will merge the pr but we still need more testing @moneymanagerex/desktop-mmex-team

gitys:
I don't seem to have access to this link.
In any case, I will do some more testing of error flows.

PR #663, opened 2016-01-24
feat(CustomFields): Add custom fields feature

gabriele-v:


src/customfielddialog.cpp
```diff
@@ -0,0 +1,311 @@
...
+void mmCustomFieldDialog::OnMove(const wxPoint & RefPos, const wxSize & RefSize)
+{
+    this->Move(wxPoint(RefPos.x + RefSize.GetWidth(), RefPos.y));
+}
+
+void mmCustomFieldDialog::OnSave(const bool OpenStatus)
+{
+    if (m_RefreshRequested)
+    {
+        Destroy();
```
vomikan: Detected memory leaks!
How to reproduce:
Create new transaction by press the '+' ico in tool bar. Then immediatly press the "Close" button. Close the app.

src/model/Model_CustomField.cpp
```diff
@@ -0,0 +1,193 @@
...
+    for (const auto& item : FIELDTYPE_CHOICES) types.Add(item.second);
+    return types;
+}
+
+wxString Model_CustomField::getTooltip(const wxString& Properties)
+{
+    json::Object jsonProperties;
+    std::wstringstream jsonPropertiesStream;
+
+    jsonPropertiesStream << Properties.ToStdWstring();
```
vomikan: To start working with this branch I've thanged it to :
~~~~~
jsonPropertiesStream << (Properties.empty() ? L"{}" : Properties.ToStdWstring());
~~~~~
for any jsonPropertiesStream in this file.
In other case the Error message appears: Unexpected end o ftocken string.

gabriele-v: As already said this isn't needed in production. You have used a very alpha implementation that has a different DB structure. If you want to fix excute this query on your DB:
UPDATE CUSTOMFIELD_V1 SET PROPERTIES = '{}' WHERE PROPERTIES = ''

src/customfielddialog.cpp
gabriele-v: Cannot reproduce with latest Stefan fixes: https://github.com/gabriele-v/moneymanagerex

gabriele-v: PS: How to update pull request?

src/model/Model_CustomField.cpp
vomikan: Stefan's path does not help here for me.

stef145g:
I have realigned my custom_fields branch to @gabriele-v master for testing:
now mmex will crash when I do the following:
1. edit a transaction (or New)
2. open the custom field dialog
3. close the custom dialog
4. Enter the transaction using OK...

vomikan:
When I try to  open the custom field dialog:

 [image: Встроенное изображение 1]

2016-01-25 17:27 GMT+03:00 Stefano <notifications@github.com>:

stef145g:
I cannot see the image as suggested above.

vomikan:
I've attached the image.

2016-01-25 17:48 GMT+03:00 Stefano <notifications@github.com>:

src/model/Model_CustomField.cpp
vomikan: Thank you. The update has fixed the issue.

2016-01-25 15:46 GMT+03:00 Gabriele-V <notifications@github.com>:

src/customfielddialog.cpp
vomikan: Close it first then commit again.

2016-01-25 16:10 GMT+03:00 Gabriele-V <notifications@github.com>:

gabriele-v:
@stef145g Should be fixed now, it's a way that I don't like (hide frame instead destroy) but I haven't found another solution that won't add lot of redundant code to parent windows (transaction now, but other dialogs in future too)..

@vomikan I can't see image too, could you post directly to GitHub?

stef145g:
@gabriele-v Your fix works.
Hiding frame is fine, as it will be destroyed when the transaction dialog is closed, regardless of the number of frames created.

I have discovered more places where it crashes and a fix provided in https://github.com/stef145g/moneymanagerex/tree/custom_fields branch for your evaluation.

vomikan:
@stef145g From my point of view, better to clone  Gabriele's repo then sent PR to it.

stef145g:
pull request created: https://github.com/gabriele-v/moneymanagerex/pull/1

vomikan:
@stef145g & @gabriele-v guys if this PR stable enought - merge!

Issue #622, opened 2015-12-31
New feature: copy (CTRL+C) + paste in Excel in Account view


aquirin:
It would be nice to be able to select several operations in the account view, copy them easily with CTRL+C and paste them in Excel, with the header column being the title of the column.

guanlisheng:
@siena123 help to take a look?
wxTheClipboard from the following link may help here
http://docs.wxwidgets.org/3.0/classwx_clipboard.html

siena123:
That looks like it will work but also need to change the view to multi select instead of single select control / view. I will work on make some changes to support this.

aquirin:
That's great thank you!

siena123:
Pull Request submitted. The account panel showed since of being a multi-select control at one time.

PR #666, opened 2016-01-25
feat(#622): multiline select copy to clipboard

siena123:


siena123:
The account panel showed since of being a multi-select control at one time.

vomikan:
1. Copy/Past single transaction has lost;
2. Amounts transformed into dates

siena123:
1. Copy/Paste single transaction is working for me. If multi select copy is done the prior single select copy in mmex gets cleared so paste is not available until another single select copy is done. 
2. Amounts are not transforming into dates for me. I have pasted data into both a text editor (notepad) and into the OpenOffice spreadsheet application. Both Excel and OpenOffice have options to specify the format of the column when data gets pasted into the program (may require multi-line paste). You may need to specifically specify the format for the column if you still have the problem as it is an issue with the spreadsheet application incorrectly auto-detecting the column format. 

One other item to note is that the format is tab delimited for the copy/paste so as to not need to deal with a text delimiter if using comma delimited format or one of the other commonly used delimiters which could be in the notes field. 

- James

vomikan:
Sorry I've updated the posting after first comit.


   1. The amounts transformed into dates in Excel.
   2. After Copy/Past single transaction into the account the selection
   contains both: original and new transaction.
   3. If, for example, 3 transaction selected then right click on them - as
   result only 2 transaction selected.
   4. After delete single transaction - two transactions selected


2016-01-25 16:25 GMT+03:00 siena123 <notifications@github.com>:

siena123:
1. When you paste the data into Excel it should open the import data wizard and on the 3rd page you have the option to specify the column format for the imported data from general to number which should fix the problem with Excel auto detect format for general transforming the data. Also on the 3rd page you have an advanced button that you can specify the decimal and thousands seperators.

2-4. I have been able to reproduce the problem and will work on a fix.

siena123:
Fixes pushed up for the 3 issues I could reproduce.
The other issue I don't see as being a problem in mmex but a user issue with Excel.

PR #686, opened 2016-01-31
fix(Export.XML): existing file bug and thousand separator bug

gitys:
- Don't try to delete the export file if it didn't exist.
- Export raw numbers with no formatting except precision.
   so that they can be easily parsed by external programs (e.g. Excel)

### DIFF
--- a/src/import_export/univcsvdialog.cpp
+++ b/src/import_export/univcsvdialog.cpp
@@ -880,17 +880,19 @@ void mmUnivCSVDialog::OnExport(wxCommandEvent& /*event*/)
         return mmErrorDialogs::InvalidFile(m_text_ctrl_);
 
     wxFileName out_file(fileName);
-    if (out_file.Exists()) {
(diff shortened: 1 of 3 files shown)


src/import_export/univcsvdialog.cpp
```diff
@@ -932,8 +934,8 @@ void mmUnivCSVDialog::OnExport(wxCommandEvent& /*event*/)
         double value = Model_Checking::balance(pBankTransaction, fromAccountID);
         account_balance += value;
 
-        const wxString& amount = Model_Currency::toString(value, currency);
-        const wxString& amount_abs = Model_Currency::toString(fabs(value), currency);
+        const wxString& amount = Model_Currency::toStringNoFormatting(value, currency);
```
vomikan: May be just 
~~~~~
wxString::Format("%ld", value)
~~~~~

gitys: That is actually what I thought of doing at first but that will not take the precision from "currency".
I can use wxString::Format is you think using the precision is not important.
In any case I would not use "%ld" because I would want decimal point accuracy.

PR #689, opened 2016-02-02
refactor(BuildSystem): Initial work on CMake

dwrj87:
Please don't merge yet. Only the *nix side is working so far. This pull request at the moment is only so I can test Travis out


guanlisheng:
got it @dwrj87

guanlisheng:
and thanks @dwrj87

vomikan:
Personally, I believe that 'po' files may be excluded from build process.

stef145g:
I noticed that in the **cmake** branch,
- The file: **mmhomepagepanel.cpp** was modified. Checking the modification, I noticed that some of the code has been reverted. Not shure how this would have happened. Ref pull request: #687 
- Would it be better to leave the existing build setups in place until the new build structure becomes stable. Then removing the deprecated builds in a follow up commit.

dwrj87:
@stef145g Hmmm, I must have accidentaly overwritten the changes there. I shall fix it shortly. It probably would be a good idea to have them both side by side until it's stable. I shall restore them later as well

@vomikan The current build system converts them to .mo and copies them in to place as well, and I'm not sure what else you could do with them

vomikan:
@dwrj87 I mean nobody needs mo file. I suggest to skip po -> mo conversion on travis-c side.
It will simplify your work and avoid some build process errors in future.

Issue #340, opened 2015-01-08
System language auto detecting


guanlisheng:
The program should always uses the system language (fallback to english if not available).
I do think it will simplify both the program and end user operation. 
@vomikan @gabriele-v @stef145g @siena123 , what do you think ?

vomikan:
How many mmex used as portable?
system language is not a best solution for this. In other ases it's OK.

2015-01-08 8:29 GMT+03:00 Lisheng <notifications@github.com>:

guanlisheng:
What's the difference between portable and non-portable? they are identical.
Generally, any application should follow its system's language setting.

gabriele-v:
I think that problem here is that we force locale to "en" at start-up instead to use the system one. 
I think that this is also the reason why all system messages (for example caps lock on password) are in English instead of proper system language

guanlisheng:
Agree with @gabriele-v . 
For all, i want to push this out in v1.3 as part of the UX improvement & simplification.

guanlisheng:
Any concern to make this change since I don't want to give users too many options

gabriele-v:
I don't think that it should be closed, because system language it's still English, this only change MMEX Language.

vomikan:
Personally I believe that the app language is user choice.
My fix just help to choose the same language as OS language.

What do you mean 'system language it's still English'?

суббота, 30 января 2016 г. пользователь Gabriele-V написал:

gabriele-v:
Yes, I agree on language. 

I refer to this:
    /* Setting Locale causes unexpected problems, so default to English Locale */
    app->getLocale().Init(wxLANGUAGE_ENGLISH);


This will cause to show all wx system messages and errors in English, I don't know what problems comment refers to, but maybe it could be referred to old wx versions, so we should make some tests..

guanlisheng:
hi @moneymanagerex/desktop-mmex-team , how about to roll out this simplification soon?

PR #704, opened 2016-02-12
fix(msvc.build.warning.C4702): unreachable code

stef145g:


src/db/DB_Table_Subcategory_V1.h
```diff
@@ -100,60 +99,6 @@ struct DB_Table_SUBCATEGORY_V1 : public DB_Table
         return true;
     }
 
-    bool ensure_data(wxSQLite3Database* db)
-    {
-        try
-        {
-            db->ExecuteUpdate(wxString::Format("INSERT INTO SUBCATEGORY_V1 VALUES (1, '%s', 1)", wxTRANSLATE("Telephone")));
```
vomikan: Should be
db->ExecuteUpdate(wxString::Format("INSERT INTO SUBCATEGORY_V1 VALUES (1, '%s', 1)", _("Telephone")));

src/db/DB_Table_Subcategory_V1.h
```diff
@@ -100,60 +99,6 @@ struct DB_Table_SUBCATEGORY_V1 : public DB_Table
...
-            db->ExecuteUpdate(wxString::Format("INSERT INTO SUBCATEGORY_V1 VALUES (2, '%s', 1)", wxTRANSLATE("Electricity")));
```
vomikan: May be better to init two integer variables like c=1, s=1 nad use it
 db->ExecuteUpdate(wxString::Format("INSERT INTO SUBCATEGORY_V1 VALUES (++c, '%s', s)", wxTRANSLATE("Electricity")));

util/sqlite2cpp.py
```diff
@@ -178,26 +177,8 @@ def to_string(self, sql = None):
...
     
-        s += '''
-    bool ensure_data(wxSQLite3Database* db)
-    {
-        try
-        {'''
 
-        for r in self._data:
-            s += '''
-            db->ExecuteUpdate(wxString::Format("INSERT INTO %s VALUES (%s)", %s));''' % (self._table, ', '.join(["'%s'" if isinstance(i, unicode) else str(i) for i in r]), ', '.join([ 'wxTRANSLATE("' + i + '")' for i in r if isinstance(i, unicode)]))
```
vomikan: Please remove wxTRANSLATE from this line.

stef145g:
New commit reverts the previous code and adds extra code. Also fixed the translation issue from a previous update

guanlisheng:
i don't think we such dummy code here and am wondering why you guys have so strong option to remove a waring...

i think the translation is incorrect while wxTRANSLATE 
http://docs.wxwidgets.org/3.0/group__group__funcmacro__string.html#gab8862de0f74049708bfea75ae9e14c45

guanlisheng:
_() is not correct here

http://docs.wxwidgets.org/3.0/group__group__funcmacro__string.html#ga8a02b8875a521df57263a9e6f090f2d0

stef145g:
@guanlisheng I have tested the category translations by creating a new database using Italian as my base language.
Using the wxTRANSLATE ("string")  - the categories in the database end up in English.
Using the _("string")  - the categories in the database end up in Italian.
 
It appears that @vomikan is correct and _("string") should be used here.

guanlisheng:
The desired behavior is to store **Englis** in the database.

stef145g:
This means that the current implementation needs to be fixed to show the correct language for categories. Now the current implementation will only show English.

vomikan:
In that case you can'can't get translated categories by sql. And any new
user created categories will be in national language.

суббота, 13 февраля 2016 г. пользователь Lisheng написал:

gabriele-v:
Completely agree with @stef145g and @vomikan : 
Why we need to store categories in English and add more code to translate them?
Categories are only labels, so better to show them directly as written in DB.
If new category created how does it work with translation and DB save? It's a no-sense..

Regarding build warning, you said lots of times to improve coding and then you leave a warning that could be fixed? I don't understand..
Also, current implementation of ensure_data it's useless because if it doesn't execute any statement in try, what error will be catched?

guanlisheng:
@vomikan @stef145g @gabriele-v 
Regarding English or translated strings, are you guys kidding for such simple question? do we know why we store the ISO date format (YYYY-MM-DD) in database? if negative,  considering the simplest case that final users switch the language setting from **Italian** to **Russian**, how to deal with the existing records in table CATEGORY_V1, SUBCATEGORY_V1 and CURRENCYFORMATS_V1? as what i said times, please distinguish MVC during the software developing. to this particular case, all transaction should be done during the displaying (V in the MVC). 

About the building warnings and code quality/improvement, the empty try statemenet is not perfect, even problematic while we should put more time on how we design, use the c++ class and initialize its members. 
Leave current warnings there and the coming demo data will fill with the empty try statement.

gabriele-v:
I could ask you the same about kidding, you can't absolutely compare a date with a string..
Date it's stored in ISO date only because SQLite doesn't have a proper date field. If MMEX will be on SQL Server or MySQL with proper date type field, problem won't exist at all.

Switching language has only to switch GUI interface translation (and fixed enum like void, reconciled etc..), not data!
I didn't know a software that will switch a user-customizable label or data when you change language. It's like asking to Gmail or Outlook to change categories\tags when you change GUI language..
Completely no-sense up to me.

Consider the case that a user adds a new category outside of ones provided, in which language you will add it to database?

vomikan:
I am on the same point as Gabriele.

воскресенье, 14 февраля 2016 г. пользователь Gabriele-V написал:

gabriele-v:
@guanlisheng , I think that there was a misunderstanding here.

I was referring to "normal" user data, not default ones.
I could agree with you regarding "default" data, for example in Outlook.com default categories are translated if you change language: practically it would mean that, when viewing\showing a category, we have to search if there is a translation for that category, otherwise display as saved, correctly?

This approach works, but IMHO it will add some complexity to MMEX that we currently don't need, in particular with GRM and SQL..

gabriele-v:
Consider also case when a category translation is changed in a new MMEX version or if a translation it's different from Desktop \ Android \ WebApp \ iOS (because translators are different people), users will be really disoriented..

This to obtain what? Only simpler language change after DB initialization?
I don't think that lots of users will change language after initialization.
Technically speaking I could agree with you but IMHO here it isn't worth the trouble..

vomikan:
There are only two 'standard' categories  in mmex:
Transfer and Unknown
All others may be redefine by user.

воскресенье, 14 февраля 2016 г. пользователь Gabriele-V написал:

guanlisheng:
I think everything is pretty clear now since we were/are talking the default/predefined data instead of the regular user data which input from the GUI instead of the raw SQL here.

Regarding the value, i think it's a pretty interesting topic and i would like use the system language to disable the language switch (via #340).  Even this,  we have to have fully examine on whether current code works since i18n is variables.


NOTE, i don't change any existing behavior and what i did is to move the DB initialization from manual written cpp code to auto generated one....  you guys' response is really weird.

gabriele-v:
I don't know why you continue to close this pull request.

This it's not true:

> NOTE, i don't change any existing behavior and what i did is to move the DB initialization from manual written cpp code to auto generated one.... you guys' response is really weird.

because old code will initialize (and show) categories with user language, now they are always initialized (and showed) in english.

So at least until you have completed the automatic language translation IMHO it should be reverted as before.

guanlisheng:
you are wasting my time for such stupid discussion and please read the code firstly before putting your statement!


here are the link to the old code before my commit.
https://github.com/moneymanagerex/moneymanagerex/blob/v1.2.5/src/model/Model_Currency.cpp
https://github.com/moneymanagerex/moneymanagerex/blob/v1.2.5/src/model/Model_Category.cpp

gabriele-v:
1.2.5 => Italian language, creating new DB categories inserted in italian
master @https://github.com/moneymanagerex/moneymanagerex/commit/53ff8912bef5d0b3437f305290a4a43873651352 => Italian language, creating new DB categories inserted in italian
master @https://github.com/moneymanagerex/moneymanagerex/commit/2cfc7feafabc120f74fd9560f7f59cd168836db4 => Italian language, creating new DB categories inserted in english

You are wasting my time to make tests and please test your code after changes!

guanlisheng:
Then you can switch your language setting back to English to see what happened....

i did not notice the special code for category/sub category which is not consist w/ other ones.

Anyway, it's not acceptable to store translated strings in database tables.

stef145g:
A new pull request #712 has been opened to correct the warning only. Improved version.

Issue #674, opened 2016-01-28
New Account created after Cancelling the creation process


MisterY:
- Start creating an account
- After second step click Cancel

Result: account is still created.
Expected result: account should not be created.

PR #706, opened 2016-02-13
fix(#674): Create new account only on finish

siena123:


stef145g:
This still does not work as expected. using Cancel does not cancel the operation.

stef145g:
OOPs, it does work. Mistaken New account for New Database. Merged

PR #712, opened 2016-02-15
fix(msvc.build.warning.C4702): unreachable code

stef145g:


stef145g:
This is to improve code quality

Issue #659, opened 2016-01-23
amounts can't have symbols (CSV)


vomikan:
I found the amounts can't have symbols in the column either.
£123 will fail as well.

http://forum.moneymanagerex.org/viewtopic.php?f=10&t=6331&p=13164#p13164

vomikan:
Related issue is: https://github.com/moneymanagerex/moneymanagerex/issues/557

PR #720, opened 2016-02-21
 fix(#659): ignore preceding currency symbol when parsing amounts

gitys:
Instead of Trim() that ignores only white spaces, ignore all chars up to the first number or '-' char.
This will not handle the case of a trailing currency symbol (I'm not personalty aware there is such a thing)

### DIFF
--- a/src/model/Model_Currency.cpp
+++ b/src/model/Model_Currency.cpp
@@ -175,8 +175,9 @@ wxString Model_Currency::toString(double value, const Data* currency, int precis
 
 const wxString Model_Currency::fromString2Default(const wxString &s, const Data* currency)
 {
-    wxString str = s;
-    str.Trim();
+    // Remove any characters before the number stars. Useful especially for removing currency symbols.
+    wxString str = s.Mid(s.find_first_of("-0123456789"));
+
     str.Replace(" ", "");
     if (currency)
     {


GitCop:

 * Commit: d97abcbe71d2499efdf116899eba2db4dbbd31a7
  * Invalid type. Valid types are **feat, fix, docs, style, refactor, perf, test, chore, revert, cleanup, format**

---

src/model/Model_Currency.cpp
```diff
@@ -175,8 +175,9 @@ wxString Model_Currency::toString(double value, const Data* currency, int precis
 
 const wxString Model_Currency::fromString2Default(const wxString &s, const Data* currency)
 {
-    wxString str = s;
-    str.Trim();
+    // Remove any characters before the number stars. Useful especially for removing currency symbols.
+    wxString str = s.Mid(s.find_first_of("-0123456789"));
```
vomikan: Personally, I feel that it's incorrect. It broke an imput like .20
I beleave that better to use regex
http://stackoverflow.com/questions/1357946/regexp-to-parse-currency-value
https://bytes.com/topic/c-sharp/answers/508843-regular-expression-parse-currency-any-symbol

gitys: Hi Vomikan

1. Why does git cop complain that "fix" is not a valid type?
2. if a number in the format of ".20" is legal the easy solution would simply be to add the decimal point to the find_first_of list
3. Since you prefer regex I'll spend some more time on that and get back to you.
Regards,

vomikan:
Sorry, regex not applicable in that case because an arithmetic may be used.
I suggest the next code:
~~~~~
const wxString Model_Currency::fromString2Default(const wxString &s, const Data* currency)
{
    wxString str = s;
    str.Trim().Replace("$", "");
    wxChar *sep = 0;
    wxNumberFormatter::GetThousandsSeparatorIfUsed(sep);

    if (currency)
    {
        if (!currency->PFX_SYMBOL.empty())
            str.Replace(currency->PFX_SYMBOL, "");
        if (!currency->SFX_SYMBOL.empty())
            str.Replace(currency->SFX_SYMBOL, "");
        if (!currency->GROUP_SEPARATOR.empty())
            str.Replace(currency->GROUP_SEPARATOR, "");
        if (!currency->DECIMAL_POINT.empty())
            str.Replace(currency->DECIMAL_POINT, wxNumberFormatter::GetDecimalSeparator());
    }
    else
    {
        str.Replace(sep, "");
        if (wxNumberFormatter::GetDecimalSeparator()!=0)
            str.Replace(wxNumberFormatter::GetDecimalSeparator(), ".");
    }
    wxLogDebug("%s = %s", s, str);
    return str;
}
~~~~~
I'll create PR with this changes

gitys:
Hi,
Be careful with the PFX_SYMBOL and SFX_SYMBOL.
I looked at the structure in the debugger and neither of them contained my currency symbol.
They contained "USD" and not "$".
You also will still have a problem if the currency is not the same as what is in the file.
Regards,

vomikan:
~~~~~
const wxString Model_Currency::fromString2Default(const wxString &s, const Data* currency)
{
    wxString str = s;
    str.Trim();
    wxChar *sep = 0;
    wxNumberFormatter::GetThousandsSeparatorIfUsed(sep);
    const Data* c = currency ? currency : Model_Currency::GetBaseCurrency();;

    if (!c->PFX_SYMBOL.empty())
        str.Replace(c->PFX_SYMBOL, "");
    if (!c->SFX_SYMBOL.empty())
        str.Replace(c->SFX_SYMBOL, "");
    if (!c->GROUP_SEPARATOR.empty())
        str.Replace(c->GROUP_SEPARATOR, "");
    if (!c->DECIMAL_POINT.empty())
        str.Replace(c->DECIMAL_POINT, wxNumberFormatter::GetDecimalSeparator());

    wxLogDebug("%s = %s", s, str);
    return str;
}
~~~~~

vomikan:
https://github.com/moneymanagerex/moneymanagerex/pull/726

PR #516, opened 2015-07-03
fix(bug.sf615): stop nav tree items from disappearing on Mac

omalleypat:

PR #754, opened 2016-03-09
fix(#516): stop nav tree items from disappearing, minimize news checks

omalleypat:
Changed displayWebsiteNews output to be static so it retains it value on subsequent calls. Only check if it is wxEmptyString (the first time or if pref changed).


omalleypat:
Needs more testing...will reopen later.

Issue #788, opened 2016-04-10
Transactions color tag is lost in some specific case


lexa2:
**MMEx v.1.2.6** on **Win7 x64.**

**Workflow:**
1. Add new transaction by pressing F9 and entering transaction details.
2. Press Ctrl+1 (or any other Ctrl+Digit combo) to add a color tag to the created transaction.
3. Press F9 again and add yet another transaction.

**Expected results:**
Color tag will stay attached to the transaction added at p.1.

**Actual results:**
Color tag gets lost and I have to re-add it by selecting transaction added at p.1 in the list view and pressing Ctrl+1 once again.

guanlisheng:
i am closing this since it will be fixed via 1.3.0 release.

vomikan:
It still actual for 1.3.0

stef145g:
It appears that fix #788 has not been added to master. @vomikan could you check this please.

vomikan:
@stef145g this fixes are useless. In 1.3.0 is multyline selection present.

stef145g:
@vomikan, I originally noticed that this fix was not stopped because of branch sycn issues, and then it was closed. Not sure why. Thanks for pointing this out.

stef145g:
#841 is a related issue. Discovered true problem in 1.3. Will submit fix later.

PR #805, opened 2016-04-25
fix(#788): transactions color

vomikan:


stef145g:
Please rebase and recommit. This should allow AppVeyor to reinitialise.

PR #809, opened 2016-04-30
Translink shares

stef145g:


stef145g:
Some problems in calculations found. Need to recheck and resubmit

Issue #815, opened 2016-05-22
Option: Last Used Payee has no effect.


stef145g:
- OS = Windows [?] Mac OSX  [?] Linux
- MMEX version  - 1.3.x

Option setting: **Last Used Payee** setting set to **None** still brings up an unwanted payee in the transaction dialog and therefore a unwanted category. Attempting to set a different category often confuses the category selection process.

PR #819, opened 2016-06-04
Issue 815

stef145g:
Ref #815


stef145g:
There is still an issue when a selection is made from a limited payee drop-down list. Initially the category for the first payee in the limited list is displayed. On selecting a category, now the correct category can be used.
I have been unable to track down when the category is set as it is occurs in the onchangefocus event.

Issue #567, opened 2015-09-30
Minimum account balance


MisterY:
A request from the users for implementation of the minimum account balance. The user is notified if the new transaction would bring the balance below the set minimum.
This would require a new field, so I'm adding a request here. Related issue for Android - https://github.com/moneymanagerex/android-money-manager-ex/issues/547.

>for second one is to set the "minimum balance for checking A/C".
it could be works as reserve for checking a/c's.
if a checking A/c having balance of INR 100000, and is set at minimum balance like "INR 10000". then this function will stop the transaction from this A/c, after spending INR90000.
It looks like reserve for A/c.

gabriele-v:
I think that we have finally to decide a common and simple way to patch DB adding\removing\updating filelds.

@guanlisheng what do you think? We already spoken about it but without any decision

MisterY:
I would contribute some thoughts from this issue - https://github.com/moneymanagerex/database/issues/4

As Android provides a convenient way to upgrade a database, by providing an onUpgrade method that gets executed when the database needs to be upgraded and provides both old and new version numbers, it is very easy to upgrade by calling only the migration scripts for the missing versions.

Each migration would be a separate script, for example "database_version_x.sql" and on upgrade we would loop `for version = oldVersion .. newVersion` and execute any "database_version_" + version + ".sql" scripts.
I find this by far the most convenient way and easy to implement on all platforms.

The thing to keep in mind would be to maintain the db version information at two location or to use the default PRAGMA UserVersion, provided by SQLite. Android SQLite libraries use this pragma to maintain the version information whereas in desktop version we use a value in the info table. Let me know how easy it is to read/write pragma values and perhaps this could be the most practical way.
Pragma values are available before any tables are created and should be more reliable than a value in a table, technically.
Just my $0.02.

guanlisheng:
@gabriele-v  i'm fine if we could find a easy way to patch database.

stef145g:
This feature has now been completed. Edit the account to set the initial values.

Extra account data is stored for each account in the table: INFOTABLE_V1 using json string variables.
The implementation contains: Credit Limit and Minimum value. The difference is that the Credit limit will be used as a negative value.

stef145g:
Completed. The extra account information has been added to the accounts table, without JSON data.

MisterY:
Great. I've updated the db schema to v7 in the mobile app.

Issue #538, opened 2015-07-31
Reconciled Statement Date on Accounts


AgentB1:
Most accounts are dynamic (changing) in the last month and older transactions never change.

Each month the bank statements need reconciling.

It would be helpful to have a "Latest Reconciled Statement Date" or "Statement Date" on each account.  

If enabled and a date set, no transactions can be removed, altered or added, up to to and including that date. 

If disabled, all transactions can be modified or deleted.

The default date if set is current date.

A Transaction Filter could then also be created "Since Statement Date" which would show the transactions you expect to see on next statement.

vomikan:
I suggest to add warning message if some transaction older than last month added or edited.

AgentB1:
Yes that would be helpful, or even "older than most recent reconciled transaction".

stef145g:
This feature has now been completed using a 'json store'.

Extra account data is stored for each account in the table: INFOTABLE_V1 using json string variables.

Edit the account to set the initial values for that account.

stef145g:
Completed. Extra account info added to account table without JSON data

Issue #451, opened 2015-04-14
Add credit limit to credit card accounts


MisterY:
Credit card accounts usually involve a credit limit and it can be costly to hit this or go over. It would be useful to indicate how far is the credit on the credit card from this limit. This can be indicated visually.
For this there needs to be a limit value stored in the database.
If you agree, could you provide some hints in regards to where this would be stored so that we can also work on implementation for the Android version?
Cheers

AlessandroLazzari:
:+1:

stef145g:
:+1:  I like it. Hopefully we can work together and define the specifications first.

-------------------------------
Starting this discussion...
This would involve adding several fields to the Account table to expand the functionality.

**Proposal**
Provide a new table providing a 1 - 1 relationship, linking the credit record to the account record.

**New table Name:** CREDIT_V1
**Fields:**
CREDITID - integer ... primary Key
ACCOUNTID - integer  ... the associated account id.
MAXLIMIT - numeric ... Upper credit card limit.
DUEDATE - text (date) ... data at which extra charges could be added
INTERESTRATE - numeric ... working out interest charges if any.
AMOUNT: numeric ... useful for providing a fixed repayment amount if necessary.

stef145g:
Suggestion of changing **ID** field names in a table as follows:
CREDITID ...... set to **ID** .... primary field.
ACCOUNTID .. set to **ID_ACCOUNT**
This would allow easy identification of links, especially if there are a many links in the table.
Old tables cannot be changed for obvious reasons, but suggested for new tables use.

giuseppepaoletti:
Hi all,
sorry to say that in the past month I've had no time to dedicate to this 
project, due to work and family problems.
Regarding to advanced credit card information, I have started to develop 
a similar solution, but haven't done any commit for the above problems.
I've added a new table, named CREDITCARD_V1, with those fields:
CARDID integer primary key
ACCOUNTID integer NOT NULL UNIQUE (credit card account id)
HELDAT integer (bank account where card is helded and money will be 
withdrawed)
CARDTYPE integer (debit, credit (partial refund or revolving), credit 
(full refund))
RATE numeric (interest rate)
CARDLIMIT numeric (credit card limit)
Those are the fields added to account dialog properties:

New features are:
- warning message when credit card limit is reached
- automatic creation of a recurring transaction if "Auto Balance" is 
checked; if card is a credit card with partial refund, balance amount 
[...]

PR #859, opened 2016-07-16
Json store 

stef145g:
Using json to store extra account information using the existing table infotable_v1
Features implemented in the following issues: #451, #538, #567

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/859)


gabriele-v:
Stefan, I completely disagree on this implementation.

Surely JSON it's the simplest way, but now that we have upgrade DB process working, why we need to complicate DB structure inserting JSON data and in another table that should contain only information? Good luck if you try to extract them with a SQL query for GRM..

Simply alter table account_v1 adding needed fields.

@moneymanagerex/desktop-mmex-team what do you think?

MisterY:
@gabriele-v, this was my first thought as well. I just came here to see if it is possible at all to use this in the Android version. Without any spec on schema, now it is even more fluid in how we implement features or at least keep them in sync among different platforms.
It is unfortunate that we are not making more efforts in pushing the schema upgrade mechanism to production. There are many useful features waiting on this.
I thought extending the credit card account information would be one of those that would really benefit and bring some UI goodness in the mobile app as people track their spending. However, by mixing JSON structures with SQL tables, it just gets messier. Besides, I do not understand the reasoning behind. Are there any issues with extending the database schema?

On the positive side, @stef145g, I'm really happy that you moved in the direction of implementing these features. Wish it was done a bit earlier so that I could implement them for Android in parallel before the Summer holidays but it will do for an Autumn update. I'm really looking forward to adding progress bars on credit cards. This is one of the most useful options for credit accounts.

stef145g:
For me, creating a JSON store was a way of understanding JSON. It proved very useful to me as it was a quick way of prototyping long awaited features. Now that these features are in place, it now becomes a simple procedure to transfer the storage to the database file. Another reason for creating methods in a JsonAccount class. The JsonStore class should be able to simplify other existing data storage cases. 

While I agree with Gabriele about GRM, the current implemented fields are mainly used for control purposes, and possibly would not be used in a report.

I will withdraw this pull request after a few days to allow others to comment on the GUI implementation. These features can be excluded from v1.3 and added in a future release, when we have updated the database.

stef145g:
Removed the Account storage using JSON data. New fields have now been added to the account table.
Converting this to database storage was easier that I thought.

MisterY:
Who can merge the database changes? https://github.com/moneymanagerex/database/pull/16
I'd love to update the schema in the Android app and perhaps start development utilizing the new fields.

Issue #906, opened 2016-09-23
Performance killers


vomikan:
- Check OS
  - [x] Windows
  - [ ] Mac OSX
  - [ ] Linux 
-  Check MMEX version
  - [x] 1.3.x

mmcheckingpanel.cpp:209
~~~~~
        {
            if (!Model_Checking::TRANSDATE(tran).IsBetween(m_quickFilterBeginDate, m_quickFilterEndDate))
                continue;
        }
~~~~~

vomikan:
Also what about to change 
-  static std::unordered_map<wxString, wxDate> cache;
+ static std::map<wxString, wxDate, std::greater<wxString>> cache;

vomikan:
~~~~~
bool in_the_future = Model_Checking::TRANSDATE(&tran) > wxDateTime::Now();
~~~~~
~~~~~
const wxString mmGetDateForDisplay(const wxDateTime &dt)
~~~~~
~~~~~
void mmHTMLBuilder::addTableCell(const wxDateTime& date)
~~~~~

vomikan:
~~~~~
bool Model_Billsdeposits::AllowTransaction(const Data& r)
{
    Model_Account::Data* account = Model_Account::instance().get(r.ACCOUNTID);
    double current_account_balance = Model_Account::balance(account); //speedup needed
~~~~~

PR #907, opened 2016-09-23
perf(#906): transactions filters improvements

vomikan:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/907)


guanlisheng:
I prefer the original one as it's more straightforward. I think as wxdate object is better than a date string.

Is this the performance bottleneck ?

stef145g:
Improved performance verified

PR #916, opened 2016-09-29
perf(#906): speedup, cache

vomikan:
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/916)

### DIFF
--- a/src/assetspanel.cpp
+++ b/src/assetspanel.cpp
@@ -630,7 +630,7 @@ wxString mmAssetsPanel::getItem(long item, long column)
     case COL_VALUE_CURRENT:
         return Model_Currency::toCurrency(Model_Asset::value(asset));
     case COL_DATE:
-        return mmGetDateForDisplay(Model_Asset::STARTDATE(asset));
+        return mmGetDateForDisplay(asset.STARTDATE);
     case COL_NOTES:
     {
         wxString full_notes = asset.NOTES;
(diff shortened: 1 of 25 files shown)


guanlisheng:
i think it's invalid as current performance is pretty okay in mac platform. would you mind to provide some proof here to tell the wxDate is the bottleneck.

vomikan:
It's a final changes to avoid date parsing and cleanup.
I think it may be accepted.

stef145g:
I will test this before accepting this

guanlisheng:
Thanks and we may test whether it will improve the performance 

 
Thanks
Lisheng 

Sent from mobile

stef145g:
mmex_tests used to test performance of patch. Tests builds without patch and with patch, created.
Comparing runs for each build and looking at times for: Account View: xx 
Each run produces different times, depending on other process from CPU
Tests can be built using mmex_texts update (#922)

guanlisheng:
And what's the result for the performance test?seems you just merged the patch

stef145g:
Tests were acceptable so I merged the patch.
Tests updated in order to allow others to build and get test results on their machines.

stef145g:
I am also using my updated build with my data at present.

guanlisheng:
The question is whether there is any runtime number here to tell us that It does improve the performance

PR #968, opened 2016-12-26
fix(#957): possibly fix

vomikan:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/968)


vomikan:
https://openclipart.org/detail/33583/architetto-pulsante-02

PR #1054, opened 2017-01-29
chore(*): pot update and web server sql api fix

vomikan:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1054)

### DIFF
--- a/src/webserver.cpp
+++ b/src/webserver.cpp
@@ -46,9 +46,9 @@ static void handle_sql(struct mg_connection* nc, struct http_message* hm)
 
     json::Object result;
     result[L"query"] = json::String(wxString(query).ToStdWstring());
-    //bool ret = Model_Report::instance().get_objects_from_sql(wxString(query), result); 
+    Model_Report::instance().get_objects_from_sql(wxString(query), result); 
 
-    for (const auto & r : Model_Setting::instance().all())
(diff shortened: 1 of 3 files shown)


src/webserver.cpp
```diff
@@ -46,17 +46,17 @@ static void handle_sql(struct mg_connection* nc, struct http_message* hm)
 
     json::Object result;
     result[L"query"] = json::String(wxString(query).ToStdWstring());
-    //bool ret = Model_Report::instance().get_objects_from_sql(wxString(query), result); 
+    Model_Report::instance().get_objects_from_sql(wxString(query), result); 
 
-    for (const auto & r : Model_Setting::instance().all())
+    /*for (const auto & r : Model_Setting::instance().all())
```
guanlisheng: why commented out this ?

vomikan: Becouse uncommented this:
-    //bool ret = Model_Report::instance().get_objects_from_sql(wxString(query), result); 
 +    Model_Report::instance().get_objects_from_sql(wxString(query), result);

vomikan: Output all setting but not sql query result has looks like test debug solution.
I've commented out settings output and uncommented sql query output.

Issue #1085, opened 2017-03-06
Custom Dates for GRM


vomikan:
I suggest to add date ranges for GRM. 

![screenshot 75](https://cloud.githubusercontent.com/assets/6836805/23610025/f8c83844-0281-11e7-9325-ae5c351a6f0d.jpg)

This dates may be used in SQL script like &begin_date     &end_date
-  Check MMEX version
  - [x] 1.4.x

vomikan:
@siena123 James, I need help with transfer parameters to SQL script.

Here may be code to replace placeholders to real data:
https://github.com/moneymanagerex/moneymanagerex/compare/master...vomikan:%231085?expand=1#diff-f8a5929a16cdeff021409c02a03484f3R169

I've attached test DB to see how parameters looks in GRM reports
[test7.zip](https://github.com/moneymanagerex/moneymanagerex/files/823788/test7.zip)

vomikan:
I've attached test DB to see how parameters looks in GRM reports
[test7.zip](https://github.com/moneymanagerex/moneymanagerex/files/824783/test7.zip)

siena123:
I see the direction you are going. I will look at the parameter transfer in more detail this weekend.

vomikan:
@siena123 Please investigate the changes regarding this issue.
I think this way we can add more options for payees, categories etc.

vomikan:
Help file update needed and samples

vomikan:
There are four options available for sql script:
```&single_date
&begin_date
&end_date
&budget_years```

vomikan:
For example for sql:
```select  '&budget_years' as year```
The report header will contain budget years choice:
![image](https://user-images.githubusercontent.com/6836805/38904242-f2c9d9a4-42b1-11e8-8b83-b2c78c71c59a.png)

vomikan:
[test7.zip](https://github.com/moneymanagerex/moneymanagerex/files/1921958/test7.zip)

gabriele-v:
@vomikan we can close this? Documentation seems good, or you have to add something else?

PR #1088, opened 2017-03-07
feat(#1085): parameters for grm stage 1 

vomikan:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1088)


siena123:
I like your idea of changing to parameters because I was looking at making another optional parameter for some of the reports when I got the time to work on it.

Issue #1180, opened 2017-06-06
Reconciliation of transfer between accounts should be done separately in the accounts


corse63:
make sure these boxed are checked before submitting your issue - thank you!
- Check OS
  - [x ] Windows
  - [ x] Mac OSX
  - [ x] Linux 
-  Check MMEX version
  - [x ] 1.3.x
  - [ ] 1.2.x
  - [ ] 1.1 or older

corse63:
In the current version, reconciliation of a transfer between accounts is somehow painful:
When you do reconciliation of an operation, you do it either in the source or destination account, however at the end, it appears as reconciliated in both accounts. 

In fact, when you do a reconciliation, you do it based on an account statement and you do it for a given account (say the source one). When you receive the statement for the other account (say the destination one), it is then possible to make the reconciliation, and only then.

For this reason, I would suggest to make reconciliation only in the concerned account where you are working.

stef145g:
This has been a long standing issue reported many times before. I think I have a way of solving this problem using the current database schema. I will investigate this further.

stef145g:
PR #1208 has been created that will address this issue.

PR #1208, opened 2017-07-01
#1180 Split transaction Status for Transfers

stef145g:
This sets the status of a transfer as a 2 character token instead of one.
The database status is recorded as NN for both sides being unreconciled but is viewed as normal in Account View.
When the A side is reconciled and B side as Follow-up the database shows RF.
Viewing the transactions, the A side will show the status as R and viewing the B side will show as F

On changeover, both sides of a transaction will assume the current status for both sides until it is modified by the update. Reverting back and remodifying the transaction will restore to original.

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1208)

### DIFF
--- a/src/filtertransdialog.cpp
+++ b/src/filtertransdialog.cpp
@@ -725,7 +725,7 @@ bool mmFilterTransactionsDialog::checkAll(const Model_Checking::Full_Data &tran,
         ok = false;
     else if (getCategoryCheckBox() && !checkCategory<Model_Checking>(tran))
         ok = false;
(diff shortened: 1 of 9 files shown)


stef145g:
PR #1208 has been created that will address this issue.

slodki:
We shouldn't change dictionary changes to the database without db version change and existing entries update.
This breaks android app without warning.

gabriele-v:
Agree with @slodki 
We should also provide a patch to set all previous transfers to RR, otherwise we will have some transfers with two characters and some with just one

PR #1213, opened 2017-07-03
chore(Submodule.Database): update

stef145g:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1213)


grumpy235:
Is UK Pound missing a group separator in patch file as £12345.67 displays as £12 345.67 rather than the more normal £12,345.67
On my CURRENCYFORMATS_V1, I have ',' but of course that might have been a local update, long time since I set it up.

PR #1275, opened 2017-08-05
fix(Usage.Statistics): add title to path for unknown dialog

stef145g:
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1275)


stef145g:
On reviewing Google/analytics the same data can be obtained by manipulating the 'BEHAVIOUR/Overview' and 'BEHAVIOUR/Site Content/All Pages' view therefore this may not be required.

PR #1281, opened 2017-08-10
Source tree cleanup - part 2

slodki:
For changes description see commit messages and #1268.

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1281)


slodki:
CMake generates multiple targets inside project:
- libraries (lua etc.)
- translations
- mmex executable
- all (default target)
- install
- package (it runs all above then start CPack tool)

If you compile only executable no other files will be ready. Default target `all` is similar - files are prepared but in their own directories. You must use `install` target to copy all required files to specific location. All the logic for copying files is build into CMake script and is changed often.

I don't use VS GUI - there should be option in the IDE to select target (not confuse with target platform etc.) to build. Then no additional steps/batch files will be required.

If you want run this step manually (not using IDE) it's even simpler: in place of `update_cmake_build.bat` run `cmake -DCMAKE_INSTALL_PREFIX=/new/tmp/dir -P cmake_install.cmake`. Or define this as custom command somewhere in IDE started from custom button. You can select any target directory you like (even %TMP%\test\mmex or similar).

You can see `cmake_install.cmake` is very similar to old bat file, but it's written in CMake script language and it's portable. It's auto-generated and synced with changes in CMake toolchain. There is no need to duplicate the logic in separate batch file and sync changes between them.

Could you try this method? And check if its possible to select build target from the VS GUI?

Issue #1290, opened 2017-08-12
CSV import with no file selected crashed MMEX


slodki:
1. Start MMEX with any database open
2. Open CSV import dialog: `File` -> `Import` -> `CSV Files...`
3. Add *Date* and *Amount* fields to the list
4. Click `Import` button
5. Pop-up window is displayed with *Total imported: 0* message
5. MMEX crashes afetr few seconds without any user actions (when pop-up is displayed)

```log
*** Error in `mmex': free(): invalid next size (fast): 0x00007ffe1ff27eb0 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x7908b)[0x7f9dedc0008b]
/lib/x86_64-linux-gnu/libc.so.6(+0x82c3a)[0x7f9dedc09c3a]
/lib/x86_64-linux-gnu/libc.so.6(cfree+0x4c)[0x7f9dedc0dd2c]
/usr/lib/x86_64-linux-gnu/libwx_gtk2u_core-3.0.so.0(_ZN12wxWindowBase7DestroyEv+0x37)[0x7f9df0b6d9a7]
/usr/lib/x86_64-linux-gnu/libwx_gtk2u_core-3.0.so.0(_ZN12wxWindowBase15DestroyChildrenEv+0x25)[0x7f9df0b6d9e5]
/usr/lib/x86_64-linux-gnu/libwx_gtk2u_core-3.0.so.0(_ZN8wxWindowD1Ev+0xa7)[0x7f9df0984037]
/usr/lib/x86_64-linux-gnu/libwx_gtk2u_adv-3.0.so.0(+0x13103a)[0x7f9def83e03a]
/usr/lib/x86_64-linux-gnu/libwx_baseu-3.0.so.0(_ZN16wxAppConsoleBase20DeletePendingObjectsEv+0x42)[0x7f9deec78392]
/usr/lib/x86_64-linux-gnu/libwx_baseu-3.0.so.0(_ZN16wxAppConsoleBase11ProcessIdleEv+0x61)[0x7f9deec78411]
/usr/lib/x86_64-linux-gnu/libwx_gtk2u_core-3.0.so.0(_ZN9wxAppBase11ProcessIdleEv+0x21)[0x7f9df0a2c661]
/usr/lib/x86_64-linux-gnu/libwx_gtk2u_core-3.0.so.0(_ZN5wxApp6DoIdleEv+0x76)[0x7f9df09392e6]
[...]

PR #1291, opened 2017-08-12
Fix errors and add detailed messages in mmUnivCSVDialog

slodki:
Add details to diagnostic messages (for user and log).
Fix wrong total lines number displayed.
Update progress meter to show real values.
Change all counters to the same type (long).
Select summary pop-up icon and buttons depending on import result.
Commented out messages never displayed to user.

This should help users to diagnose problems with CSV imports (like #1277).

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1291)


slodki:
And now should fix #1290 too

PR #1296, opened 2017-08-15
fix(CMake): set correct location of mmex.exe for Windows Install

stef145g:
This allows the Install project in MSVC IDE to work correctly

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1296)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,7 +286,7 @@ ELSEIF(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
             BUNDLE DESTINATION .)
 ELSEIF(CMAKE_SYSTEM_NAME STREQUAL "Windows")
     INSTALL(PROGRAMS
-            "${CMAKE_SOURCE_DIR}/bin/${PROJECT_TARGET_BIN}.exe"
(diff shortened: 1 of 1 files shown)


stef145g:
This does not appear to work in AppVeyor, but works in the local MSVC IDE

CMakeLists.txt
```diff
@@ -286,7 +286,7 @@ ELSEIF(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
             BUNDLE DESTINATION .)
 ELSEIF(CMAKE_SYSTEM_NAME STREQUAL "Windows")
     INSTALL(PROGRAMS
-            "${CMAKE_SOURCE_DIR}/bin/${PROJECT_TARGET_BIN}.exe"
+            "${CMAKE_SOURCE_DIR}/bin/Release/${PROJECT_TARGET_BIN}.exe"
```
slodki: Under VS build multi-tagret generator is used, so build type is unknown at configure time, so we cannot use hardcoded paths here - we must replace `INSTALL(PROGRAMS` with `INSTALL(TARGETS` here...

Issue #1300, opened 2017-08-19
Privacy tab not displayed at startup


slodki:
After removing one page from About dialog notebook Privacy tab is not displayed at startup:

![mmex-assert-about](https://user-images.githubusercontent.com/3155552/29485293-d56dfe4e-84cf-11e7-80d2-3fa33d1bfcf0.png)

Trivial fix will be delivered...

PR #1303, opened 2017-08-19
feat(CryptoCurrencies): add crypto currencies and wallets

gabriele-v:
Add basic support for crypto currencies and wallets.

Currency type it's needed to manage different change rate sources (yahoo don't provide quotes for crypto currencies) that will be implemented later.

Opened PR on database repository too:
https://github.com/moneymanagerex/database/pull/19

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1303)


slodki:
I see you have found location of #1300

gabriele-v:
Yes, because I written that piece of code :)

I think that with one of latest commits one page has been deleted.

gabriele-v:
Actually don't exists crypto currency bank account or credit card, they are all digital wallet.

I think we need a new account type for two reasons:
- Grouping in home page to undestrand how much money are "virtual" and how much are "real"
- Grouping in left tree to keep them collapsed > considering that one account can have only one currency, one account it's needed for each crypto currency and so lot's of accounts (generally with just few transactions) will be created

slodki:
It was me who deleted one page in about dialog :smile:
And I changed this only in /tests tree :stuck_out_tongue_closed_eyes:

slodki:
JPY or MNT are just as virtual for me as the XBT - I cannot use them in Europe :smile: 

There are [real virtual currencies](https://www.currency-iso.org/en/home/tables/table-a2.html) used in banking without correlation to blockchain:
>The CLF is a daily economically-financial unit calculated by the Central Bank of Chile according to inflation (as measured by the Chilean Consumer Price Index of the previous month). The value of the CLF is expressed in terms of Chilean Pesos per CLF. The use of CLF has been widely extended to all types of bank loans, financial investments (time deposits, mortgages and other public or private indexed instruments), contracts and fees in some cases

Two enchantments could be more usefull:
1. ability to group accounts (even of different types!) in MMEX GUI and display group summaries. Build account multi-level hierarchy could be even better...
2. ability to define change rate source as an attribute of selected currency/share - different for different currency. I think the same share could have different attributes for different accounts - when trading take place at multiple stock exchanges...

Your approach is not reusable (what additional *currency types* can we have except *traditional* and *virtual*? *historical* like DEM maybe).

gabriele-v:
JPY it's not virtual here in Europe: if you have a JPY account you can pay Euro with card and change is done by your bank or card circuit. 

The same thing it's not valid (at least yet) for BTC or CLF. We could divide them in "digital" and fiat currencies if you prefer. 

Regarding the different grouping could be useful, but it's a long-term goal, actually you need to change quite a lot to do it, with this change we could advertise as "digital currencies ready" with few minor changes.

Instead agree about second proposal, main issue remains that we need a manual C++ code parser for each type of source, maybe we could consider some interface or something like an API to simplify adding a new source?

slodki:
Yeh, this temporary solution is good for advertisement :+1:

gabriele-v:
What about naming conventions? Better "Traditional" and "Crypto" or "Fiat" and "Digital" ?
@moneymanagerex/desktop-mmex-team what do you think?

slodki:
This can be simple boolean column as it will be hard to add any additional value to this - it simple digital and non-digital flag. If we want additional currency attribute (like historic for DEM etc.) we must use next column because we must allow matrix of traditional-historic and crypto-historic...

stef145g:
@gabriele-v The update to the database submodule can now be included in this PR

vomikan:
@gabriele-v could you apply my fixes and update the brunch?

vomikan:
[crypto_fixes.zip](https://github.com/moneymanagerex/moneymanagerex/files/1689453/crypto_fixes.zip)

gabriele-v:
@vomikan could you take care of this? As explained I have no more free time for mmex, but just few minor changes are needed to implement it, you can copy them (or try to merge) from my branch and adding your suggested changes.

Thanks

Issue #1334, opened 2017-10-04
Reports sort order


vomikan:
I'll prefer not to mix GRM reports and built ones.

![screenshot 558](https://user-images.githubusercontent.com/6836805/31174593-afcce0ba-a914-11e7-85c3-5b86b80aeb0d.jpg)

Your OS:
- [x] Windows
- [x] Mac OSX
- [x] Linux 
- [ ] Android -> please report bugs to moneymanagerex/android-money-manager-ex repo!

MMEX version:
- [x] 1.4.x
- [ ] 1.3.x
- [ ] 1.2.x
- [ ] 1.1 or older
- [ ] development version (git master branch)

It depend on (mmframerepott.cpp)
```
    m_nav_tree_ctrl->SortChildren(reports); 
```

slodki:
We can divide reports tree to two separate: *Reports* and *User Reports*.

PR #1351, opened 2017-10-18
fix(#1334): reports sorting

vomikan:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1351)


slodki:
The sort order will depend on GUI language selected or not?

vomikan:
untitled

PR #1379, opened 2017-12-09
fix build and copyright information

siena123:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1379)

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -105,8 +105,7 @@ tools to manage projects in VS IDE.
 
 1. Generate build environment using [CMake]
 
-       mkdir c:\projects\mmex\build
-       cd c:\projects\mmex\build
+       cd c:\projects\mmex
        set "PATH=%PATH%;%DevEnvDir%CommonExtensions\Microsoft\CMake\CMake\bin"
        cmake -G "Visual Studio 15 2017" ..
 
--- a/contrib.txt
+++ b/contrib.txt
@@ -363,7 +363,7 @@ willmetz
 ##GNU general public license
 Copyright (C) 2009-2017 Nikolay
 Copyright (C) 2011-2017 LiSheng [guanlisheng]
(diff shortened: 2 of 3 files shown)


GitCop:
Thanks for your pull request while there were the following issues

 * Commit: 80afdd6c6fa2fccf28cf80c6c32d56b838931040
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

---

BUILD.md
```diff
@@ -105,8 +105,7 @@ tools to manage projects in VS IDE.
 
 1. Generate build environment using [CMake]
 
-       mkdir c:\projects\mmex\build
-       cd c:\projects\mmex\build
+       cd c:\projects\mmex
```
slodki: You don't want build mmex in-place mixing sources with different binary versions - CMake standard way is to build out-of-source, leaving git repo clean.
Your change brakes this example too - `cmake ..` below will not find required `CMakeLists.txt`

siena123:
The directions were wrong to start with as the CMakeList.txt does not exist in the empty build directory that was just created. If the intent is to build in different directory then the instructions should be to copy the needed files into the directory after creating it and I am not familiar enough with the new build process to know what files would need to be copied into the newly created directory and if any changes would need to be made to them to work in there. 

James

siena123:
I tested just copying the CMakeList.txt into the newly created build directory and as I mentioned for it too work it also needs the build files updated for it to work. The way the files are currently coded it requires cmake to be run the the main directory. 

James

slodki:
@siena123: you missed important two dots (`..`) at the end of the next cmake command - this points cmake tool to project dir, where CMakeList.txt is located.

You can also start from `C:\TEMP\MY_BUILD\` directory then run `cmake -G "Visual Studio 15 2017" c:\projects\mmex` from there.

Issue #1456, opened 2018-03-01
Currency separator not displaying on latest Master


grumpy235:
I have downloaded and compiled (with latest VS2017) the source at tag v1.4.0-Alpha.3 commit 79562d4 and the Home Page displays totals correctly i.e. 12345.67 displays as £12,345.67 with currency set to UK Pounds. Transactions display correctly as well - no currency symbol but thousands separator is present.

Download and compile with latest master at commit 532956f and 12345.67 now appears as £12345.67 on the Home Page and as 12345.67 in a transaction display - thousands separator has disappeared.
 
Using the same test database for both levels of source and CURRENCYFORMATS_V1 table looks fine.
Can you reproduce or is this "user error" ?

Your OS:
- [X] Windows
- [ ] Mac OSX
- [ ] Linux 
- [ ] Android -> please report bugs to moneymanagerex/android-money-manager-ex repo!

MMEX version:
- [X] 1.4.x
- [ ] 1.3.x
- [ ] 1.2.x
- [ ] 1.1 or older
- [ ] development version (git master branch)

vomikan:
![default](https://user-images.githubusercontent.com/6836805/36856183-70d65288-1d86-11e8-85c2-58b20000be41.JPG)

vomikan:
https://forums.wxwidgets.org/viewtopic.php?t=39002

slodki:
In 21ad7d9b76dba08348714fa88eb10d1ddafe37ba we [stopped always use English locale](https://github.com/moneymanagerex/moneymanagerex/commit/21ad7d9b76dba08348714fa88eb10d1ddafe37ba#diff-4c1142bf1ba27682b8dfed154c0c56a3L145) in mmex. Now system configured locale settings are used - user can change them outside mmex app.

Please verify your system settings - locale installed, used, currency settings.

grumpy235:
Thanks guys. I'll have a play.
So although 1.4.0Alpha3 tag is dated Aug 11th and 21ad7d9 is dated earlier (Aug 3rd) it did not make it to Alpha3 ?

grumpy235:
Changed language to English and to English(US) under VIEW menu on latest MMEX and still NO Grouping char for either language.
Checked Windows 10 "Region & Language" panel and they are as expected - United Kingdom for Region and English(United Kingdom) for Language. They have always been that for years!
[...]

Issue #1358, opened 2017-10-22
The transaction dialog shows date in MM/DD/YYYY format whatever the locale


lonemadmax:
The input widgets for dates (transaction dialog, recurring transactions, etc) show dates in MM/DD/YYYY format, despite the locale indicating a different one and the format chosen in preferences being also different. Calendar widgets are in English and start the week on sundays, again not in accordance with the locale.

The dates in transactions lists and reports correctly use the format from the preferences.

My locale:

```
$ locale
LANG=es_ES.utf8
LC_CTYPE="es_ES.utf8"
LC_NUMERIC="es_ES.utf8"
LC_TIME="es_ES.utf8"
LC_COLLATE="es_ES.utf8"
LC_MONETARY="es_ES.utf8"
LC_MESSAGES="es_ES.utf8"
LC_PAPER="es_ES.utf8"
LC_NAME="es_ES.utf8"
LC_ADDRESS="es_ES.utf8"
LC_TELEPHONE="es_ES.utf8"
LC_MEASUREMENT="es_ES.utf8"
LC_IDENTIFICATION="es_ES.utf8"
LC_ALL=
$ locale -k LC_TIME
abday="dom;lun;mar;mié;jue;vie;sáb"
day="domingo;lunes;martes;miércoles;jueves;viernes;sábado"
abmon="ene;feb;mar;abr;may;jun;jul;ago;sep;oct;nov;dic"
mon="enero;febrero;marzo;abril;mayo;junio;julio;agosto;septiembre;octubre;noviembre;diciembre"
am_pm=";"
d_t_fmt="%a %d %b %Y %T %Z"
d_fmt="%d/%m/%y"
t_fmt="%T"
t_fmt_ampm=""
era=
era_year=""
era_d_fmt=""
alt_digits=
era_d_t_fmt=""
era_t_fmt=""
time-era-num-entries=0
time-era-entries="d"
week-ndays=7
week-1stday=19971130
week-1stweek=4
first_weekday=2
first_workday=2
cal_direction=1
timezone=""
date_fmt="%a %b %e %H:%M:%S %Z %Y"
time-codeset="UTF-8"
```

Defining LC_ALL or LANG doesn't make a difference. Changing to another language does change the interface (when set to use system locale) but doesn't change the format in the date and calendar widgets.

<hr><p><b>Version</b>: 1.4.0-Alpha.3 ⬥ db 7 ⬥ git d0f05a17 (2017-10-19) master</p>
[...]

PR #1225, opened 2017-07-09
feat(bug.report): new pre-filled bug report template for GitHub

slodki:
Pre-fill bug report template for Help->Report a Bug.
Update GitHub ISSUE_TEMPLATE.md.

When user use `Help`->`Report a Bug` menu command following page will be opened:
mmex-newbug-1

Rendered as:
mmex-newbug-2

Similar changes are introduced to issues created directly from GitHub website:
mmex-newbug-3

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1225)

PR #1495, opened 2018-03-23
Translations

slodki:
See commit messages for details.
Fixes for #1459 #1358 #1456 #1225

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1495)


src/constants.cpp
```diff
@@ -117,7 +118,7 @@ const wxString mmex::getProgramDescription()
         << bull + LUA_RELEASE << "\n"
         << bull + curl << "\n\n"
 
-        << wxString::Format(_("Build on %s %s with:"), __DATE__, __TIME__) << "\n"
+        << _("Build on") << " " << __DATE__ << " " << __TIME__  << " " << _("with:") << "\n"
```
vomikan: It's bad for translators. What does "with" mean?

slodki: It could be simpler if i change this function to `const wxString mmex::getProgramDescription(bool translate = true)`...

Issue #1528, opened 2018-03-31
Some currency names cannot be translated


slodki:
ISK, NIO, STD, VEF names are not marked for translation and cannot be displayed in user selected language.

stef145g:
#1531 has been created to test this translation issue. The names contain UTF-8 characters in the names and there were not translated for a reason. Please test before accepting this PR as this may cause other issues.

slodki:
@stef145g Please look at moneymanagerex/database@c7f3114606d289313919bb3fe376952327da65f2 and moneymanagerex/database@81227d73453dbcb129a52934096c35b3ca22f3fe
Global solution for this issue and moneymanagerex/database#20

stef145g:
@slodki I had a look at database and found an error. Duplicate entry hs now been commented out.
Also sqlite2cpp has been modified to accommodate translation of Unicode names as well and added to existing PR. Please review

slodki:
I've tested on Linux and Windows and this is enough to use `wxTRANSLATE(L"unicode string")` and this will work without problems. I cannot reproduce #739 with current wxWidgets and our CI toolchain (no `#pragma execution_character_set("utf-8")`required).

You can test with [windows binaries](https://ci.appveyor.com/project/slodki/moneymanagerex/build/1439) yourself from my [test-utf8-translation](https://github.com/slodki/moneymanagerex/commits/test-utf8-translation) branch. *Venezuelan Bolívar* is translated into Polish, and *Icelandic króna*, *Nicaraguan córdoba*, *São Tomé and Príncipe dobra* are left without translation and are displayed correctly in Linux, Windows, English or Polish translation.

vomikan:
@slodki no problems with this release
![currecy](https://user-images.githubusercontent.com/6836805/38437062-0157db14-39df-11e8-8cd4-63ac1c0071b6.png)

Issue #739, opened 2016-02-29
Currency not Latin fields issue


vomikan:
![screenshot 29](https://cloud.githubusercontent.com/assets/6836805/13382946/6f54c20e-de91-11e5-8c1d-b20412b1b8b6.jpg)
- [x] Windows 7x64
- [x] 1.3.0-alpha

vomikan:
~~~~~~
INSERT INTO CURRENCYFORMATS_V1 VALUES(119,'São Tomé and Príncipe dobra','Db','','.',' ','','',100,1,'STD');
~~~~~

guanlisheng:
it works fine for mac

vomikan:
On INSERT for all currencies the result incorrect if non ASCII symbols used
![screenshot 37](https://cloud.githubusercontent.com/assets/6836805/13668088/72c97230-e6cb-11e5-995f-2eef00ccc3e8.jpg)

vomikan:
Also incorrect code
~~~~~
wxTRANSLATE("United States dollar"), wxTRANSLATE("$"), wxTRANSLATE(""), wxTRANSLATE("."), wxTRANSLATE(" "), wxTRANSLATE(""), wxTRANSLATE(""), wxTRANSLATE("USD")));
~~~~~
wxTRANSLATE(...) only for name needed. Should be 
~~~~~
wxTRANSLATE("United States dollar"), L"$", L"", ".", " ", L"", L"", "USD")));
~~~~~

vomikan:
I've discovered that the code:
~~~~~
    wxString s = wxString::Format(L"INSERT INTO CURRENCYFORMATS_V1 VALUES (2, '%s', '%s', '%s', '%s', '%s', '%s', '%s', 100, 1, '%s')"
        , "European euro", L"€", "", ".", " ", "", "", "EUR");

~~~~~
Working fine if placed, for example, in maincurrencydialog.cpp
![screenshot 39](https://cloud.githubusercontent.com/assets/6836805/13697091/32161ef8-e77c-11e5-978f-1538dadc452f.jpg)

But this code in DB_Table_Currencyformats_V1.h provide another result:
![screenshot 40](https://cloud.githubusercontent.com/assets/6836805/13697267/632a2736-e77d-11e5-8ae5-77835f031a30.jpg)

vomikan:
@stef145g could you reproduce this issue?

stef145g:
I have now tested this and it works to some degree but this is not the solution.

[...]

PR #1531, opened 2018-03-31
fix(#1528): translate names ISK, NIO, STD, VEF

stef145g:
Submodule database has been set to branch pre_CryptoCurrencies to enable to create files for testing
I believe these were not translated for a reason but this needs to be tested


src/db/DB_Table_Currencyformats_V1.h
```diff
@@ -173,7 +173,7 @@ struct DB_Table_CURRENCYFORMATS_V1 : public DB_Table
         db->ExecuteUpdate(wxString::Format("INSERT INTO CURRENCYFORMATS_V1 VALUES ('62', '%s', 'L', '', '.', ' ', '', '', '100', '1', 'HNL')", wxTRANSLATE("Honduran lempira")));
         db->ExecuteUpdate(wxString::Format("INSERT INTO CURRENCYFORMATS_V1 VALUES ('63', '%s', 'HK$', '', '.', ' ', '', '', '100', '1', 'HKD')", wxTRANSLATE("Hong Kong dollar")));
         db->ExecuteUpdate(wxString::Format("INSERT INTO CURRENCYFORMATS_V1 VALUES ('64', '%s', 'Ft', '', '.', ' ', '', '', '1', '1', 'HUF')", wxTRANSLATE("Hungarian forint")));
-        db->ExecuteUpdate(wxString::Format("INSERT INTO CURRENCYFORMATS_V1 VALUES ('65', '%s', 'kr', '', '.', ' ', '', '', '1', '1', 'ISK')", L"Icelandic króna"));
+        db->ExecuteUpdate(wxString::Format("INSERT INTO CURRENCYFORMATS_V1 VALUES ('65', '%s', 'kr', '', '.', ' ', '', '', '1', '1', 'ISK')", L"_tr_Icelandic króna"));
```
slodki: `L"_tr_Icelandic króna"` is wrong
Should be `wxTRANSLATE(L"Icelandic króna")` IMHO

vomikan: Both are incorrect.
This trik with '_tr_' prefix is result of bug   #739 fix. I think is better to use only Latin names for currencies.
```wxTRANSLATE("Icelandic krona")```

slodki:
There is no requirement that source files have to be in English.
We set some info about source code before translation at app initialization step (see [wxTranslations::AddCatalog](http://docs.wxwidgets.org/3.0/classwx_translations.html#a7e049aa6b090972e7924392c6f5d6a99) spec).

Maybe we should add UTF-8 as last parameter to [our AddCatalog call](https://github.com/moneymanagerex/moneymanagerex/blob/master/src/mmex.cpp#L72)?

BTW: my proposal to use currency names as used in ISO document should remove all non ASCII chars...

slodki:
Maybe default depends on platform used - UTF-8 can be assumed on Unix, but some CP-xxxx on Windows? I'll try to test this.

src/db/DB_Table_Currencyformats_V1.h
stef145g: Please check out the latest fix as generated by sqlite2cpp.py

vomikan:
@slodki  This PR is actual or outdated?

stef145g:
@slodki  while this was a PR for testing, the changes to **util/sqlite2cpp.py** should be included in your new pr. This provides the fix to allow translations for all fields marked with '_tr_'

slodki:
I will consolidate your and my changes

PR #1544, opened 2018-04-10
fix(warnings): eliminate compile warnings, dead code and typos

slodki:
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1544)

### DIFF
--- a/src/platfdep.h
+++ b/src/platfdep.h
@@ -34,7 +34,6 @@ namespace mmex
     const wxFileName GetSharedDir();
     const wxFileName GetUserDir(bool create);
     const wxFileName GetLogDir(bool create);
-    static const wxFileName GetInstallDir();
 
     /*
         wxStandardPaths uses wxApp::GetAppName(), so you should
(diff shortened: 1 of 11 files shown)


src/platfdep.h
```diff
@@ -34,7 +34,6 @@ namespace mmex
     const wxFileName GetSharedDir();
     const wxFileName GetUserDir(bool create);
     const wxFileName GetLogDir(bool create);
-    static const wxFileName GetInstallDir();
```
stef145g: This line (which has been removed) is required for a successful build.

stef145g:
Checked conflict by switching to fix_warnings branch and rebasing onto master. Conflicting file is not existent in master causing the conflict.

src/platfdep.h
slodki: We have implemented this function in Windows only and it's must be static and not public.

stef145g: I see that fix provided in calling function which corrects the build. Now OK

Issue #1559, opened 2018-04-17
wxDatePickerCtrl oldest date is 1601-01-01


vomikan:
Your OS:
- [X] Windows
- [ ] Mac OSX
- [ ] Linux 
- [ ] Android -> please report bugs to moneymanagerex/android-money-manager-ex repo!

MMEX version:
- [ ] 1.4.x
- [x] development version (git master branch)

vomikan:
https://user-images.githubusercontent.com/3155552/39152278-1e132c4c-4748-11e8-8613-74040e9b0f8a.png

PR #1560, opened 2018-04-17
fix(#1559): oldest date

vomikan:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1560)


slodki:
This 1601 limit exists [only in Windows port](http://docs.wxwidgets.org/3.1/classwx_date_picker_ctrl.html#a5178783537a642100342f01ef6abdbaa):

> Notice that when using a native MSW implementation of this control the lower range is always set, even if SetRange() hadn't been called explicitly, as the native control only supports dates later than year 1601.

Issue #1566, opened 2018-04-19
resetNavTreeControl crashes MMEX


slodki:
1. Open existing database
2. Click open icon from toolbar and select existing encrypted database
3. Enter **wrong** password when prompted
4. Pop-up is displayed...
5. MMEX crashes with following error:
```
10:48:19: Debug: SQLITE_UPDATE
10:48:19: Debug: database: main, table: INFOTABLE_V1, rowid: 22
10:48:19: Debug: =========== navTreeStateToJson =============================
10:48:19: Debug: {
    "item@Home Page": true
}
10:48:19: Debug: SQLITE_UPDATE
10:48:19: Debug: database: main, table: INFOTABLE_V1, rowid: 22
10:48:20: Debug: =========== navTreeStateToJson =============================
10:48:20: Debug: {
    "item@Home Page": true,
    "item@Bank Accounts": true
}
10:48:20: Debug: SQLITE_UPDATE
10:48:20: Debug: database: main, table: INFOTABLE_V1, rowid: 22
terminate called after throwing an instance of 'wxSQLite3Exception'
```

This error happens while executing `resetNavTreeControl()` in [`mmGUIFrame::SetDatabaseFile()`](https://github.com/moneymanagerex/moneymanagerex/blob/master/src/mmframe.cpp#L3046)

BUT this problem not exist in the similar case:
1. Start MMEX
2. Select Open existing database from startup dialog
3. **Cancel** file selection
4. Click open icon from toolbar and select existing encrypted database
5. Enter **wrong** password when prompted
6. Pop-up is displayed...
7. No crash at `resetNavTreeControl()` in [`mmGUIFrame::mmGUIFrame()`](https://github.com/moneymanagerex/moneymanagerex/blob/master/src/mmframe.cpp#L293) (because nav tree was empty?)

[...]

PR #1571, opened 2018-04-19
fix(#1566): typo fix

vomikan:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1571)


src/model/Model_Infotable.cpp
```diff
@@ -70,7 +70,7 @@ void Model_Infotable::Set(const wxString& key, const wxDateTime& date)
 
 void Model_Infotable::Set(const wxString& key, const wxString& value)
 {
-    if (this->db_) return;
```
slodki: Yeah, this caused INFOTABLE was empty in mmb :-(

Issue #1089, opened 2017-03-07
"Custom fields" windows doesn't work


J316:
make sure these boxed are checked before submitting your issue - thank you!
- Check OS
  - [ ] Windows
  - [ ] Mac OSX
  - [X] Linux
-  Check MMEX version
  - [X] 1.3.2
  - [ ] 1.2.x
  - [ ] 1.1 or older

When I click on "Open custom fields window" in a transaction I get a dead window, buttons don't click and even the X to close the window doesn't work (maximize/minimize works), I have to close the transaction I'm editing to close it. OS is manjaro Linux

gabriele-v:
This feature hasn't been tested on Linux\MacOS, maybe it's a wxWidget bug or some coding problem on wxFrame element: in MMEX it's used only here and it's needed to correctly handle modal and focus on side window.

Unfortunately I haven't any Linux machine so I cannot deep debug into the problem.

J316:
[Here](https://manjaro.org/get-manjaro/) you can download it and run it live from usb without installing it if you want

slodki:
The same problem with latest git master on kubuntu zesty.

vomikan:
I suggest to use not separate window for Custom Fields
![image](https://user-images.githubusercontent.com/6836805/37601101-3e80add4-2b9a-11e8-8436-9789128e6928.png)



```
 wxNotebook* trx_notebook = new wxNotebook(this
        , wxID_ANY, wxDefaultPosition, wxDefaultSize, wxNB_MULTILINE);
    wxPanel* notes_tab = new wxPanel(trx_notebook, wxID_ANY);
    trx_notebook->AddPage(notes_tab, _("Notes"));
    wxBoxSizer *notes_sizer = new wxBoxSizer(wxVERTICAL);
    notes_tab->SetSizer(notes_sizer);

    textNotes_ = new wxTextCtrl(notes_tab, ID_DIALOG_TRANS_TEXTNOTES, ""
        , wxDefaultPosition, wxSize(-1, 120), wxTE_MULTILINE);
    notes_sizer->Add(textNotes_, g_flagsExpand);
....
    wxScrolledWindow* custom_tab = new wxScrolledWindow(trx_notebook, wxID_ANY);
    trx_notebook->AddPage(custom_tab, _("Custom"));
    custom_tab->SetMaxSize(wxSize(-1, 120));
    wxBoxSizer *custom_sizer = new wxBoxSizer(wxVERTICAL);
[...]

PR #1580, opened 2018-04-21
fix(#1089) refactoring

vomikan:
image

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1580)


slodki:
Could you show GUI project for custom fields manger/editor?
Will it be possible to display custom field as column in transaction panel and search for user entered value?

vomikan:
@slodki 
1. 
image

2. In future some custom fields should be presented in checking panel. Also a custom data should be searchable.

vomikan:
Merged but still one bug is present. 
I can't fix issue: If transaction is new custom settings will fals to save.

gabriele-v:
@vomikan 
1. Why merge changes that break some functionality (save on new transaction)? Also, I think that changes should be reviewed by someone else, not auto-merged by yourself.

2. Personally I disagree about changes of this pull request, because you have moved all custom fields logic into transaction dialog, increasing the complexity of transdialog (that it's already huge) and specially now you have isolated customfields to transaction only. With previous implementation, custom fields will be ready to be used everywhere in MMEX (like attachments), now they are usable just in transactions.

For that reasons I suggest to revert back this pull request (we just need to make previous implementation working on Linux/Mac), @moneymanagerex/desktop-mmex-team what do you think?

Sorry but for work reasons I cannot be more present/developing, but I still follow MMEX main changes.

PR #1583, opened 2018-04-22
fix(#№1089): save for new transaction

vomikan:

### DIFF
--- a/src/model/Model_CustomField.cpp
+++ b/src/model/Model_CustomField.cpp
@@ -146,12 +146,13 @@ wxString Model_CustomField::getRegEx(const wxString& Properties)
 bool Model_CustomField::getAutocomplete(const wxString& Properties)
 {
     Document json_doc;
-    if (json_doc.Parse(Properties.c_str()).HasParseError())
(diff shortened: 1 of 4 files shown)


src/model/Model_CustomField.cpp
```diff
@@ -146,12 +146,13 @@ wxString Model_CustomField::getRegEx(const wxString& Properties)
...
 {
     Document json_doc;
-    if (json_doc.Parse(Properties.c_str()).HasParseError())
+    if (!json_doc.Parse(Properties.c_str()).HasParseError())
+    {
+        if (json_doc.HasMember("Autocomplete") && json_doc["Autocomplete"].IsBool()) {
+            Value& b = json_doc["Autocomplete"];
+            return b.GetBool();
+        }
         return false;
```
stef145g: Code produces warning:
model_customfield.cpp(157): warning C4715: 'Model_CustomField::getAutocomplete': not all control paths return a value

slodki: There is no `else` for https://github.com/moneymanagerex/moneymanagerex/blob/master/src/model/Model_CustomField.cpp#L149

Issue #1522, opened 2018-03-27
Webapp parser tests needed after to RapidJSON migration


vomikan:
https://github.com/moneymanagerex/moneymanagerex/blob/2d808bb215c0e1836330a283444b66cbf364a49a/src/webapp.cpp#L373

slodki:
binary packages ready to download: https://github.com/slodki/moneymanagerex/releases/tag/test2

vomikan:
Also php server needed.....

gabriele-v:
I've tried to keep at minimum requirements for web app (actual webapp just need php > 5.4 and uploading some files), Laravel one it's harder to setup (needs Php 7.0 and MySql too).

Anyway I've tried binaries from slodki and I confirm that there are some problems with numbers, amount it's not downloaded correctly and transactions are never deleted here:
https://github.com/moneymanagerex/moneymanagerex/blob/2d808bb215c0e1836330a283444b66cbf364a49a/src/webapp.cpp#L647

I'm checking if it's MMEX or WebApp side.

vomikan:
@gabriele-v Hi, nice to see you here.

PR #1586, opened 2018-04-22
fix(#1522): update webapp for RapidJSON and minor improvements

gabriele-v:


gabriele-v:
Agree, but actually WebApp returns the JSON as string and with this commit we can still guarantee WebApp compatibility with all versions of MMEX: if I update WebApp code to format it as integer, then it won't work with old MMEX desktop versions with Cajun.

PR #1588, opened 2018-04-23
fix(#1089): more that 1 multichoice support

vomikan:
image

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1588)

### DIFF
--- a/src/transdialog.cpp
+++ b/src/transdialog.cpp
@@ -46,7 +46,7 @@
 wxIMPLEMENT_DYNAMIC_CLASS(mmTransDialog, wxDialog);
 
 wxBEGIN_EVENT_TABLE(mmTransDialog, wxDialog)
-    EVT_BUTTON(wxID_HOME, mmTransDialog::OnMoreFields)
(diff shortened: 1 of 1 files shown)


gabriele-v:
I think that we should decide how to make it working on Linux and for other entities first, otherwise we could generate conflicts or problem when removing it from transdialog.

vomikan:
I found a way for new GUI looking.
image
Then I'll move custom settings saving from transdialog to separate file.

gabriele-v:
Nice, with the same logic do you think will be possible to move completetely custom fields (not just saving) to a dedicate and common file? Maybe passing the sizer as parameter?

gabriele-v:
To fix problems notices by slodki we need to introduce more code there (Regex validation I.E.), will be nice to have everything not duplicated.

vomikan:
> To fix problems notices by slodki we need to introduce more code there (Regex validation I.E.), will be nice to have everything not duplicated.

I will take care of it.

PR #1594, opened 2018-04-24
fix(#1089): cleanup

vomikan:


gabriele-v:
For custom date field we shouldn't set it to now as default because there is a custom field option (default value) that should be "now" (code to do it was already present, I don't know if still there after migration) to set date default to now. Another option that I would like to add to custom files is "mandatory", so it must have a value for each transaction. In case of date, mandatory will mean different from 01/01/1900. If we automatically put now as default, then how to check for mandatory option? IMHO if it's possible, it will be better to save a "null" when date hasn't been changed. Or, if not possible, save 01/01/1900 and then null it on show.

Issue #1595, opened 2018-04-24
Cannot set custom fields in recurring transaction template


slodki:
Custom field values cannot be set in template. They cannot be set while entering recurring transaction too.

vomikan:
I'll provide improvements here:
https://github.com/moneymanagerex/moneymanagerex/issues/1585

PR #1596, opened 2018-04-25
fix(#1595): separate class for custom data

vomikan:


gabriele-v:
I cannot test it right now, but code seems in the right direction. Need to isolate in another file and fix some problems with database submodule.

PR #1610, opened 2018-04-30
feat(release.alpha): bump version to alpha4

slodki:


gabriele-v:
@moneymanagerex/desktop-mmex-team what do you think about identify major issues that are needed for 1.4.0 release and focus on them, so we can release it quickly?

Issue #1632, opened 2018-05-05
MMEX database upgrade to version 8 failed!


MartinX3:
Your OS:
- [ ] Windows
- [ ] Mac OSX
- [X] Linux 
- [X] Android -> please report bugs to moneymanagerex/android-money-manager-ex repo!

MMEX version:
- [X] 1.4.x Alpha 4
- [ ] 1.3.x
- [ ] 1.2.x
- [ ] 1.1 or older
- [ ] development version (git master branch)

Opened the DB file in the Android App
Then opened in 1.4.x Alpha 4

Error while opening = crash

MMEX database upgrade to version 8 failed!

Please restore DB from autocreated pre-upgrade backup and retry or contact MMEX support

SQL logic error[1]: duplicate column name: CURRENCY_TYPE

PS: I don't need a DB patch. I use a backup solution against such things. :)

gabriele-v:
@MisterY could you check if in Android version your are upgrading correctly the DB based on https://github.com/moneymanagerex/database repository? Seems that we have some problems there..

@MartinX3 Please extract and apply attached patch
[currency_type_patch.zip](https://github.com/moneymanagerex/moneymanagerex/files/1976550/currency_type_patch.zip)

MartinX3:
@gabriele-v 
Thank you for your fast response :)

But I also don't need the patch, I have a backup from before the "Android App Upgrade". :)

I also think, that the Android App needs a new version build.
The stable is from February, the beta is from March

MisterY:
Android should be upgrading natively. No extra effort required. Will try to confirm next week when I get back home.
-- 
Sent from my Android phone with GMX Mail.On 5/5/18, 12:42 Gabriele-V <notifications@github.com> wrote:
@MisterY could you check if in Android version your are upgrading correctly the DB based on https://github.com/moneymanagerex/database repository? Seems that we have some problems there..
@MartinX3 Please extract and apply attached patch
currency_type_patch.zip

—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or mute the thread.


  
  




[...]

Issue #1501, opened 2018-03-24
MMEX blocked by macOS Gatekeeper quarantine


wirmike:
Your OS:
- [ ] Windows
- [x] Mac OSX (High Sierra 10.13.3)
- [ ] Linux 
- [ ] Android -> please report bugs to moneymanagerex/android-money-manager-ex repo!

MMEX version:
- [ ] 1.4.x
- [x] 1.3.x (1.3.3)
- [ ] 1.2.x
- [ ] 1.1 or older
- [ ] development version (git master branch)

This report is probably a mix of two issues.  The original re-occurence
of report #1195, and principally the issue of being unable to reinstall.
The narrative follows.  Unfortunately, events have passed that would have enabled a screen-shot/s to explain. 
The narrative ...
Re the bug/issue I reported on #1195: the ‘bug/issue’ has re-appeared and has proved impossible to correct!
It has to be said that since loading MMEX v1.3.3 everything has been (hitherto) good.
What seems to have happened was that I caught my MacMouse on the edge of my MacBookPro (the top edge at the opposite end to the mouse’s Apple logo.
At that point the Navigation panel ‘flipped’ from the LHS of the screen to the top - between the toolbar and the open accounts file.  Nothing that I did seemed to restore the Panel to its ‘rightful’ position on the left!
It seems that an erroneous 'click/s' on the mouse may have triggered something with the pointer at whatever place it was on the screen - maybe - I may never know, so fleeting a moment it was.
So, I tried to reinstall the App.  Big mistake!
When I tried to launch the result of a reinstall the app reported ““mmex.app” is damaged and can’t be opened.”
No matter how many times I downloaded and retried installing, restarted the machine and ‘trashed’ the newly created apps, they all resulted in an unusable MMEX app.
I then remembered that I had some similar issue when I tried reinstalling - resulting in the error report #1195.

At present I cannot use my MMEX from the Mac and really need some guidance.

HOWEVER, MMEX in the Win7 parallels side of the Mac does work and seems to be able to access the shared files.  I have started using this (now sole route) - I use MMEX REALLY HEAVILY!!!.

I actually FAR prefer the Mac version of MMEX - the Windows one is nice and fine - but for looks and general usability OSX has it.

My Mac is at High Sierra 10.13.3.

I would be grateful for any help.

Regards,

bestlem:
For the crash. Try starting mmex from the terminal.

run 


/Applications/mmex.app/Contents/MacOS/mmex

[...]

PR #1636, opened 2018-05-05
cleanup(db): rename tables and bump db version to v12

slodki:
implementation of moneymanagerex/database#27
Tested and working on Linux (gcc compiler).
There is problem with MSVC string literal limit 16kB. We must store longer sql as `string += string` - python script needs to be updated.

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1636)


gabriele-v:
@slodki 
Seems that v9 and v10 has been changed: this is not allowed, because if someone has updated the DB to previous v10 with nightly builds, it don't receive the upgrade correctly.
We have to be sure that every DB will be upgraded (not for this upgrade only but also futures), if in doubt better to add a new version.
There will be some conflicts with my upgrade too: your it's bigger so I prefer to commit this and then re-base mine, but please fix it soon because I can re-base and fix conflicts for mine just until this evening.
Thanks

gabriele-v:
Also, pragma at the end of the upgrade it's not needed because it's handled in upgrade procedure itself.

gabriele-v:
I've solved the problem on Windows, please find attached py file to generate upgrade file.
I've choose to concatenate directly without += to keep it cleaner.
In this way on Windows we can arrive up to 65kB, it wasn't enough for v9 patch but it could be shorted a lot replacing
`INSERT OR IGNORE INTO CURRENCYFORMATS_V1 (CURRENCYNAME, PFX_SYMBOL, SFX_SYMBOL, DECIMAL_POINT, GROUP_SEPARATOR, SCALE, BASECONVRATE, CURRENCY_SYMBOL, CURRENCY_TYPE) VALUES (`
with
`INSERT OR IGNORE INTO CURRENCYFORMATS_V1 VALUES (null,`

CURRENCYID it's automatically handled as autoincrement with null because is INT PRIMARY KEY 

[sqliteupgrade2cpp.zip](https://github.com/moneymanagerex/moneymanagerex/files/1976866/sqliteupgrade2cpp.zip)

gabriele-v:
FYI I've executed the patch on a 15MB sample DB with 35 Accounts, 1.000 Payees, 1.000 Categories, 6.000 Subcategories, 77.000 Splits, 100.000 Transactions

It takes around 3 secs (including vacuum) on a VMs with 2 Cores of Intel Xeon 1230v2 and SSD Datastore, so performance side we shouldn't have any issue.

gabriele-v:
@slodki : so how many versions you need for your patch? I left you v12 too so you can put in v.11 the new updates (that cannot be inserted in v.9/10 because they are already closed) and v.12 for recreating tables?

slodki:
pragma at the end of sql script will allow to create correct mmb files outside MMEX app - like CI platforms etc. It should stay here.

slodki:
@gabriele-v inserting values without specifying column names:

It's safer to specify column names if we don't know exact column order in the db. Are we sure user have columns in this order in old (possibly broken) database?

slodki:
The oldest db I have (from 0.9.7.0) have all columns (except the last one) so I think we can assume all users should have them in the right order.

gabriele-v:
I was referring IE to update for old currencies that will be useful for everyone.
Anyway it's ok to leave as is, I don't think lot's of people have already upgraded to v10.


For column names, for general upgrades I agree with you, but in that specific upgrade you are creating fresh-new the table:
```
CREATE TABLE CURRENCYFORMATS_V1_NEW
INSERT INTO CURRENCYFORMATS_V1_NEW SELECT
DROP TABLE CURRENCYFORMATS_V1;
ALTER TABLE CURRENCYFORMATS_V1_NEW RENAME TO CURRENCYFORMATS_V1;
```
So that table has that structure for sure, otherwise upgrade have failed in previous steps

gabriele-v:
@slodki for windows, this is still needed (it will increase string length up to 65kB):
https://github.com/moneymanagerex/moneymanagerex/files/1976866/sqliteupgrade2cpp.zip

slodki:
I must wait all required changes for db v11 be implemented to rebase this PR on them
Python script will be upgraded of course.

gabriele-v:
A massive upgrade should be needed here as well: https://github.com/moneymanagerex/general-reports

gabriele-v:
Do we need to wait https://github.com/moneymanagerex/moneymanagerex/issues/1632 fix in Android or we can release? I have doubt that now with nightly available, more people will use unofficial versions

gabriele-v:
Problem is that lots of conflicts will come out if we don't merge now.. We can stop nightly until Android release?

slodki:
Alpha (and nightly) should not be used on production data. These versions are simply "not working" or "work in progress".
Beta should work, users should test it on production data **copy**.

Alpha can be incompatible with Android, but beta should have minimal compatibility (to test data exchange).

slodki:
Every one can compile from sources - we can place red warnings at releases page.
Nightly is hidden from Check for Updated function too.

slodki:
I don't understand "too many changes".

gabriele-v:
Agree on add a warning on release page, but just with "not compatible with Android": Nightly could be not working on UI side, but on DB they should keep it as clean as possible. If not sure about a change, better to leave it as PR. A commit in to master branch should (at least basically) works.

Anyway ok to merge now for me.

slodki:
Nightly is simply not tested software - one or two developers tried to compile and it worked for them at all, but can have serious bugs (crash, deleting data ...) not detected so far.

We have following warnings in the installation instructions already:

> Use unstable pre-release version (alpha, beta, rc) at your risk (work on your database copies!) to check or test newly added functions.

> To try the latest nightly developer build you can go to AppVeyor, click on Win32 or x64 platform then click on ARTIFACTS tab on the right (above blue log window). Please remember this version can not work at all or have serious bugs!

vomikan:
>I don't understand "too many changes".

zbyt wiele zmian

I nean we made a lot of complex changes. But any changes mean new bugs. 
I suggest fix critical bugs then release 1.4.0 beta.

slodki:
Should we create v1.4.x branch and put bigger changes into master for future 1.5.9?

gabriele-v:
If it's not possible to run it on Mac, yes. We have no alternatives..
@omalleypat do you have some free time to looking into it?

For 1.4.x I think for now could stay as the master, considering that we are just 3 active developers with not so much time, we should agree on use our time and focus on bug fixing and not adding new features..

I've seen a lot of issue regarding Linux in 1.4.0 milestone, could you check if still valid and try to fix them if possible?

vomikan:
>Should we create v1.4.x branch and put bigger changes into master for future 1.5.0?

Please, no. 1.4.0 should be master. Any new features may be put into 1.5.x branch. In future 1.5.x should be mergrd into master.

xraynaud:
Hi, maybe I can help for Mac. I am pulling from master regularly and compilling everything. At the moment the latest version seems to work but I have a problem with upgrade to db12.

gabriele-v:
Which error do you have?

slodki:
@xraynaud Could you verify [issues tagged with 'mac'](https://github.com/moneymanagerex/moneymanagerex/labels/mac)?
Is the app signing the only way to solve #1501?

xraynaud:
Here it is

MMEX database upgrade to version 12 failed!
Please restore DB from autocreated pre-upgrade backup and retry or contact MMEX support
constraint failed[2067]: UNIQUE constraint failed: PAYEE.PAYEENAME

xraynaud:
@slodki: I am not sure I have encountered something similar with #1501. Generaly, when I download and app from a unknown developper (according to apple), OSX refuse to start it. To force the system to start the app, you have to go to System preference/Security. There, under the "general" tab, there is a button that appear when there is an application that as been blocked. You click on the buttom and the program starts. I'm not sure the problem described in #1501 could have been solved like this.

@gabriele-v : for some reason, it does not work. I have removed all duplicates (I check that they were all removed by opening the db using a software to open a sqlite db) but I still get the same error.

Issue #1585, opened 2018-04-22
Custom Fields improvements


gabriele-v:
New *Custom Fields Manager* implementation is working under wxgtk:
- [x] It has "Personalize custom fields | Transaction" title - I don't understand transaction word here
- [x] `OK` and `Cancel` buttons should be replaced with `Close` button - there is no way to accept or not changes made
- [x] First field has *Description* name - is this description (string describing usage etc.) really or name (short id)?
- [x] *Default* field can has value not matching field type - this should be validated
- [x] *RegEx* can be used to validate numbers too
- [x] *Default* is not reactivated after type change
- [x] *Choices* accept duplicated entries like "a;a;a;a" (?)
- [x] *Massive Content Update* - unclear function description, user is asked questions without meaning steps executed
- [x] *Description* accepts existing predefined column names (like *Payee*) and duplicates (i.e. two custom fields with the same *Description*)
- [x] Add/Edit/Remove actions should be moved to buttons in the same line as `OK` from gear button - these actions are main functionalities of this dialog
- [x] Time type is useless IMHO - there should be *Date* and *Date&Time* types. Please imagine multiple transactions from last year with "9:00" or "12:00" value - how to sort them? How to build filter rule? Field allowing storing morning/evening/night/... info is useless for me.
- [x] "true" value entered ad *Default* for boolean is not working

Transaction dialog test results:
- [x] RegEx validation is not working
- [x] multiple with duplicates results in strange behavior
- [x] there is arbitrary precision forced for decimals (2 digits)
- [x] Time custom field is broken - I can set only hour part but this changes transaction date (!)
- [x] Date type custom field with no default value defined is rendered with value "1900-01-01" when empty. We must force some default value (like "now") or render this field as empty and allow to store NULL value.
    Now new transaction with date custom field not edited stores "1900-01-01" value in database file.

Others features needed:
- [ ] display them in transaction panel, filter and search using their values and see them in reports
- [ ] support for importing/exporting values from/to csv/qif files to custom fields
- [ ] controls (like customised transaction grid, predefined filter presets, reports...) must validate custom field used still exists and have correct type and not crash after user delete/edit custom field definition
- [ ] ability to set custom field values in recurring transaction templates and confirm/modify while entering recurring transaction (with custom field existence at this time)

vomikan:
> It has "Personalize custom fields | Transaction" title - I don't understand transaction word here

[...]

PR #1641, opened 2018-05-07
fix(#1585): class for filter

vomikan:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1641)


vomikan:
@guanlisheng is expert regarding OOP.

Issue #1530, opened 2018-03-31
Download Currency Values history button creates entries for non-existing currency


slodki:
1. Create empty database and select USD as default currency
2. Open Currency Dialog (`Tools` -> `Organize Currency...`)
3. Define new currency using `Add` button with *aaa* name and *aaa* symbol
4. Select this new currency and press `Download Currency Values history` button on the right
5. Answer `Yes` when asked to add dates without any transaction
6. Multiple history entries are created with 1.0 value

Expected behavior:

Non-existing currency ISO symbol should be detected (based on on-line response?) and no incorrect entries should be created.

vomikan:
I am not sure when this https://github.com/moneymanagerex/database/blob/master/database_version_8.sql
should be implemented.

slodki:
"Non-existing" => "not existing on queried on-line server, when no response received" => "with unknown rate"...

These entries all wrong an user mus delete them or manually correct.

vomikan:
I can't reproduce this issue.
For non exiting currency Yahoo response is:
```
{
  "chart": {
    "result": null,
    "error": {
      "code": "Not Found",
      "description": "No data found, symbol may be delisted"
    }
  }
}
```
then in maincurrencydialog.cpp:869
```
    wxASSERT(chart.HasMember("error"));
    if (!chart.HasMember("error") || !chart["error"].IsNull())
        return false;
```

vomikan:
But in case if base currency not USD there is no error and Yahoo returns rates = 1
```
{
  "chart": {
    "result": [
      {
        "meta": {
          "currency": "EUR",
          "symbol": "AAAEUR=X",
          "exchangeName": "CCY",
          "instrumentType": "CURRENCY",
          "firstTradeDate": 1070236800,
          "gmtoffset": 3600,
          "timezone": "BST",
          "exchangeTimezoneName": "Europe/London",
          "chartPreviousClose": 1,
          "currentTradingPeriod": {
            "pre": {
              "timezone": "BST",
              "end": 1522969200,
              "start": 1522969200,
              "gmtoffset": 3600
            },
            "regular": {
              "timezone": "BST",
              "end": 1523055540,
              "start": 1522969200,
              "gmtoffset": 3600
            },
            "post": {
              "timezone": "BST",
              "end": 1523055540,
              "start": 1523055540,
              "gmtoffset": 3600
            }
          },
          "dataGranularity": "1d",
          "validRanges": [
            "1d",
            "5d",
[...]

Issue #1285, opened 2017-08-11
Budget plan with multiple currencies


MarekLe:
I found a BUG in budgeting window. I have some accounts in PLN (main currency) and some in CHF. 'Budget setup' window summarizing values in PLN and CHF without any convertion.
![obraz](https://user-images.githubusercontent.com/30925920/29209282-f94210b4-7e8d-11e7-9b3a-3d3e42e89072.png)

In this situation budgeting in MMEX is dificult to use... It's wrong... and confusing.

Your OS:
- [x] Windows
- [ ] Mac OSX
- [ ] Linux 
- [ ] Android -> please report bugs to moneymanagerex/android-money-manager-ex repo!

MMEX version:
- [ ] 1.4.x
- [x] 1.3.x
- [ ] 1.2.x
- [ ] 1.1 or older
- [ ] development version (git master branch)

vomikan:
I think all values are in base currency. Just update currency rates in a Currency Manager.

slodki:
Duplicate of #1183 

This is not working as you can see below. Default 1:1 ratio is used even if history is updated.

![mmex1183bud](https://user-images.githubusercontent.com/3155552/37742124-8567dc92-2d64-11e8-9547-913fb7b69d42.png)

Issue #1183, opened 2017-06-06
Currency history ignored in transaction report and home page


keinejiang:
- Check OS
  - [x] Windows
  - [ ] Mac OSX
  - [ ] Linux 
- Check MMEX version
  - [x] 1.3.x
  - [ ] 1.2.x
  - [ ] 1.1 or older

Calculate currency twice. Steps:
1. Create a database using a currency, eg. USD;
2. Create a bank account in the database using another currency, eg, HKD;
3. Create a new withdrawal transaction under the account, eg, HK$1000;
4. Update currency online;
5. Use report tool "Transaction Report" to export it by using "type=withdrawal";

Now, you can see that the "Amount" in report have error value.

slodki:
Only initial value (first stored in db) for currency value is used for transaction.
Next updates for currency values (online or manual) are ignored in existing or newly created transactions.

<hr><small><b>Version</b>: 1.4.0-Alpha.1 ⬥ db 7 ⬥ git 7e43643 (2017-07-24)<br><b>Libs</b>: wxWidgets 3.1.0 (wxMSW 6.3) ⬥ wxSQLite3 3.4.0 (SQLite 3.14.1) ⬥ Mongoose 6.5 ⬥ Lua 5.2.4<br><b>Build</b>: Jul 24 2017 20:28:26 ⬥ Microsoft Visual Studio 1900 ⬥ CMake 3.8.2 ⬥ GNU Make <br><b>OS</b>: Windows 8.1 (build 9600), 64-bit edition Little endian ⬥ Polish windows-1250 ⬥ 1680x1050x32bit 98x98ppi</small>

vomikan:
I can't reproduce this in 1.4.0

![image](https://user-images.githubusercontent.com/6836805/37603369-74c0e148-2b9f-11e8-829f-d5ccb59ee583.png)

slodki:
130 zł should be reported here:

![mmex1183cur](https://user-images.githubusercontent.com/3155552/37705792-73bd3684-2cfd-11e8-8a21-8532c3f9f021.png)
![mmex1183rep](https://user-images.githubusercontent.com/3155552/37705796-77440bb6-2cfd-11e8-9714-b491f8437471.png)

slodki:
We should remove Conversion to Base Rate from currency IMHO. Then use 1:1 ratio if history is empty. And auto-generate last entry in history window with *before* and *1* values.

User can define base rate as old history entry.

[...]

Issue #646, opened 2016-01-11
Home Page Stock Totals are Not Correct


vomikan:
On the home page under stocks, the total dollar amounts are not correct. The individual accounts and reports are, but all total account amounts do not add up.

https://sourceforge.net/p/moneymanagerex/bugs/795/

vomikan:
Investigating this problem, I found the following problems:
- The same issue for all accounts not only investments;
- In the table CURRENCIES_V1 all the values BASECONVRATE = 1 but should be equal the currency rate;

stef145g:
#805 Grand Total missing stocks on 1.2.5
Ref: http://sourceforge.net/p/moneymanagerex/bugs/805/
Fix committed to 1.2.x

vomikan:
It's referencing with this issue https://github.com/moneymanagerex/moneymanagerex/issues/647

slodki:
Still not working in 1.4.0-alpha.4

slodki:
#1183

gabriele-v:
Will be fixed with https://github.com/moneymanagerex/moneymanagerex/issues/1541

PR #1644, opened 2018-05-08
fix(baseconvrate): Remove and use only currencyhistory

### LINKED ISSUE #1541: Remove BASECONVRATE from CURRENCYFORMATS table
Now we have 2 overlapping implementations of currency rates: one simple inside CURRENCYFORMATS and more sophisticated in dedicated CURRENCYHISTORY table.

My proposal:
* use only CURRENCYHISTORY records
* remove BASECONVRATE field from CURRENCYFORMATS
* do not store meaningless "1" default value in database
* implement following logic to maintain compatibility with current assumptions:
```c
double find_currency_rate_for_given_date(currency curr, date d) {
  if (currencyhistory(curr) is empty || d<oldest in currencyhistory(curr))
    return 1 /* no conversion done if rate is unknown */
  else iterate through currencyhistory(curr) records...
}
```

This should help with #1158, #1183, #1537, improve data model design (removes redundancy) and improve UX.

Problems:
1. Migration from old model for currencies with no CURRENCYHISTORY and with BASECONVRATE <> 1 - there is not enough details available to migrate to new structure. Proposals: a) migrate user's value adding "1900-01-01" date to it b) not migrate BASECONVRATE value to CURRENCYHISTORY
2. Migration from old model BASECONVRATE not matching CURRENCYHISTORY values. Proposals: same as above.

gabriele-v:
Fixes #1541, #646, #1183, #1285, #1530

Requires https://github.com/moneymanagerex/database/pull/30

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1644)

### DIFF
--- a/src/maincurrencydialog.cpp
+++ b/src/maincurrencydialog.cpp
@@ -188,7 +188,7 @@ void mmMainCurrencyDialog::CreateControls()
         , wxALIGN_LEFT, wxDATAVIEW_COL_SORTABLE);
     currencyListBox_->AppendTextColumn(ColName_[CURR_NAME], wxDATAVIEW_CELL_INERT, 140
         , wxALIGN_LEFT, wxDATAVIEW_COL_SORTABLE);
-    currencyListBox_->AppendTextColumn(ColName_[BASE_RATE], wxDATAVIEW_CELL_EDITABLE, 80);
+    currencyListBox_->AppendTextColumn(ColName_[BASE_RATE], wxDATAVIEW_CELL_INERT, 80);
     currencyListBox_->AppendToggleColumn(ColName_[CURR_HIST], wxDATAVIEW_CELL_INERT, 30);
 
     itemBoxSizer3->Add(currencyListBox_, g_flagsExpand);


src/db/DB_Table_Currencyformats.h
```diff
@@ -109,285 +109,285 @@ struct DB_Table_CURRENCYFORMATS : public DB_Table
...
+        db->ExecuteUpdate(wxString::Format("INSERT INTO CURRENCYFORMATS VALUES ('1', '%s', '$', '', '.', ' ', '100', 'USD', 'Fiat', '0')", wxTRANSLATE("US Dollar")));
+        db->ExecuteUpdate(wxString::Format("INSERT INTO CURRENCYFORMATS VALUES ('2', '%s', '%s', '', '.', ' ', '100', 'EUR', 'Fiat', '0')", wxTRANSLATE("Euro"), L"€"));
+        db->ExecuteUpdate(wxString::Format("INSERT INTO CURRENCYFORMATS VALUES ('3', '%s', '%s', '', '.', ' ', '100', 'GBP', 'Fiat', '0')", wxTRANSLATE("Pound Sterling"), L"£"));
+        db->ExecuteUpdate(wxString::Format("INSERT INTO CURRENCYFORMATS VALUES ('4', '%s', '', '%s', ',', ' ', '100', 'RUB', 'Fiat', '0')", wxTRANSLATE("Russian Ruble"), L"р"));
+        db->ExecuteUpdate(wxString::Format("INSERT INTO CURRENCYFORMATS VALUES ('5', '%s', '%s', '', ',', ' ', '100', 'UAH', 'Fiat', '0')", wxTRANSLATE("Hryvnia"), L"₴"));
+        db->ExecuteUpdate(wxString::Format("INSERT INTO CURRENCYFORMATS VALUES ('6', '%s', '%s', '', '.', ' ', '100', 'AFN', 'Fiat', '0')", wxTRANSLATE("Afghani"), L"؋"));
```
slodki: here is ok

src/maincurrencydialog.cpp
```diff
@@ -939,11 +929,14 @@ bool mmMainCurrencyDialog::GetOnlineHistory(std::map<wxDateTime, double> &histor
         if (quotes_closed[i].IsFloat())
         {
             double rate = quotes_closed[i].GetFloat();
-            history[time] = rate;
-            if (first_date > time)
+            if (rate != 1) // Skip rates = 1 (Yahoo returns 1 with invalid Symbols)
```
slodki: What with following downloaded history values:
1.11
1.09
1.03
1.00
0.98
0.99
1.00
1.01
?
Two of them ("1.00") will be missing?

src/mmhomepagepanel.cpp
```diff
@@ -131,24 +131,26 @@ void htmlWidgetStocks::calculate_stats(std::map<int, std::pair<double, double> >
...
-        double conv_rate = 1;
+        double today_conv_rate = 1;
+        double purchase_conv_rate = 1;
         Model_Account::Data *account = Model_Account::instance().get(stock.HELDAT);
         if (account)
-        {   
-            Model_Currency::Data *currency = Model_Currency::instance().get(account->CURRENCYID);
-            conv_rate = currency->BASECONVRATE;
+        {
+            today_conv_rate = Model_CurrencyHistory::getDayRate(account->CURRENCYID, today);
```
slodki: We could call just `getDayRate(account->CURRENCYID)` here...

src/model/Model_CurrencyHistory.cpp
```diff
@@ -85,41 +85,68 @@ int Model_CurrencyHistory::addUpdate(const int& currencyID, const wxDate& date,
 /** Return the rate for a specific currency in a specific day*/
 double Model_CurrencyHistory::getDayRate(const int& currencyID, const wxString& DateISO)
```
slodki: we should add here overloaded version to simplify code:
`double Model_CurrencyHistory::getDayRate(const int& currencyID, const wxDateTime& Date = wxDateTime::Today())`

Then remove all `const wxString today = wxDateTime::Today().FormatISODate();` and simply call `getDayRate(currencyID)`

src/model/Model_CurrencyHistory.cpp
```diff
@@ -85,41 +85,68 @@ int Model_CurrencyHistory::addUpdate(const int& currencyID, const wxDate& date,
...
+        {
+            dFuture.ParseDate(rsFuture.GetAsString(0));
+        }
+
+        if (dPast.IsValid() && dFuture.IsValid())
+        {
+            const wxTimeSpan spanPast = Date.Subtract(dPast);
+            const wxTimeSpan spanFuture = dFuture.Subtract(Date);
+
+            dNearest = spanPast <= spanFuture ? dPast : dFuture;
```
slodki: This should be ok for currencies - they should be continuous and we can do linear interpolation here.
But for shares (not updated every day) it should be better to use dPast always.

src/model/Model_CurrencyHistory.cpp
```diff
@@ -85,41 +85,68 @@ int Model_CurrencyHistory::addUpdate(const int& currencyID, const wxDate& date,
...
+        {
+            dNearest = dPast;
+        }
+        else if (dFuture.IsValid())
+        {
+            dNearest = dFuture;
+        }
+        else
+        {
+            //TODO: Show warning alert but only one time?
```
slodki: one time, but per MMEX session. User should be warned every time he runs MMEX that displayed values are wrong because rates are missing.
With buttons: `Download now`, `manually add missing` and `continue [with errors]`

src/model/Model_CurrencyHistory.cpp
```diff
@@ -85,41 +85,68 @@ int Model_CurrencyHistory::addUpdate(const int& currencyID, const wxDate& date,
 /** Return the rate for a specific currency in a specific day*/
 double Model_CurrencyHistory::getDayRate(const int& currencyID, const wxString& DateISO)
 {
-    if (currencyID == Model_Currency::GetBaseCurrency()->CURRENCYID)
+    if (currencyID == Model_Currency::GetBaseCurrency()->CURRENCYID || currencyID == -1)
         return 1;
     
     wxDateTime Date;
     Date.ParseDate(DateISO);
```
slodki: What we do if DateISO>today?
Can user enter predicted/expected rates into Currency History?
Is it better to use today+1 value then real today-2 one?

src/model/Model_CurrencyHistory.cpp
```diff
@@ -85,41 +85,68 @@ int Model_CurrencyHistory::addUpdate(const int& currencyID, const wxDate& date,
...
-        //}
-        Model_Currency::Data* Currency = Model_Currency::instance().get(currencyID);
-        return Currency->BASECONVRATE;
+        return 1;
     }
 }
 
-/** Return the last attachment number linked to a specific object */
+/** Return the last currency rate */
 double Model_CurrencyHistory::getLastRate(const int& currencyID)
```
slodki: Should we use `date<today` filtering here?
Should we return *last in table* or *last known* value here?

src/model/Model_CurrencyHistory.cpp
```diff
@@ -130,7 +157,7 @@ double Model_CurrencyHistory::getLastRate(const int& currencyID)
     else
     {
         Model_Currency::Data* Currency = Model_Currency::instance().get(currencyID);
```
slodki: unused variable?

src/reports/cashflow.cpp
```diff
@@ -69,17 +70,20 @@ void mmReportCashFlow::getStats(double& tInitialBalance, std::vector<ValueTrio>&
             if (wxNOT_FOUND == accountArray_->Index(account.ACCOUNTNAME)) continue;
         }
 
-        const Model_Currency::Data* currency = Model_Account::currency(account);
-        tInitialBalance += account.INITIALBAL * currency->BASECONVRATE;
+        // Use account first transaction date for initial balance
```
slodki: this is wrong. I have old account with initial amount not used for years - we should add account start date (date for initial amount) to use correct rate to calculate initial amount value.

src/db/DB_Table_Currencyformats.h
gabriele-v: Maybe is something related to repository autoencoding?

slodki:
Please add `double Model_CurrencyHistory::getDayRate(const int& currencyID, const wxDateTime& Date = wxDateTime::Today())`
then call it from `double Model_CurrencyHistory::getDayRate(const int& currencyID, const wxString& DateISO)`
This should be trivial and will allow simplify calling it in above PR.

src/model/Model_CurrencyHistory.cpp
gabriele-v: Agree

src/maincurrencydialog.cpp
gabriele-v: I think it's nearly impossibile that two currencies has exact 1:1 conversion rate

vomikan:
Just reminder. wxDateTime::Today() is very slow function.

gabriele-v:
In that case better to leave as is, I've already put every wxDateTime::Today() as const outside for loops looking for currency rates

src/maincurrencydialog.cpp
```diff
@@ -663,8 +660,7 @@ void mmMainCurrencyDialog::OnHistoryUpdate(wxCommandEvent& /*event*/)
         {
             for (auto &entry : historical_rates)
             {
-                Model_CurrencyHistory::instance().addUpdate(m_currency_id
-                    , entry.first, entry.second, Model_CurrencyHistory::ONLINE);
+                Model_CurrencyHistory::instance().addUpdate(m_currency_id, entry.first, entry.second, Model_CurrencyHistory::ONLINE);
```
vomikan: Too long lines is bad. It decrease a code redability.

gabriele-v: Up to me it's the opposite: I don't like code that could stay in one line and instead it's divided into two/three. Generally I try to keep it inline if it stays in one line on my FullHD monitor.

src/model/Model_CurrencyHistory.cpp
gabriele-v: Actually it's used only in maincurrencydialog to see rates in the left panel, so it's not important, but I think should be last known

gabriele-v: Problem is that it will slow down due to  wxDateTime::Today() slowness

src/model/Model_CurrencyHistory.cpp
gabriele-v: What do you mean for shares? If you refer to shares value, than there is StockHistory, if you refer to conversion of shares value, then should be fine because it will try to find nearest conversion for the date of share value, isn't it?

src/maincurrencydialog.cpp
gabriele-v: Screens are increasing in resolution, if you keep on one line code that could stay on one line, then you need less scrolling to see all the cpp code

src/maincurrencydialog.cpp
slodki: There are technical currency clones (EUR vs WIR EUR etc.) and some currencies can have similar rates for political reasons. I can imagine ratio oscillating around 1:1

src/model/Model_CurrencyHistory.cpp
slodki: Yes, I'm writing about StockHistory here, for the future...

src/model/Model_CurrencyHistory.cpp
slodki: OK, I will open separate issue for this (to warn user displayed values are pointless)

src/model/Model_CurrencyHistory.cpp
slodki: I'm thinking about blocking entering future values

src/model/Model_CurrencyHistory.cpp
slodki: We can add `getLastRate(const int& currencyID, const wxDateTime& date)` variant for this and call `wxDateTime::Today()` inside maincurrencydialog once

src/reports/cashflow.cpp
slodki: This can be separate PR as it require changes in account table, but we should think about it.

slodki:
`double Model_CurrencyHistory::getDayRate(const int& currencyID, const wxDateTime& Date = wxDateTime::Today())` will not make any additional Today() call if we pass existing value (if we call getDayRate() from loop) or with default (when called not from loop).
This will remove double conversion wxString<->wxDateTime using now. And we still can use wxString variant for values read from sqlite.
This will simplify and speed up our code only.

vomikan:
There is other issue FYI.
Now().OnlyDate() may has hours = timezone. Then if you try compare wxDateTime values they may be not eqvivslent when should be.

Also date is like int type. It's no reasons to use pointer like const wxDateTime& Date

slodki:
@vomikan There is no issue regarding hours. The first step inside getDatRate is to use wxDateTime. My patch change nothing but removes unnecessary wxDateTime->wxString->wxDateTime conversions.

wxDateTime is object instance, duplicating object instance to use it as const is unnecessary.

slodki:
Yahoo response will be skipped only if all values are equal 1

Issue #1659, opened 2018-05-12
Optimize headers/include files to improve build process


slodki:
1. Remove "include all possible headers" files from mmex sources
2. Remove orphan includes left after code refactoring

Tools like [iwyu](https://github.com/include-what-you-use/include-what-you-use) or [cppclean](https://github.com/myint/cppclean) (this is only for .h header files) and be used to automate this and verify at CI platform at every commit.

iwyu cannot be used with MMEX because of include-what-you-use/include-what-you-use#424 now and we should wait for include-what-you-use/include-what-you-use#529.
Including iwyu into MMEX build process will be as easy as adding [`CPP_INCLUDE_WHAT_YOU_USE`](https://cmake.org/cmake/help/latest/prop_tgt/LANG_INCLUDE_WHAT_YOU_USE.html#prop_tgt:%3CLANG%3E_INCLUDE_WHAT_YOU_USE) to mmex target in CMakeLists.txt.

slodki:
Example log from iwyu compile run (**without --tolerate_transitive** it recommends low-level .h file always):

```
[ 33%] Building CXX object src/CMakeFiles/MMEXLib.dir/aboutdialog.cpp.o
Warning: include-what-you-use reported diagnostics:

/home/slodki/src/moneymanagerex/src/aboutdialog.h should add these lines:
#include "wx/defs.h"      // for wxCLANG_IF_VALID_WARNING_1
#include "wx/event.h"     // for wxCommandEvent (ptr only), wxDECLARE_EVENT_...
#include "wx/gdicmn.h"    // for wxPoint, wxSize (ptr only)
#include "wx/object.h"    // for wxDECLARE_DYNAMIC_CLASS, wxObject (ptr only)
#include "wx/string.h"    // for wxString
#include "wx/windowid.h"  // for wxWindowID
class wxHtmlLinkEvent;
class wxWindow;

/home/slodki/src/moneymanagerex/src/aboutdialog.h should remove these lines:
- #include <wx/notebook.h>  // lines 23-23
- #include "defs.h"  // lines 24-24

The full include-list for /home/slodki/src/moneymanagerex/src/aboutdialog.h:
#include <wx/dialog.h>    // for wxDialog
#include "wx/defs.h"      // for wxCLANG_IF_VALID_WARNING_1
#include "wx/event.h"     // for wxCommandEvent (ptr only), wxDECLARE_EVENT_...
[...]

Issue #755, opened 2016-03-11
db files structure


vomikan:
@guanlisheng 
Could you refactor (split)
DB_Table_Currencyformats_V1.h to h and cpp files
DB_Table_Currencyformats_V1.cpp (may be rename it to db_table_currencyformats_v1.cpp)

```
void DB_Table_CURRENCYFORMATS_V1::ensure_data(wxSQLite3Database* db)
{
     .... 
}

```

DB_Table_Currencyformats_V1.h

```
void ensure_data(wxSQLite3Database* db);
```

guanlisheng:
i wanted to understand why we need this change ?

vomikan:
If any changes has done in header file it takes a lot of time to compile a project.
No any other reasons.
You can close the issue if the alteration takes a lot of time.

stef145g:
I totally agree with @vomikan 

@guanlisheng  to answer your question:
> i wanted to understand why we need this change ?

1. When a .cpp file is compiled, all .h files referenced are also compiled. This means that if a .h file is referenced by 20 files, the same .h file is compiled 20 times. The bigger it is, the longer it takes.

2. When the code is contained in a .cpp file, the .h file is still referenced but the actual code is only compiled once. This reduces the actual overall compile time.

3. All .h files should be used as an interface to the actual code. Any change to the actual code, become localised, and keeps the code "private".

4. In MSVC at least, I can compile a single .cpp file to check for syntax and code errors. I cannot do this when the code is placed in a .h file.

I believe that the general rule is: **Keep .h files small and place all code in .cpp files**.

guanlisheng:
i think it's not the time yet.

guanlisheng:
will close it soon if there is no objection ...

slodki:
http://www.hwaci.com/sw/mkhdr/makeheaders.html

PR #1666, opened 2018-05-14
Fix headers

slodki:
This is partial implementation of #1659.

It eliminates unnecessary dependencies not touching MMEX code.
I moved all includes required for implementation only to cpp files from headers.
Replaced includes with forward class declarations where possible in header files

This reverts 9a8591e4438cd46c0c9aa66cff57a6d2f7c0616c by removing allreport.h and allmodel.h.

All `#include "model/Model_xxx.h"` replaced with simpler `#include "Model_xxx.h"`

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1666)


slodki:
#755 is in the queue. Idea of using automatic checking using static analysis tools is described in #1659.

PR #1702, opened 2018-06-04
UX improvement for Relocate Payee Dialog

### LINKED ISSUE #1696: UX improvement for Relocate Payee Dialog
- [x] Add `Delete source payee` checkbox and delete unused payee after successful operation
- [x] Do not close dialog after operation - this will allow multiple operations in a row
- [x] Hide unused payees from source selection combo - selecting unused payee disables `OK` button

slodki:
this fixes #1696
mmex1

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1702)

### DIFF
--- a/src/relocatepayeedialog.cpp
+++ b/src/relocatepayeedialog.cpp
@@ -80,7 +80,8 @@ void relocatePayeeDialog::CreateControls()
     wxStaticLine* lineTop = new wxStaticLine(this,wxID_STATIC
         , wxDefaultPosition, wxDefaultSize, wxLI_HORIZONTAL);
 
-    cbSourcePayee_ = new wxComboBox(this, wxID_BOTTOM, ""
+    cbSourcePayee_ = new wxComboBox(this, wxID_BOTTOM
+        , sourcePayeeID_ == -1 ? "" : Model_Payee::get_payee_name(sourcePayeeID_)
         , wxDefaultPosition, btnSize
         , Model_Payee::instance().used_payee_names());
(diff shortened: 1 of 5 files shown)


src/relocatepayeedialog.cpp
```diff
@@ -80,12 +84,16 @@ void relocatePayeeDialog::CreateControls()
...
         , wxDefaultPosition, wxDefaultSize, wxLI_HORIZONTAL);
 
-    cbSourcePayee_ = new wxComboBox(this, wxID_BOTTOM, ""
+    cbSourcePayee_ = new wxComboBox(this, wxID_BOTTOM
+        , sourcePayeeID_ == -1 ? "" : Model_Payee::get_payee_name(sourcePayeeID_)
         , wxDefaultPosition, btnSize
-        , Model_Payee::instance().all_payee_names());
-    cbSourcePayee_->AutoComplete(Model_Payee::instance().all_payee_names());
+        , Model_Payee::instance().all_payee_names(true));
+    cbSourcePayee_->AutoComplete(Model_Payee::instance().all_payee_names(true));
```
vomikan: for example, all_payee_names(true) - what does true mean? (I know that info available in the function header).
It bad style of progtamming to use Boolean parameters. 
May be better used_payee_names() ?

slodki: The problem is not boolean parameter, but `all_*_*(*_only)` - all should not have any filtering.
This method is quite simple so it can be duplicated.

PR #1803, opened 2018-09-05
Clean sources from unused code and fix cmake (wxwidgets part)

slodki:
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1803)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -225,14 +225,7 @@ ENDIF()
 
 # --------- wxWidgets dependency ---------
 SET(wxWidgets_USE_REL_AND_DBG ON)
-IF(WIN32)
-    #wxScintilla is needed on Windows
-    FIND_PACKAGE(wxWidgets REQUIRED
-        core qa net html xml aui adv stc webview base scintilla REQUIRED)
-ELSE()
-    FIND_PACKAGE(wxWidgets REQUIRED
-        core qa net html xml aui adv stc webview base REQUIRED)
-ENDIF()
+FIND_PACKAGE(wxWidgets REQUIRED core qa html xml aui adv stc webview base REQUIRED)
 IF(wxWidgets_VERSION_STRING AND wxWidgets_VERSION_STRING VERSION_LESS 2.9.2)
     MESSAGE(SEND_ERROR "wxWidgets version to old.")
 ENDIF()
@@ -349,7 +342,6 @@ IF(WIN32)
         IF(NOT CMAKE_INSTALL_DEBUG_LIBRARIES_ONLY)
             LIST(APPEND MMEX_WXDLLS
                 wxbase${WXDLLPREF}
-                wxbase${WXDLLPREF}_net
                 wxbase${WXDLLPREF}_xml
                 wxmsw${WXDLLPREF}_adv
                 wxmsw${WXDLLPREF}_aui
@@ -363,7 +355,6 @@ IF(WIN32)
             SET(WXDLLPREF ${WXDLLPREF}d)
             LIST(APPEND MMEX_WXDLLS
                 wxbase${WXDLLPREF}
-                wxbase${WXDLLPREF}_net
                 wxbase${WXDLLPREF}_xml
                 wxmsw${WXDLLPREF}_adv
                 wxmsw${WXDLLPREF}_aui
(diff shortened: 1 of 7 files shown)


grumpy235:
@slodki 
Applied this change and got a whole bunch of  Scintilla linkage errors out of wxmsw31u when building on Windows 10 and VS2017. 
My Widgets modules are at 3.1.1 level and compiled months ago with standard MSW .sln file delivered with wxwidgets using VS2017.
Looks like removing the Scintilla dependancy for WIN32 has caused my pain.
I assume Scintilla is not required by Moneymanager so a tailored build of Wxwidgets is required - my guess is setting STC=0 should do the trick ?

slodki:
As you can see in AppVeyor logs there are no linking problems when using precompiled binaries.

CMake module for wxWidgets does not require to specify scintilla library as it should be added internally:
https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/FindwxWidgets.cmake#L242
https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/FindwxWidgets.cmake#L321
like other windows libs:
https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/FindwxWidgets.cmake#L201

slodki:
stc wxw library is used in MMEX as it delivers `wxStyledTextCtrl`, the base of minimal_editor and GRM.
Read https://github.com/wxWidgets/wxWidgets/blob/master/src/stc/README.txt about stc and scintilla relations.

grumpy235:
After this fix we have...
**FIND_PACKAGE(wxWidgets REQUIRED core qa html xml aui adv stc webview base REQUIRED)**
which does not contain "net" as expected but it does not contain "scintilla" either
Was this intended?
Previously "scintilla" was included for a WIN32 build but for some reason not for all other builds.

slodki:
see comment above: `scintilla` is added by CMake if running under windows...

grumpy235:
That may be the intention, however adding "scintilla" to the FIND_PACKAGE statement resolves my linkage issue and I now have a MMEX module. 
For some reason when I try and execute the new module I get "LIBCURL.DLL not found". I did upgrade to curl-7.61.1 from 7.61.0 and curl-7.61.1/bin does contain libcurl.dll but curl-7.61.0/bin does not - need to do some digging and will start by falling back to 7.61.0
Versions reported by the build are...
-- Versions
-- ========
-- Windows 10.0.17134
-- CMake 3.11.18040201-MSVC_2
-- MSBuild 15.8.168.64424
-- Visual Studio 15.0
-- MSVC 19.15.26726.0
-- Win SDK 10.0.17134.0
-- wxWidgets 3.1.1
-- wxSQLite3 4.0.4
-- Lua 5.3.4
-- curl 7.61.1
-- gettext 0.19.8.1

and I have just noticed that CMake 3.12.1 is now available so will try with that although the version delivered with VS2017 is still 3.11
Last good MMEX compile was after commit 15979FA - update pl translation.

grumpy235:
Ran the "cmake -G "Visual Studio 15 2017" -DCMAKE_PREFIX_PATH=c:\libcurl .." with CMake 3.12.1 and fell back to curl 7.61.0 but still got the same linkage error for scintilla. Added scintilla to line 228 in CMakeLists and now links OK and even executes as well.
I can see mention of scintilla in FindwxWidgets.cmake but I am not familiar enough with CMake to know how it is working and so how it gets included automatically.
I guess it looks like I made an error building curl 7.61.1 although I thought I used a copy of the cmake command defined in BUILD.md which I assume forces a static link build via "-DCURL_STATICLIB=ON" rather than a DLL.

grumpy235:
Thanks for your efforts.
I'll download the pre-compiled wxW and see if it works for me if so then I guess this can be dropped as I have a bypass and you seem to have more than enough to do keeping the Unix/Linux community happy !
I only starting using the source for wxW when I encountered a wxDate problem some months ago which eventually went away when Microsoft got its act together and fixed the problem introduced by one of their fixes.

grumpy235:
Pre-compiled wxW works like a dream, once I had sorted out how to smash the 3 wxW folders together and got my CMAKE_INSTALL_PREFIX correct. Resulting MMEX executable is much smaller as well since now I guess it is using wxW DLLs rather than a static link.
Thanks again for your efforts.

Issue #1777, opened 2018-08-23
Filtering statuses using Transaction report problems


slodki:
Test case:
![mmex](https://user-images.githubusercontent.com/3155552/44509890-5b8b1b00-a6b3-11e8-9f31-763fa066a035.png)

Filtering for status = none: two transactions are missing (7 & 9):
![mmex2](https://user-images.githubusercontent.com/3155552/44509942-82e1e800-a6b3-11e8-9e04-ad8a20810716.png)

Filtering for status = reconciled: 6th transaction is missing:
![mmex3](https://user-images.githubusercontent.com/3155552/44510076-e0763480-a6b3-11e8-9f3e-6679185b78fa.png)

Filtering for status = void: 5th transaction is missing:
![mmex4](https://user-images.githubusercontent.com/3155552/44510322-ac4f4380-a6b4-11e8-8924-e21b461310ba.png)

Filtering for status = all except reconciled: two transactions are missing (5 & 7):
![mmex5](https://user-images.githubusercontent.com/3155552/44510472-10720780-a6b5-11e8-9e78-5d2c344a5e03.png)

Related to #1180 and #1208? All withdrawals are reported OK.

<hr><small><b>Version</b>: 1.4.0-alpha.6 • db 13 (aes128cbc) • git dd01f5ac (2018-08-20) master<br><b>Libs</b>: wxWidgets 3.1.1 (wxGTK 2.24) • wxSQLite3 4.0.4 (SQLite 3.24.0) • Mongoose 6.12 • RapidJSON 1.1.0 • Lua 5.3.3 • libcurl 7.58.0 • OpenSSL 1.1.0g • zlib 1.2.11 • libidn2 2.0.4 • libpsl 0.19.1 • ( libidn2 2.0.4) • nghttp2 1.30.0 • librtmp 2.3<br><b>Build</b>: Aug 21 2018 08:51:45 • CMake 3.10.2 • GNU Make 4.1 • gettext 0.19.8.1 • GCC 7.3.0 • Ubuntu 18.04 'bionic'<br><b>OS</b>: Ubuntu 18.04.1 LTS "bionic" • Linux 4.15.0-32-generic x86_64 • KDE Polish (UTF-8) • 1680x1050 24bit 96x96ppi</small>

slodki:
This should be rewritten?
[...]

PR #1837, opened 2018-10-05
fix(#1777): filter statuses for transfers

vomikan:
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1837)


slodki:
What do you thing about this:

```c++
if (account_selected_in_filter) { // transaction as viewed from selected account
  switch (filterStatus) {
    case "N": return itemStatus.empty() || itemStatus.StartsWith("N");
    case "U": return itemStatus.empty() || itemStatus.StartsWith("N") || itemStatus.StartsWith("F");
    case "A": return !itemStatus.StartsWith("R");
    default: return itemStatus.StartsWith(filterStatus);
  }
} else { // all accounts
  switch (filterStatus) {
    case "N": return itemStatus.empty() || itemStatus.Contains("N");
    case "U": return itemStatus.empty() || itemStatus.Contains("N") || itemStatus.Contains("F");
    case "A": return !itemStatus.Contains("R");
    default: return itemStatus.Contains(filterStatus);
  }
}
```
And description for filters U and A should be changed to something easy to understand.

vomikan:
@slodki it does not working

image

slodki:
Above example is not C++ code - this is an idea only. And still not good.
`if (account_selected_in_filter) ` is still not enough - this code must distinct between *from* account and *to* account...
I will describe the new approach here...

vomikan:
@slodki I suggest to commit this code as is. It's a bug fix.
Then you can reorginize it as you wish.

src/filtertransdialog.cpp
```diff
@@ -86,6 +86,7 @@ mmFilterTransactionsDialog::mmFilterTransactionsDialog(wxWindow* parent, int acc
     , payeeID_(-1)
     , refAccountID_(account_id)
     , refAccountStr_("")
+    , m_filterStatus("")
```
slodki: ‘mmFilterTransactionsDialog::m_filterStatus’ will be initialized after ‘double mmFilterTransactionsDialog::m_min_amount’ when initialized here [-Wreorder]

Issue #1942, opened 2019-01-20
Add button in categories dialog remains disabled when changing the selection from a subcategory to the root node


D-Barber:
**Describe the bug**
Re-selecting the root 'Categories' node in the categories dialog doesn't enable the 'Add' button but instead leaves it in whatever enabled state it was already in

**To Reproduce**
1. Open the 'Organize Categories' dialog
2. The root 'Categories' node will be selected and the 'Add' button enabled
3. Select a sub-category (e.g. Automobile->Gas in the default set), the 'Add' button will be disabled
4. Re-select the root 'Categories' node, the 'Add' button will erroneously remain disabled
5. Select a category (e.g. Automobile in the default set), the 'Add' button will be enabled
6. Re-select the root 'Categories' node, the 'Add' button will correctly remain enabled

**Expected behavior**
The 'Add' button should be enabled whenever the root 'Categories' node is selected

**Screenshots**
The incorrectly disabled button:
![image](https://user-images.githubusercontent.com/12411743/51443688-c2554480-1ce4-11e9-9062-eb051d21a712.png)

**Additional context**
N/A

**Your OS:**
- [x] Windows
- [ ] Mac OSX
- [ ] Linux 
- [ ] Android -> please report bugs to moneymanagerex/android-money-manager-ex repo!

**MMEX version:**
- [x] 1.4.x
- [x] 1.3.x
- [ ] 1.2.x
- [ ] 1.1 or older
- [x] development version (git master branch)

PR #1977, opened 2019-02-02
fix(#1942 comment): disable select button in organising categories dialog

D-Barber:
Fixes the issue mentioned in [this comment](https://github.com/moneymanagerex/moneymanagerex/issues/1942#issuecomment-459982369) on #1942. I noticed this behaviour while investigating the other issue and meant to investigate and fix it too. Seems to have broken in a refactor in 2c475bf6, I have checked all other uses of this constructor and they were all either dealt with in the refactor or added later so I believe this was the only break as a result of the change in parameters.

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1977)

### DIFF
--- a/src/mmframe.cpp
+++ b/src/mmframe.cpp
@@ -2394,7 +2394,7 @@ void mmGUIFrame::refreshPanelData()
 
 void mmGUIFrame::OnOrgCategories(wxCommandEvent& WXUNUSED(event))
 {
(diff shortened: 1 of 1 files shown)


slodki:
Please verify #1978

D-Barber:
@slodki Nice, much better

PR #1982, opened 2019-02-03
fix Polish translations

slodki:
@vomikan Crowdin updates can still downgrade existing translations (replace already translated string with an empty one).

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1982)


vomikan:
@slodki it's not a problem in case if original string is the same as translated. For example 'USD'. 
I'll prefer if all translation would be done in crowdin becouse any changes in pl_PO file will create PR.

PR #1987, opened 2019-02-03
fix(l10n): remove empty and wrong english translations

slodki:
en_US will never be loaded because it's set as sources' native language:

https://github.com/moneymanagerex/moneymanagerex/blob/0d4570c80bf21225e79ba010ce8b9319a351d93d/src/mmex.cpp#L71

en_GB is crap - only few strings are really transtated to UK English. Do we need two English MMEX versions?

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/1987)


vomikan:
It comes from forum many year's ago.
IMHO it useless.

Issue #2032, opened 2019-02-16
Import error - internal transfers recorded incorrectly (in reverse)


t4777sd:
**Describe the bug**
I am evaluating the software, so I imported from Quicken. Most of the import is correct however the import software is making a simple mistake consistently. **It is recording internal transfers in the reverse of what they actually are**.

For example, pretend there is an internal transfer from Bank#1 to Bank#2. It would be recorded as a withdrawal of money from Bank#1 and a deposit of money to Bank#2. However, the reverse of this is actually done when importing. Bank#1 records a deposit of money and Bank#2 records the withdrawal. Put simply, it records the transfer as Bank#2 -> Bank#1 instead of Bank#1 -> Bank#2

So, it seems to be getting the direction of these transfers wrong. The end result is very incorrect account balances. Unfortunately, in my case, I probably have thousands of these over the last years, so it means I cannot try the software.

**To Reproduce**
1. Export QIF file with all QIF options checked (accounts, payees, etc). Ensure there are internal transfers in the file.
2. Import QIF said QIF file and the transfers will be recorded in reverse

**Expected behavior**
It is expected that a transfer from Bank#1 to Bank#2 will result in a withdrawal in Bank#1 and a deposit in Bank#2 and not the reverse.

**Your OS:**
- [X] Windows

**MMEX version:**
- [X] 1.3.x - It says this is the latest version when I go to Help -> Check For Updates

vomikan:
If you upload QIF sample file with even one transfer transaction it would be helpful.

t4777sd:
Hello Vomikan I am working to find an example case. In an attempt to get an example case with the least possible transactions in it I came across another issue in #2033. I will be uploading an example case for that issue (which appears different than this issue).

Still working on getting a case for this isseu.

Issue #2016, opened 2019-02-10
csv import notes


vomikan:
I suggest to add  option instead "Don't care":
 "Append to Notes"
 In that case some useful info may be accepted.

![image](https://user-images.githubusercontent.com/6836805/52532394-73279000-2d35-11e9-9f12-b58cc9313d3d.png)

Issue #1963, opened 2019-01-29
Importing zero amount transactions from a CSV file


Jimmyson:
**Is your feature request related to a problem? Please describe.**
Presenting an option to import transactions with a Withdrawal or Deposit of 0 _(Zero)_ on CSV files
I am in the process of importing digitised statements, and some of these transactions have no impact on the final balance. The importer is parseing these records as "Invalid Amount", yet the application supports user-created records with a Transaction Amount of 0 _(Zero)_.

**Describe the solution you'd like**
Presenting a checkbox on the UI to allow the user to process Transactions with a Amount, Deposit or Withdrawal of 0 _(Zero)_.
This can allow user to decide to either record or skip "Nil Amount" transactions

**Describe alternatives you've considered**
An alternative could be to have the user add additional records to the Import CSV that can negate the transaction to record a zero, but this would not be consistent with the recorded transaction, in the form of "Phantom Transactions"

**Additional context**
In my case, these recorded transaction are either a combination of Individual Transactions with will be later split in the system, or a reoccuring Interest transaction that deposits no money into the account.

slodki:
1. We should allow importing transactions with 0 amount to be consistent with MMEX GUI functionality
2. There should be an option in the import dialog to *ignore 0 amount transaction records*

Issue #1700, opened 2018-06-03
QIF Import - Transfers wrong direction


ashleycheekuk:
**Describe the bug**
Bug #999  Item 3 still not resolved.
I'm experiencing the same issue which is preventing me from importing from MS Money.
Transfers into my account are showing as debits and transfers out as credits (wrong way round)
The corresponding account correctly mirrors the transaction, but again is the wrong way round as a credit/debit.

> Most (all?) transfer transaction are imported wrongly, eg. this is a MS Money transfer transaction from a bank account to a cash account, when exporting the cash account:

```
^
D04/12'2006
CX
T50.00
L[Fineco]
^
```

> so this is transfer of 50 euro exiting from bank [Fineco] and entering to cash account. MMEX imports this as an outgoing transfer from the cash account (reducing the cash account balance, instead of increasing it). WRONG. MS Money exports outgoing transfers as negative values, and incoming transfers as positive values. This is positive so INCOMING.
> 

**To Reproduce**
Import the transaction in the example from Bug #999 

**Expected behavior**
Imported transfers should be positive or negative in the account being imported and opposite in the other transfer account.

**Your OS:**
- [x] Windows
- [ ] Mac OSX
- [ ] Linux 
- [ ] Android -> please report bugs to moneymanagerex/android-money-manager-ex repo!

**MMEX version:**
- [ ] 1.4.x
- [x] 1.3.x
- [ ] 1.2.x
- [ ] 1.1 or older
- [ ] development version (git master branch)

jfrank14:
I've just run into this as well. It seems like a pretty serious bug, making MMEX unsuitable for migration from MS Money--which is what I need to do--since even one transfer ANYWHERE in the import will completely mess up the whole account. How hard would this be to fix?

Also, can it be worked around by just changing the sign on the transfer transaction? If I pre-processed the QIF file to do this, would the import be handled correctly, or cause still more errors?

vomikan:
OK
I've created two accounts
![image](https://user-images.githubusercontent.com/6836805/64097722-ab43fe00-cd6d-11e9-89e0-ef1a5c57502b.png)
Then export data from Cash account.
```
!Account
NCash
TInvoice
D[RUB]
^
!Type:Cash
D02.09.19
C
T50.00
P50.00 RUB Fineco -> 50.00 RUB Cash
N#1
L[Fineco]
^

```

vomikan:
Then the transaction deleted and imported from the QIF file
[...]

PR #2040, opened 2019-02-21
fix(#1122): bulk fixes

### LINKED ISSUE #2033: Import Error - Failed to import transactions (test file included)
**Describe the bug**
When trying to debug more #2032 I came across another issue. I exported a single account in quicken versus every account to narrow down just the bank transfer behavior.

I then attempted to import this. What I discovered is that no transactions were created. Attached is an example QIF file. 

**To Reproduce**
1. Import the attached QIF file

**Expected behavior**
The expected bahavior would be that Bank1 has $250,000 in withdrawals. And that Bank2 has $250,000 in debits. Instead, that is not what we see. No transactions are imported and errors are given. This is different than the issue #2032 where transactions are imported, but transfer direction is incorrect (withdrawal / deposit reversed).

**Your OS:**
- [X] Windows

**MMEX version:**
- [X] development version (git master branch)

vomikan:
#2033
#2032
#2016
#1963
#1700 ??
#1122

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/2040)

### DIFF
--- a/src/import_export/qif_import_gui.cpp
+++ b/src/import_export/qif_import_gui.cpp
@@ -52,8 +52,7 @@ wxBEGIN_EVENT_TABLE(mmQIFImportDialog, wxDialog)
 wxEND_EVENT_TABLE()
 
 mmQIFImportDialog::mmQIFImportDialog(wxWindow* parent, int account_id)
-    : m_init_account_id(account_id)
-    , m_userDefinedDateMask(false)
+    : m_userDefinedDateMask(false)
     , m_today(wxDate::Today())
     , m_fresh(wxDate::Today().Subtract(wxDateSpan::Months(1)))
     , dataListBox_(nullptr)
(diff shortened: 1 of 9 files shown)


src/import_export/qif_import_gui.cpp
```diff
@@ -357,8 +360,18 @@ bool mmQIFImportDialog::mmReadQIFFile()
     wxLongLong start = wxGetUTCTimeMillis();
     wxLongLong interval = wxGetUTCTimeMillis() - start;
 
-    wxString accName = ""; //TODO: check Account check box
```
slodki: :+1:

src/import_export/univcsvdialog.cpp
```diff
@@ -647,10 +668,10 @@ const wxString mmUnivCSVDialog::getCSVFieldName(int index) const
 {
     std::map<int, wxString>::const_iterator it = CSVFieldName_.find(index);
     if (it != CSVFieldName_.end()) {
-        return wxGetTranslation(it->second);
+        return (it->second);
     }
 
-    return _("Unknown");
+    return wxTRANSLATE("Unknown");
```
slodki: why?

vomikan: it's OK
https://github.com/vomikan/moneymanagerex/blob/%231122/src/import_export/univcsvdialog.cpp#L1154

src/import_export/qif_import_gui.cpp
vomikan: Bad style to leave todos in the PR but I want to commit as is.

Issue #1775, opened 2018-08-21
Update status for 2 characters transfer


gabriele-v:
We should release a patch for https://github.com/moneymanagerex/database/issues/32 and consequently drop support in model (if exists) for 1 character transfer.
I think it's the right time to do it before official 1.4.0 that will anyway beak android compatibility because of table name changes.

slodki:
Maybe old behavior (from #1773) should be default for 1.4.0 release? If old database will be auto converted to new statuses user won't be warned...

gabriele-v:
Agree that have be fixed for 1.4.0 but we should have just one DB model with 2 characters, we cannot support both 1 and 2 characters in DB for transfers, because it will mean different SQL reports etc..

If global option "reconcile transfers separately" is disabled (default) then
- Single character shown in the GUI (but still 2 characters in DB that will always be equal)
- Reconcile as previous mmex versions
If it's enabled:
- Two characters shown in the GUI (and obviously in DB)
- Reconcile as actually in master

Issue #1773, opened 2018-08-21
1.4.0-alpha.6 "transfer" transactions states


48mWsPK:
**Describe the bug**
Marking the state of a "transfer" transaction in one account does not update it automatically in the other, for example, reconciling a transfer from "VISA CARD" checking account to "WALLET" cash account from withing "WALLET" still leaves the transaction as "unconfirmed" in "VISA CARD" account, i.e. it still necessary to go and manually chage the transfer transaction state to "reconciled" or whatever...
MMEX < 1.x.x.x did this better!

**To Reproduce**

**Expected behavior**
"Transfer" transactions should automatically chage state in both affected accounts.

**Your OS:**
- [X] Windows
- [ ] Mac OSX
- [ ] Linux 
- [ ] Android -> please report bugs to moneymanagerex/android-money-manager-ex repo!

**MMEX version:**
- [X] 1.4.x - Version: 1.4.0-alpha.6, Git commit: dd01f5ac (2018-08-20)
- [ ] 1.3.x
- [ ] 1.2.x
- [ ] 1.1 or older
- [ ] development version (git master branch)

gabriele-v:
See https://github.com/moneymanagerex/moneymanagerex/issues/1180

Generally "reconciled" status on a transfer it's not done at the same time on both ends, in your example when you withdraw money from your credit card, it's reconciled immediately on wallet/cash account but not on credit card account (it could take few days).
Also, I (and most of users) will reconcile a transaction when they compare MMEX with bank statement, consider for example the case when today I reconcile an account with his statement, but I haven't reconciled the other account (because statement it's not ready, or simply I don't have time to do it now for the other), this is correctly reflected in MMEX with one account reconciled and not the other.

We could anyway think about adding a global option to restore the old behavior, or reconcile both automatically just in case the destination is a cash account type (if you deposit cash, it could take few days to see it on bank account).
@stef145g What do you think about?

stef145g:
For many years users have complained that in a transfer transaction, the status was automatically changed in the other account which had not been reconciled yet. A solution was finally found and implemented.

As @gabriele-v has suggested that an option could be provided to restore the old behavior to suite some users. Not sure about having account types automatically setting both ends of the transaction.

cliveju:
stef145g please can you explain how to prevent transfers being reconciled in both accounts at the same time? I have searched the forum and haven't found it.

gabriele-v:
[...]

PR #2221, opened 2019-06-20
Fix #1782 - update db to v14

### LINKED ISSUE #1621: DATAVERSION<>user_version
We use hardcoded DATAVERSION value 3:

https://github.com/moneymanagerex/database/blob/c7250e9f14bd7d18c897d809d47cfb42742f5c6b/tables_v1.sql#L454

https://github.com/moneymanagerex/database/blob/c7250e9f14bd7d18c897d809d47cfb42742f5c6b/database_version_3.sql#L3

https://github.com/moneymanagerex/moneymanagerex/blob/master/src/constants.cpp#L72

But it's ignored later:
https://github.com/moneymanagerex/moneymanagerex/blob/master/src/dbupgrade.cpp#L51

And not updated:
https://github.com/moneymanagerex/moneymanagerex/blob/master/src/dbupgrade.cpp#L93

We should:
a) remove DATAVERSION from INFOTABLE_V1
b) store in DATAVERSION the same value as user_version pragma

What will happen when old (<1.3.0?) MMEX version try to open db with missing DATAVERSION record in INFOTABLE_V1?

### LINKED ISSUE #1782: DB v.14 required before 1.4.0-beta
@slodki could you implement https://github.com/moneymanagerex/database/issues/33 ?
Unfortunately I won't have development machine for a while so it will be difficoult to do it.

Thanks

slodki:
This should fix #1621 #1782

New MMEX option for  #1773 and #1775 should be developed before 1.4.0

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/2221)

### DIFF
--- a/src/constants.cpp
+++ b/src/constants.cpp
@@ -46,8 +46,6 @@ const wxString g_CloseLabel =
     wxTRANSLATE("&Close ");
 #endif
 //---------------------------------------------------------------------------
-const int mmex::MIN_DATAVERSION = 2;
-const wxString mmex::DATAVERSION = "3";
 const wxString mmex::DEFDATEFORMAT = "%Y-%m-%d"; //ISO 8601
 const wxString mmex::DEFDELIMTER = ",";
 
--- a/src/constants.h
+++ b/src/constants.h
@@ -36,9 +36,6 @@ extern const wxString g_CloseLabel;
 
 namespace mmex
 {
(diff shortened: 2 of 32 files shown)


slodki:
There are still many places in MMEX where two-letter statuses for transfers are not correctly used!
This must be fixed before merging this PR.

Issue #2443, opened 2020-04-30
Web app lost category and sub-category


sharkyiwc:
Hi,

After update from 1.3.3 to 1.3.4, when we use webapp, dropdown menu for category and sub-category is empty.

Thanks for your work

- Check OS
  - [X] Windows
  - [X] Linux 
-  Check MMEX version
  - [X] 1.3.4

sharkyiwc:
Re,

I have try mmex-1.4.0-alpha.7-win64, it's same.
after Sync WebApp, in web page dropdown menu for category and sub-category is empty.
I restore 1.33 and it's back.

Thanks

vomikan:
Screenshot please

sharkyiwc:
Yes,
If you want to highlight that this is not a drop-down menu, OK
On the other hand in version 1.33, this completes automatically in category and sub- is filtre from first category selected.
And this function this loses as much as I do an upgrade.
Thanks

sharkyiwc:
And is not browser history
Just change version, make a import webapp (=sync)

V1.33
![2020-04-30 19_51_36-Window133](https://user-images.githubusercontent.com/251279/80743483-4f5c8f80-8b1d-11ea-93d1-c0acd1cf14c4.png)

V1.34 and +
![2020-04-30 19_50_48-Window144](https://user-images.githubusercontent.com/251279/80743502-584d6100-8b1d-11ea-82d3-97afef232421.png)

vomikan:
I am warking on this case right now.
Can't fix this case https://github.com/moneymanagerex/moneymanagerex/issues/2413

vomikan:
Is't OK?
![image](https://user-images.githubusercontent.com/6836805/80757761-5c897680-8b3d-11ea-899a-5f80a22f4d32.png)

vomikan:
This build should work fine
https://ci.appveyor.com/project/moneymanagerex/moneymanagerex/builds/32566162/job/l3v93ake7tb0evew/artifacts

sharkyiwc:
Great!
And fast .... Thanks you

sharkyiwc:
Hi,
I'm back after real test and with the version 1.3.5 B2 i have same problem.
I'm sorry for bad news

vomikan:
Because last changes are not included there.

vomikan:
https://github.com/moneymanagerex/moneymanagerex/releases/tag/v1.3.5-beta.3

[...]

PR #2452, opened 2020-05-04
fix(#2443): Convert to UTF8

gabriele-v:
@vomikan could you check if it solves #2443 for you as well?
Now it's working for me with russian characters

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/2452)


vomikan:
@gabriele-v  Export would be fine. But WebApp does not accept JSON string with nan Latin charecters.

gabriele-v:
After these changes mine it's working fine for the entire process:
image
image
image
image
image
image


At which step is your broken?

PR #2464, opened 2020-05-06
fix(#1585): Improve CustomFields for 1.3.5

gabriele-v:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/2464)


gabriele-v:
@moneymanagerex/desktop-mmex-team I've removed a .cpp and a .h file not needed anymore, should I change something in make files? Because looks like AppVeyor and Travis are not building..

Issue #2512, opened 2020-05-24
lost all currency exchange rates


vomikan:
https://forum.moneymanagerex.org/viewtopic.php?f=7&t=9338

have installed the version 1.3.5-RC1
I noticed that I lost all currency exchange rates.

- Check OS
  - [ ] Windows
  - [ ] Mac OSX
  - [ ] Linux 
-  Check MMEX version
  - [x] 1.3.x

vomikan:
![ch](https://user-images.githubusercontent.com/6836805/82761352-9d09a680-9e02-11ea-861f-837d6bff2fa5.PNG)

gabriele-v:
Don't know how to fix/manage it without issue another DB version..

vomikan:
may be make it true by default?

gabriele-v:
You were thinking to completely remove it :)

I think MMEX should be easier by default, then add complexity only if someone wants it..

vomikan:
I mean switch the option "Use currency history" on by default.
        m_currencyHistoryEnabled = Model_Infotable::instance().GetBoolInfo(INIDB_USE_CURRENCY_HISTORY, *false*);

gabriele-v:
I may have another idea, I'll send a PR tomorrow

vomikan:
![image](https://user-images.githubusercontent.com/6836805/82764261-2aa2c180-9e16-11ea-8ec4-9efce1b520bc.png)

gabriele-v:
Probably the DB has skipped some patches in the past, at the beginning of DB patch procedure there was a bit of mess, that's why we have introduced version 10 (I think) that normalize all tables..

gabriele-v:
Or.. CURRENCYHISTORY it's not initialized during DB procedure, I'll have a look at tu is tomorrow too

vomikan:
![image](https://user-images.githubusercontent.com/6836805/82764446-6db16480-9e17-11ea-80a2-f43e1e860d9b.png)

vomikan:
https://github.com/moneymanagerex/moneymanagerex/issues/2350

gabriele-v:
> ![image](https://user-images.githubusercontent.com/6836805/82764446-6db16480-9e17-11ea-80a2-f43e1e860d9b.png)
> 

Yep, this should be CURRENCYHISTORY_V1 wrong copy/paste

vomikan:
I remember that iI've fixed that
[...]

PR #2598, opened 2020-06-05
fix(#2512): lost all currency exchange rates

vomikan:
true by default

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/2598)


gabriele-v:
@vomikan if you want to use currency history as default, also this should be updated https://github.com/moneymanagerex/moneymanagerex/blob/17b0a81e2289f5335db785f225970c082b06fdaf/src/option.cpp#L69

vomikan:
@gabriele-v  are you sure? that is exactly what I've updated

image

gabriele-v:
> @gabriele-v are you sure? that is exactly what I've updated
> 
> image
Sorry, you are right, I've messed it up.. That's fine

Issue #2740, opened 2020-12-14
The application could not be started correctly (0x000007b)


vomikan:
MMEX version:
 - [x] 1.3.6-RC.1

Operating System:
 - [x] Windows

# Description of the bug

https://ci.appveyor.com/project/moneymanagerex/moneymanagerex/builds/36808941/job/ute6l07dbn2isrrx/artifacts
32-x bit application can't start

![image](https://user-images.githubusercontent.com/6836805/102077497-7355b280-3e1a-11eb-821f-2a9410cfaac5.png)

vomikan:
![image](https://user-images.githubusercontent.com/6836805/102084904-e9601680-3e26-11eb-98da-f2bcdecc8f86.png)

vomikan:
https://curl.se/changes.html#7_64_0

PR #2741, opened 2020-12-14
fix(#2740): The application could not be started correctly (0x000007b)

vomikan:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/2741)


vomikan:
image

PR #2790, opened 2021-01-07
fix(home): stock widget fix for shadow container

whalley:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/2790)


whalley:
For consistency, also added the same style to the reports.

Issue #2840, opened 2021-01-30
multi transactions update


zer0feel:
I have tried the version 1.4.0 Alpha 7, in which the multi transactions update function is very useful to edit multiple transactions at a time, especially when I want to edit the payee and the categories of multiple transactions. 

**Version 1.4 Alpha7**
![1 4 0 Alpha7_resize](https://user-images.githubusercontent.com/78265137/106358965-4a499c80-634a-11eb-92fd-ae2d5c15612c.jpg)

![image](https://user-images.githubusercontent.com/78265137/106358509-a4952e00-6347-11eb-9222-e10df796c789.png)

However, this function is not available in the stable version 1.3.6. When multiple transactions are selected and right-clicked, the edit transaction function is unselectable. I really hope this function to be available in the stable version too.

**Version 1.3.6**
![1 3 6 Stable_resize](https://user-images.githubusercontent.com/78265137/106358947-356d0900-634a-11eb-8d81-011610824f66.jpg)

vomikan:
https://github.com/moneymanagerex/moneymanagerex/issues/2659

vomikan:
https://github.com/vomikan/moneymanagerex/commit/d0f6db24bc698a228cd080e5aba68b085337cb71

PR #2842, opened 2021-01-30
fix(#2840): multi transactions update

vomikan:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/2842)


whalley:
Should be a welcome addition, go for it!.  I'll do some extensive testing on Mac once merged.

Issue #2881, opened 2021-02-04
Build date/time in about.cpp does not get updated on every build


whalley:
MMEX version:
 - [X] 1.3.7
 - [ ] 1.3.6

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [ ] Windows
 - [X] Mac OSX
 - [ ] Linux 

# Description of the bug

As the build date/time comes from __DATE__ and __TIME__ in the file it only gets updated when this file is compiled. Obviously with a full build it is correct...

Just a minor annoyance. Possibly worth looking at changing makefile to force this file to be built....

vomikan:
https://stackoverflow.com/questions/24292898/compile-date-and-time-using-cmake

vomikan:
Testing
1) https://ci.appveyor.com/project/moneymanagerex/moneymanagerex/builds/37647356
![image](https://user-images.githubusercontent.com/6836805/107140447-0995e880-6933-11eb-99e5-332b56478851.png)

vomikan:
2)
![image](https://user-images.githubusercontent.com/6836805/107140991-fb49cb80-6936-11eb-933b-decdf3208d19.png)

vomikan:
Seems it's working fine as is.
But on my local PC the build.h file does not rewrite on every building.

vomikan:
# By setting the application version here, CMake will transfer the variable
# values to the corresponding variable names in the following files:
# ./src/build.h
# ./resources/mmex.rc

vomikan:
Overall, I've made improvements. This works quite well. If you want to make it better, you are welcome.

PR #2885, opened 2021-02-07
fix(#2881): build date/time in about.cpp

vomikan:
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/2885)


whalley (commented):
Think we would would want to exclude build.h from the src tree and/or include in the .gitignore file.

PR #3023, opened 2021-03-08
fix(#3109): settings/options fix

whalley:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/3023)


whalley:
This should of course be #3019!

Issue #3209, opened 2021-04-05
All Transactions panel currencies


vomikan:
All values in the panel are presented in the base currency format.

I suggest redoing this to the following view:
![image](https://user-images.githubusercontent.com/6836805/113613508-03ce3200-965a-11eb-8993-d8446055f937.png)

vomikan:
Hi Vomikan,

I am on 1.5.1 RC2 and can see this implementation. One issue I see is if there was a cross currency transfer, then the deposited value is showing the 'from' currency rather than the 'to' currency. Hope I made myself clear.

reddyshyam:
Hi Vomikan,

I am on 1.5.1 RC2 and can see this implementation. One issue I see is if there was a cross currency transfer, then the deposited value is showing the 'from' currency rather than the 'to' currency. Hope I made myself clear.

Adding this again on the main feature ticket so that the issue is known.

vomikan:
It is possible to make the values appear in two cells. But I don't want to load the UI and do unnecessary calculations in this screen.

reddyshyam:
> It is possible to make the values appear in two cells. But I don't want to load the UI and do unnecessary calculations in this screen.

I understand, atleast those that are manually filled (both withdrawal and deposit) as in my case?

whalley:
This can be closed it has been implemented.

PR #3297, opened 2021-04-19
fix(#3209): All Transactions panel currencies

vomikan:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/3297)


reddyshyam:
Hi Vomikan,

I am on 1.5.1 RC2 and can see this implementation. One issue I see is if there was a cross currency transfer, then the deposited value is showing the 'from' currency rather than the 'to' currency. Hope I made myself clear.

Issue #3286, opened 2021-04-17
Hebrew help pages


vomikan:
https://crowdin.com/project/moneymanagerex/he#

vomikan:
![image](https://user-images.githubusercontent.com/6836805/115118941-8b973300-9fae-11eb-87c2-50aa234276a8.png)

vomikan:
https://flaglane.com/download/israeli-flag/israeli-flag.svg

PR #3313, opened 2021-04-22
(fix)(#3286):Add Right to Left orientation to help pages. <div dir="rtl" lang="he">

avma:

This fixes Help page reading direction to the right

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/3313)


vomikan:
fixed

PR #3354, opened 2021-04-30
fix(mmex.cpp:362): useless cast to type 'double'

vomikan:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/3354)

### DIFF
--- a/src/mmex.cpp
+++ b/src/mmex.cpp
@@ -359,9 +359,9 @@ bool OnInitImpl(mmGUIApp* app)
         if (inter.IsEmpty())
             percent = 0;
         else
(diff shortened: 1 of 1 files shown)


whalley (commented):
Don't think this change is right as it will do the division using integers and return ZERO not the fractional amount required.

whalley:
Actually I think there is another bug there in my original code as the wxRect:Intersect will change rect1, so need to extract the height and width before the percentage calculation.

vomikan:
My fixes comes from gcc (linux) building. 
I'm not sure I didn't make a mistake. If the fix is obvious, just fix it.

Issue #3101, opened 2021-03-17
Transaction goes mad


Tulevik:
MMEX version:
 - [ X ] 1.3.7
 - [ X ] 1.3.6

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [ X ] Windows
 - [ ] Mac OSX
 - [ ] Linux 

# Description of the bug

After clicking with mouse in Edit Transaction, the new value of money you correct ends up multiplied by an order of 100x or more, whenever you change the field or press Accept button.

# Reproduction

Is the bug reproducible? 
 - [ X ] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:

1. Click on any transaction with right mouse button, and then click Edit Transaction
2. Modify the value in money
3. Press TAB or click on Accept. The value will be multiplied by an order of 100x or more, without a reason.

Expected result: That it don't multiply itself

Actual result: it multiplies itself

# Additional information

If you have a backtrace for a crash or a warning, paste it here.

whalley:
What currency are you using? Could you attach a screen shot of the transaction before and after you TAB to field and it changes.

Tulevik:
I am using Euros. The following video speaks for itself. It even multiplies the initial budget:

https://user-images.githubusercontent.com/12910599/111639132-3f17d680-87fb-11eb-8873-c2ad6a26e7d2.mp4

whalley:
I'm afraid I can't reproduce this. Can anybody else?

Tulevik:
Is there any chance to bypass this problem? Can I edit the wrong values on the .mmb file with another program?

vomikan:
Could you check the EUR currency settings

![image](https://user-images.githubusercontent.com/6836805/111915071-9f21ad80-8a85-11eb-8f1b-9e1b20c4a56e.png)

Change them regarding regional settings.

Tulevik:
This is what I have (I can't edit the conversion to Base Rate, and there's also a 1e+18 in the last type. It also happens when I change the decimal char to , and the grouping to .
![screenshot 1](https://user-images.githubusercontent.com/12910599/111917391-02104600-8a80-11eb-9a4a-ad782b13e38c.png)

Tulevik:
Ok, I've found the solution:

[...]

PR #3366, opened 2021-05-03
fix(#3101): removed reliance on wxNumberFormatter

whalley:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/3366)

### DIFF
--- a/src/mmTextCtrl.cpp
+++ b/src/mmTextCtrl.cpp
@@ -51,7 +51,7 @@ void mmTextCtrl::SetValue(double value, const Model_Currency::Data* currency, in
 
 bool mmTextCtrl::Calculate(int alt_precision)
 {
-    const wxString str = Model_Currency::fromString2Lua(this->GetValue(), m_currency);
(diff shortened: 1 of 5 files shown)


reddyshyam:
What does this mean?

whalley:
We have been having some issues with 'locale' and the relationship to the mmex currency settings. This change was an attempt to always use the mmex currency settings for number presentation ignoring any system settings.

reddyshyam:
Glad that this is looked into again but the issue with few countries like India, our number system is different.

We represent currency/numbers are 91,23,45,678.90 as against 912,345,678.90 which would have worked out well if locale was picked but sadly that didn't work. I am happy dependency is removed but is there a way we can allow user defined formats, like (2,3) Indian or (3) US?

@vomikan is aware of this but we are waiting for a properly solution available by some library.

whalley:
@reddyshyam Understood. Once we get this bit fixed we can look at making the group sequence more flexible.

Issue #3517, opened 2021-05-27
[Att.] is added in notes


TomAff:
MMEX version: 1.5.1
Operating System: Windows

If you change the status of a transaction with an attachment by using the keys R F V... 
[Att.] is added to the notes-field everytime.
![2021-05-27 07_28_04-Window](https://user-images.githubusercontent.com/46572503/119771356-dc5b4f00-bebd-11eb-8999-061e9182cdf9.jpg)

tactilis:
Same incorrect behaviour in latest 1.6.0 Alpha.1

TomAff:
Where can I get 1.6.0 Alpha.1?

vomikan:
![image](https://user-images.githubusercontent.com/6836805/119805790-3d991780-beea-11eb-8315-af86f6cd2821.png)

![image](https://user-images.githubusercontent.com/6836805/119805880-5acde600-beea-11eb-9b46-d1bf6a52b9f5.png)

tactilis:
@TomAff

As @vomikan said, or go directly here [https://ci.appveyor.com/project/moneymanagerex/moneymanagerex](https://ci.appveyor.com/project/moneymanagerex/moneymanagerex) and look at 'History' for latest merge into master.

Select Platform: x64, then Artifacts.

I'd suggest that you download the portable build because then you can run that as well as keeping your 1.5.1 installed.

TomAff:
Thanks a lot. I use the portable Version since years. It is easy to Update and easy to backup.

The last months I use MMEX more intensiv, creating reports and try to use a lot of feature.

I'm looking forward to every new Version

vomikan:
Possible performance degradation
![image](https://user-images.githubusercontent.com/6836805/119870383-fd598980-bf29-11eb-9892-7821bda1b44e.png)

TomAff:
I'll try to reproduce it as I understand it.
The info, that there is an attachment is loaded into the cache and only output as info in the notes column? It is not saved in the notes field normaly?

When changing the status by pressing key (UFVRD) in the overview, only the "Status" field should be updated. The other fields do not. That should solve the problem.
Or maybe just change the output.
If attachment then notes = [Att.] & Data of the Notesfield

vomikan:
Everything should be OK now

dbolton:
FYI, The download page of the main website has links and instructions for downloading the latest stable build: https://www.moneymanagerex.org/download#Unstable

PR #3520, opened 2021-05-27
fix(#3517): att is added in notes

vomikan:


gabriele-v:
@vomikan I've seen this issue as well in the past, but I was not able to found a solution other than your one but that's wrong so I left as it was. 

Because in this way with your fix the DB will be queried for each transaction looking for attachments and that's pretty slow, the same issue was happening for the splits so we had to cache them as well as you can see in the code.
That's the reason why the previous code was caching all the attachments and then look in the cache if the transaction has an attachment.

PR #3522, opened 2021-05-27
fix(#3517): speedup

vomikan:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/3522)


gabriele-v:
Why HAS SPLIT and attachments are mixing up? I can test the artifacts but I think something could be wrong, or it's just a wrong variable names?

tactilis:
Also, a typo here:

image

vomikan:
It's time for me to go to sleep ))

Issue #3527, opened 2021-05-28
Loss of colour in Graphs 


grumpy235:
MMEX version:
 - [ ] 1.5.x
 - [x] 1.6.x

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [x] Windows
 - [ ] Mac OSX
 - [ ] Linux 

Minor issue which you may well have already found with Master at ce612a4.
Income expenses applet on home page has lost its colour
![Capture](https://user-images.githubusercontent.com/24897056/120014727-5123ab80-bfda-11eb-905d-095f1333f6a0.PNG)

Please describe your issue with details.
Add screenshot or other files if needed.

# Reproduction

Is the bug reproducible? 
 - [ ] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:

1. 
2. 
3. 

…

Expected result:

Actual result:

# Additional information

If you have a backtrace for a crash or a warning, paste it here.

grumpy235:
@vomikan - just checked and all graphs are now grey

vomikan:
It's seems OK now.
Working with the following JSON:
```
{
  "theme": {
    "name": "MMEX Default Theme",
    "author": "MMEX Team",
    "description": "This is the default MMEX theme and uses the MMEX colour scheme. It is a simple 'duo color' theme",
    "url": "https://github.com/moneymanagerex/themes/blob/master/system-themes/default/README.md"
  },
    "colors": {
        "navigationPanel": ""
    } 
}
```

Issue #3510, opened 2021-05-25
Fonts in Navigation not visible


rony925:
MMEX version:
 - [ ] 1.5.1

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [ ] Mac OSX Big Sur
 

# Description of the bug

The fonts in the navigation panel are all white same as the background.
I can see only icons no text.
[
<img width="1680" alt="Screen Shot 2021-05-26 at 3 48 01 AM" src="https://user-images.githubusercontent.com/15120867/119566398-4c41da80-bdd5-11eb-8964-8b99fb564e45.png">
]

# Reproduction

Is the bug reproducible? 
 - [ ] Always 

Reproduction steps:

1. Just start the app and it shows as the screenshot
2. 
3. 

…

Expected result:

Actual result:

# Additional information

If you have a backtrace for a crash or a warning, paste it here.

vomikan:
#3364

vomikan:
Someone should create black theme
https://github.com/moneymanagerex/themes

whalley:
Agreed. There is a way, certainly on MacOS to determine via the APIs, if Dark Mode is being used. Once somebody creates a default Dark Mode theme we could perhaps use that as the default theme if no other theme selected.

whalley:
https://github.com/moneymanagerex/moneymanagerex/issues/1779

PR #3530, opened 2021-05-28
fix(#3527): #3510

vomikan:

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -26,7 +26,7 @@ environment:
   # clcache config
   clcVer: 4.1.0
   # libcurl config
-  curlVer: 7_61_1
+  curlVer: 7_61_0
   curldir: C:\curl-$(curlVer)
   curlStatic: ON
   # wxWidgets config
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -26,7 +26,7 @@ environment:
   # clcache config
   clcVer: 4.1.0
   # libcurl config
-  curlVer: 7_61_0
+  curlVer: 7_61_1
   curldir: C:\curl-$(curlVer)
   curlStatic: ON
   # wxWidgets config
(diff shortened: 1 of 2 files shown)


GitCop:
Thanks for your pull request while there were the following issues

 * Commit: 7c84ad2bd57bdee07ca0cb30e84037fc54f5199d
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

 * Commit: db156664b84a6ee9cacc12d121c15c0a61abc4bb
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

---

whalley:
@vomikan - Not sure why the colors have been changed to being mandatory in the theme. If they are mandatory then we should also set the color to "" as a default means nothing if it is mandatory. The fact they are optional means the application default will be used if not defined in the theme.  

Am I missing something here?

vomikan:
The solution is this. If an empty value is set for the panel color, then do not apply the color. Since the colors are not set in the JSON file, they must be taken from the list of default values.

whalley:
So, original intended logic was...

If the metadata item was marked as "MANDATORY" then it **must** be defined in the theme JSON, so the default value was meaningless as it would always be taken from the theme JSON.

If the metadata item was **not** marked as "MANDATORY" then if it is defined in the theme JSON this value will be used. If it was not defined in the theme JSON then the application default will be used.

So code such as:
`md[COLOR_LISTALT0]         = std::make_tuple("/colors/listAlternative1", "#F0F5EB", true);`

should be:
`md[COLOR_LISTALT0]         = std::make_tuple("/colors/listAlternative1", "", true);`

As the data should be mandatory in the theme JSON.

vomikan:
Thank you for the clarification.
I need to think about whether there was a logical error after my changes.

My idea is as follows. The value can be set in a JSON file, even empty. If the value is empty, then it should be ignored (not applied to the UI object). But if it is REQUIRED, then find the corresponding value from the default list.
If the value is not set in the JSON file, we take the default value.

vomikan:
If you clearly see how to fix the code, please do so.

whalley:
I was not aware it was broken :-) Will have look later.

whalley:
Reverting the changes you made makes no difference. It worked fine without the changes you made.

whalley:
Actually ignore above... There does seem to be an issue.

Issue #2066, opened 2019-02-26
    wxSplitterWindow for GRM


vomikan:
It's time to add  wxSplitterWindow for GRM UI 
Sample:
```
    wxSplitterWindow *splitter = new wxSplitterWindow(this, wxID_ANY,
                                            wxDefaultPosition, wxDefaultSize,
                                            wxSP_NOBORDER);
```

vomikan:
I've added splitter between SQL text and DB structure
But can't add splitter where is a yellow line
![image](https://user-images.githubusercontent.com/6836805/53694904-eae15b80-3dc5-11e9-868e-dcdb6213aec5.png)

vomikan:
The same for 
![image](https://user-images.githubusercontent.com/6836805/117480305-1b0a8300-af6a-11eb-93d3-fe30bc2cd80b.png)

whalley:
Splitter for Themes added, that one for GRM looks trickier....

whalley:
> Splitter for Themes added, that one for GRM looks trickier....

My head is spinning trying to sort out the splitter for GRM..... I'll leave it for now.

whalley:
Starting with old school pen and paper helped.... :-) Splitter now added.

PR #3582, opened 2021-06-07
fix(#2066): theme splitter

whalley:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/3582)

### DIFF
--- a/src/themes.cpp
+++ b/src/themes.cpp
@@ -163,21 +163,28 @@ void mmThemesDialog::CreateControls()
 {
     wxBoxSizer* bSizer0 = new wxBoxSizer(wxVERTICAL);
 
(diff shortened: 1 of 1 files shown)


reddyshyam:
Is 1.6.0 not updated anymore? I am always on that stream

whalley:
@reddyshyam At present the 1.5.3 version is tracking the master branch. So please use this as the latest version.

reddyshyam:
Hi @whalley, thanks for the update but I am already on 1.6.0 stream, can I overwrite with 1.5.3 without any damage?

vomikan:
@reddyshyam yes you can. The code is the same.

reddyshyam:
> @reddyshyam yes you can. The code is the same.

Ah! Thanks for the update, @vomikan. Let me right away as I always stay on the snapshot stream. 👌🤝

Issue #3644, opened 2021-06-22
Value in the sale of stocks by adding the rates


mouraee:
MMEX version:
 - [ ] 1.5.x
 - [X ] 1.6.x

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [X ] Windows
 - [ ] Mac OSX
 - [ ] Linux 

# Description of the bug
By including a sale of a stock, the fee is being added to the amount received when the right is to reduce the rate

![mmex_22062021](https://user-images.githubusercontent.com/7883494/122925162-f2cdbc80-d33c-11eb-85cf-e0b1f5ed75e4.JPG)
![mmex_22062021b](https://user-images.githubusercontent.com/7883494/122925167-f3665300-d33c-11eb-953f-763ce11196f4.JPG)

# Reproduction

Is the bug reproducible? 
 - [ X] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:

1. Add Stock Transaction
2. 
3. 

…

Expected result:

Actual result:

# Additional information

If you have a backtrace for a crash or a warning, paste it here.

vomikan:
It seems 
8 * 91,65 + 0,20 = 733,40

mouraee:
When you sell an asset, you must subtract the rate to know the actual result of the sale, as this will be your actual profit or loss.

mouraee:
When you buy an asset, you add the fees (because you increased its cost), when you sell an asset, you subtract the fees (because you received a lower amount). The difference between the amount you sold and the amount you bought is the profit or loss

vomikan:
I may be wrong, but the sale of assets has not been implemented.

mouraee:
Yes, it has already been implemented. The print screeen above is the sale function (For each stock an auxiliary account is created, when you buy a stock a debit is generated on this account and when you sell a stock a deposit is generated).

The only problem is that the deposit generated on the sale of the stock is adding up the fee, when the right thing is to subtract the fee (if you paid what you will receive is the amount minus the fee)

whalley:
Also spotted that the amount is not updating when the no of shares, price, or commission is changed. Should be fixed now.

vomikan:
There are doubts about the correctness of such a decision. There must be a condition. In one case, we add the commission, in the other we subtract it.

mouraee:
In the buy add the fee (deposit)
[...]

PR #3675, opened 2021-06-30
fix(#3644): commision fix

whalley:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/3675)


vomikan:
There are doubts about the correctness of such a decision. There must be a condition. In one case, we add the commission, in the other we subtract it.

PR #4178, opened 2022-02-05
fix(#4031): call payee dialog by enter press

vomikan:
@whalley could you, please, investigate this changes regarding how it would works in MAC?

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/4178)


whalley (commented):
On MacOS pressing return of the empy payee field works fine and brings up the payee dialog fine.

Issue #4269, opened 2022-03-29
[Linux] Graphics like charts not working


MartinX3:
MMEX version:
 - [x] 1.5.13
 - [ ] 1.6.x

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [ ] Windows
 - [ ] Mac OSX
 - [x] Linux 

# Description of the bug

I only the an empty, white space.

# Reproduction

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:

1. Open the reports

Expected result:
My charts

Actual result:
White area

# Additional information

Everytime I open a report with charts
`** (mmex:21069): CRITICAL **: 19:17:11.814: Cannot register URI scheme memory more than once`

Specs:
```
Version: 1.5.13 64-bit
• Aufbauend auf Mär 9 2022 08:52:13
• Datenbankversion: 7 • (aes128cbc)
• Git commit: d019150be (2022-02-06)

MMEX verwendet folgene Unterstützungsprodukte:
• wxWidgets 3.1.5 (wxGTK 3.24)
• wxSQLite3 4.6.0 (SQLite 3.31.1)
• RapidJSON 1.1.0
• Lua 5.4.4
• lunasvg v2.0.1
• libcurl/7.82.0 OpenSSL/1.1.1n zlib/1.2.12 brotli/1.0.9 zstd/1.5.2 libidn2/2.3.2 libpsl/0.21.1 (+libidn2/2.3.0) libssh2/1.10.0 nghttp2/1.47.0
• gettext 0.21
• apexcharts.js

Genutztes Build:
• CMake 3.22.3
• GNU Make 4.3
• GCC 11.2.0 • EndeavourOS rolling 'rolling'

Läuft auf:
• EndeavourOS Linux "rolling"
• Linux 5.17.1-zen1-1-zen x86_64
• German (UTF-8)
• 1920x1080 24bit 96x96ppi
```

PMaff:
Reports -> Cash flow -> Daily 
shows this already when starting version 1.5.13 on commandline with  /usr/bin/mmex on Ubuntu:
[...]
** (mmex:12723): CRITICAL **: 20:44:24.501: Cannot register URI scheme memory more than once
20:44:24: Debug: futureIgnored: 0
20:44:24: Debug: ======= mmReportCashFlow:getHTMLText =======
[...]

PR #4271, opened 2022-03-31
fix(#4269): wxgtk

vomikan:
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/4271)


whalley:
As noted [here](https://github.com/moneymanagerex/moneymanagerex/issues/4269#issuecomment-1085194822) I don't think the `#if` change is needed or will fix the issue.

The refactor looks OK.

vomikan:
mmex supports only 3 systems. The virtual file system only works for MAC and Windows. I consider this #if acceptable. I suggest to commit. A miracle may happen and it will help the initiator of this issue.

whalley (approved):

Issue #4250, opened 2022-03-09
Allow more than 10 saved transaction filters


cmbarbu:
# Description of the feature

Currently, up to 10 transaction filter settings can be saved. This limit could be increased (or even removed altogether).

I sometimes discover new filter settings that I would like to save for future reference, but I am unable to save them since all available "slots" are already taken.

![mmex-feature-trans-filter-settings](https://user-images.githubusercontent.com/1184162/157521762-98de65b5-f9f2-48c1-86d9-59ce21706781.PNG)

To prevent the drop-down menu from becoming too large, a scrollbar could be added (however, I am not sure whether this is supported by wxWidgets).

vomikan:
@cmbarbu Could you please check out the new functionality?

https://ci.appveyor.com/project/moneymanagerex/moneymanagerex/builds/43545863/job/28ohv9xwc60ymagc/artifacts

whalley:
@vomikan 

Initially I can no longer select accounts:
![CleanShot 2022-05-16 at 00 43 29](https://user-images.githubusercontent.com/487345/168499395-9371d24d-37f5-4153-9802-97be5af81720.png)

I can no longer select Payees:
![CleanShot 2022-05-16 at 00 44 06](https://user-images.githubusercontent.com/487345/168499426-7e83f824-b796-4f7e-95ff-2544b53c4763.png)

**However, after saving my first Filter I can then select Accounts and Payees.**
![CleanShot 2022-05-16 at 00 46 58](https://user-images.githubusercontent.com/487345/168499569-7b08bd7d-7a5d-4f64-8d01-8f417a368d8b.png)

If I delete my filters, exit the dialog and go back in I am again unable to select accounts or payees.

One further thing. I cannot update an existing filter without removing and re-creating it.

vomikan:
I confirm this bug. Suddenly. I will fix it right after refactoring the Navigation Panel.

whalley:
> One further thing. I cannot update an existing filter without removing and re-creating it.

I can't see a fix for this, is this intentional? Perhaps a warning if an existing filter name is entered before overwriting? 

Also I think we should pop a warning when deleting a filter, asking for user confirmation.

What do you think?

whalley:
Also noted that new filters are not displayed or removed from the navigation menu (i.e. it is not refreshed) until the transaction report is run. 

[...]

PR #4369, opened 2022-05-14
fix(#4250): allow more than 10 saved transaction filters

vomikan:
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/4369)


whalley:
@vomikan Does not seem possible to re-save using the same name. E.g. If I save with Type = Deposit with Name = "My Filter". I am then unable to add a period range and re-save under "My Filter"

vomikan:
Yes, I specifically made a protection against using the same names to avoid confusion. But it's easy to fix.

Issue #4420, opened 2022-05-20
Assign color to imported transactions


omalleypat:
A feature I would find useful would be to have the option to assign a color to transactions when importing. In the import csv dialog, it would require adding the color picker box as in the new transaction dialog. This color would then be applied to all the transactions in the import.

tactilis:
+1 for this feature.

For import I would select one of the user defined colours that I have not used for highlighting transactions elsewhere.

This would be especially useful if the dates of some of the imported transactions overlapped with transactions already present. After the import, I be able to see exactly which transactions had been imported.

vomikan:
https://github.com/moneymanagerex/moneymanagerex/issues/1193

vomikan:
This request is implemented for QIF import.

![image](https://user-images.githubusercontent.com/6836805/170248339-09ef06ed-23db-479a-aa93-b2d678ed9e53.png)

Added the function of mass color change.
#4453

![image](https://user-images.githubusercontent.com/6836805/170248748-b862271b-13d4-4634-a62c-167b62bd7e43.png)

tactilis:
The layout looks good @vomikan. I'll try to do some testing if I get time in the next 2 days.

A couple of small side issues:
---
1. The translations don't work correctly for:
 
>  **Color**   (in the dialog labels)
>  **Colour** (in the pop-up menu)
  
When the application language is set to English it should say **Colour** everywhere.
Note that it _is_ consistent (i.e. **Color** everywhere) when the language is set to English(US).

I have noticed this before on the Colour pop-up on the Transaction Filter dialog but forgot to report it.

Shall I raise a separate issue?

---
2. The use of the word 'mass' or 'massive' when in English we would normally say 'bulk' when referring to updating multiple items.

> Added the function of mass color change.

In **customfieldlistdialog.cpp** there is:
> Line 188:         wxString::Format(_("This function will massive search & replace for \"%s\" custom field values\n"
> Line 263:     mainMenu->Append(new wxMenuItem(mainMenu, MENU_UPDATE_FIELD, _("&Massive content update ")));

These would be better as:
> Line 188:         wxString::Format(_("This function will bulk search & replace for \"%s\" custom field values\n"
> Line 263:     mainMenu->Append(new wxMenuItem(mainMenu, MENU_UPDATE_FIELD, _("&Bulk update ")));

Shall I raise a separate issue?

omalleypat:
Great work! Can this feature be added to csv import also?

[...]

PR #4464, opened 2022-05-25
fix(#4420): color for csv import

vomikan:


omalleypat:
@vomikan I tested it and it seems to be working. One little GUI issue is when you choose a color in the dialog, you can't change the color without unchecking the box. For example:

File>Import>CSV
Click color checkbox, click square and choose color number 1. 
Click color square again (for example to change to color number  2), nothing happens. 
Uncheck the checkbox, now you can click the square to choose color number 2. 

Can someone confirm if this is also happening on windows or Mac only? 
It seems to work as expected in the import QIF dialog, just not the CSV dialog.

vomikan:
@omalleypat this bug fixed. Thank you.

Issue #4577, opened 2022-06-07
Import XML crash


whalley:
MMEX version:
 - [x] 1.5.15
 - [ ] 1.6.x

Operating System:
 - [ ] Windows
 - [x] Mac OSX
 - [ ] Linux 

# Description of the bug

When invoking Import XML Files, mmex crashes

# Reproduction

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:

Access menu "File / Import / XML Files..."

Expected result:

Import dialog displayed

Actual result:

Program crash
[crash.txt](https://github.com/moneymanagerex/moneymanagerex/files/8856971/crash.txt)

PR #4578, opened 2022-06-07
fix(#4577): color button from other button size

whalley:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/4578)


whalley:
@vomikan Not sure where you are with the v1.5.15 release, but might be good to get this in if possible.

Issue #4439, opened 2022-05-23
comboboxes for categories, payees, accounts


vomikan:
I propose an alternative method of entering categories.

![image](https://user-images.githubusercontent.com/6836805/169801461-a4e3c1e4-4849-4a21-af8b-7eb7450d7c08.png)

#2232 #4419

whalley:
Are you suggesting wesupport this in all instances where we enter a category, e.g. Transaction Report and Transactions? If so, I'm supportive.

vomikan:
I suggest making such a choice for the transaction filter. But I don't see any barriers to doing this for transaction dialogs either.

gabriele-v:
What if there is a category with the colon inside? Or it was not possible to add it also within the old form (never tried)?

vomikan:
> What if there is a category with the colon inside?

What do you mean? category:with_colon:subcategory?

gabriele-v:
Yep, or catgeory:subactegory:withcolon

If users were able to insert category or subcategory with the old system, then it's already too late even if you prohibit it in the new one, because something won't work, unless you issue also a DB patch (that at the moment it's standby also for more important changes than this) and prohibit also everywhere else (import, WebApp etc..)

vomikan:
1) I can't enter colon. It seems autoreplace implemented.
![image](https://user-images.githubusercontent.com/6836805/171721825-7a97da71-c477-458f-a062-9825fd748945.png)
2) If you use a colon, it will negatively affect the import. I would recommend to such a user to remove colons from the category names.
3) Even if there is a colon, it does not affect this functionality in any way.
![image](https://user-images.githubusercontent.com/6836805/171723028-793441af-aeb7-48ea-a294-711efc7338d6.png)


![image](https://user-images.githubusercontent.com/6836805/171723050-601cbfa2-c4b9-4730-af85-5252f3a84ac4.png)

vomikan:
Moreover, it is now possible to calculate such categories with a single query.
![image](https://user-images.githubusercontent.com/6836805/171724119-6e45d261-8044-4998-beca-de8e474b3f46.png)

vomikan:
PS In this interface, you cannot create a new category, you can only select from an existing list. Or enter any text as a regular expression (a hidden option for advanced users).

gabriele-v:
Got it, was thinking it will be used also for adding, I don't see any drawback for it as well, a part from the colon.

[...]

PR #4580, opened 2022-06-08
fix(#4439): refactoring

vomikan:


vomikan:
@whalley please check the performance of this update on MAC. The sorting of elements could be broken, among other things.

whalley:
Works fine except for the sorting, it is reversed (and seems to be case sensitive).

vomikan:
@whalley could you, please, take care of the sorting? I can't simulate this on Windows.

Issue #4625, opened 2022-06-16
Version: 1.5.15 64-bit freezing


andybluenoes:
Just upgraded to thsi version, and it locks up (not responding) constantly when moving between accounts and sometimes transacations.

howff:
Likewise for me, definitely a regression since 1.5.14. Just clicking between accounts in 1.5.15 takes many seconds. Reverting to 1.5.14 restores the instant update when changing account. Exact same database. (Windows 10 64-bit)

andybluenoes:
> Likewise for me, definitely a regression since 1.5.14. Just clicking between accounts in 1.5.15 takes many seconds. Reverting to 1.5.14 restores the instant update when changing account. Exact same database. (Windows 10 64-bit)

Thanks for the tip, ive just rolled back and working fine again

vomikan:
@whalley , I can reproduce this problem. If you open a file with a large number of payees, then when switching to any account, the transaction filter is activated, where wxcombobox is filled in. this process is really too slow. I practically can't open a file with 7700 payees.

```
void mmComboBox::Create()
{
    unsigned int i = 0;
    wxArrayString auto_complite;
    for (const auto& item : all_elements_)
    {
        auto_complite.Add(item.first);
        this->Insert(item.first, i++); //Comment out this line and everything will work quickly.
    }
    this->AutoComplete(auto_complite);
    if (auto_complite.GetCount() == 1)
        Select(0);
}
```

vomikan:
@andybluenoes , @howff  could you please investigade fixed vertion? What is the performance in your opinion?

https://ci.appveyor.com/project/moneymanagerex/moneymanagerex/builds/43885767/job/ycasw2gpvhgghla0/artifacts

You can install the portable version from the zip archive in any folder. You don't have to reinstall the application.
Thank you in advance.

whalley:
Fix works on OK on Mac. But I have found a few additional funnies when compared to v1.5.14.

- The sort order for the Accounts and Payees is case-sensitive now - Related to the use of the std:map for holding the data, need to add a comparison function for this. **(Low issue)**
- All accounts are displayed via mmComboBoxAccount previously only open accounts were displayed **(Low issue)**
[...]

PR #4650, opened 2022-06-18
fix(#4625): speedup

vomikan:
@whalley I think I found an easy way to solve the problem. Please rate.

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/4650)


vomikan:
I like how it works, but if you close the dialog before the background process ends, the application will crash.

whalley:
@vomikan This does indeed look like an elegant way of solving the issue of very large amounts of data. I'll have a look at it more later.

PR #4678, opened 2022-06-20
fix(#4664)): bills dialog

vomikan:


tactilis:
In 

```
void mmBDDialog::setCategoryLabel()
{
    if (has_split)
    {
        fullCategoryName = _("Categories");

```

The use of "Categories", in the plural was appropriate because the transaction is split across multiple categories.

But now in the changed code:

```
void mmBDDialog::setCategoryLabel()
{
    if (has_split)
    {
        cbCategory_->SetLabelText(_("Split Transactions"));

```
I think we should say "Split Transaction", i.e. singular, because this text relates to a single transaction.

Issue #4620, opened 2022-06-16
Payee and Category new transaction defaults


renato-mmex:
MMEX version:
 - [x] 1.5.6 Beta1
 - [x] Windows

When entering a new entry, the last used category is no longer forgiven.

Is the bug reproducible? 
 - [x] Always

whalley:
Think we need to look at this logic again.

For Default Payee we have:

- None. Which results in a blank payee, seems good
- Last Used. Finds the last used payee for the account and displays that, seems good
- Unsed. Which results in a blank payee, **how is this different from 'None'?**

For Default Category we have:
- None. Which results in a blank category, seems good
- Last used for Payee. Which looks up the last used for the payee, seems good
- Unused - **Uses an "Unknown" category, creating one if not present. This will break current handling as it does not update the list for the combobox..... Do we really need this?**
- Use default for Payee. **Not convinced this is working.**



If UNUSED is specified and we

renato-mmex:
from my point of view, I would like to see the following options:

- no default (necessary ?)
- last used for this account

Reason: there are transactions that I do over and over again at the same place, such as grocery shopping, parking fees, gasoline, medications

**there is a category delimiter in the settings, it has no effect lately.**

whalley:
> from my point of view, I would like to see the following options:
> 
> * no default (necessary ?)
> * last used for this account
> 
> Reason: there are transactions that I do over and over again at the same place, such as grocery shopping, parking fees, gasoline, medications

Is that for Payee (which we already have) and category (new option)

> **there is a category delimiter in the settings, it has no effect lately.**

This works fine for me, I can change the delimiter and the application conforms.

renato-mmex:
> > from my point of view, I would like to see the following options:
> > 
> > * no default (necessary ?)
> > * last used for this account
> > 
> > Reason: there are transactions that I do over and over again at the same place, such as grocery shopping, parking fees, gasoline, medications
> 
> Is that for Payee (which we already have) and category (new option)

Correct, many times the payee has a direct reference to the category. Example: Tamoil>gasoline, Lidl>groceries
> 
> > **there is a category delimiter in the settings, it has no effect lately.**
> 
> This works fine for me, I can change the delimiter and the application conforms.

Sorry, have overlooked that **MMEX must be restartet**, that the delimiter is reset.

By the way, I find the old representation of the category selection more user-friendly, moreover we have now two different representations
[...]

PR #4697, opened 2022-06-22
fix(#4620): existing tran payee chg and tidy

whalley:

---
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/4697)


whalley:
@vomikan Whilst I thought it useful to change the category based on the defaults when editing existing transactions it goes against the definition of the option setting, which only applies to new transactions... So going to revert with just the tidy up.

Issue #5039, opened 2022-09-04
CSV files import: error when amount field is selected


artemiogr97:
Description:

Open the csv import dialog
Tyr to import the data  using the amount field
The following two options don't work, all the values are imported as withdrawals:
![image](https://user-images.githubusercontent.com/57588855/188308745-bd2e42ba-a9d0-43df-bbe3-a0fb9331a260.png)

whalley:
Seems to work fine for me.

When you say "The following two options don't work", which 2 options are you referring to - the field you have shown as 3 options? As the 'Positive of type has 'Deposit' implies you have a type field with the amount present also.

Can you provide an example CSV and the full import settings.

artemiogr97:
Options "Positive values are deposits" and "Positive values are withdrawals" doesn't work.

My import options are the following:
![image](https://user-images.githubusercontent.com/57588855/188313499-44503011-50d8-4e41-a56e-aefe845c7fb4.png)
Example of csv file:
[test.csv](https://github.com/moneymanagerex/moneymanagerex/files/9484672/test.csv)
When the first option is selected I get the following result after the import:
![image](https://user-images.githubusercontent.com/57588855/188313645-f2f3f0bf-7f30-453c-aa0a-5f354d288a92.png)
Expected result:
![image](https://user-images.githubusercontent.com/57588855/188313806-3ed4309a-701e-4a80-8d0c-3bf71a4c508c.png)

artemiogr97:
This problem seems to happen only when the field "Type" is selected at the same time that any of the first two options are selected

whalley:
Why would you use the options "Positive values are deposits" and "Positive values are withdrawals" if a transaction type is defined? As you have seen these options are only used if the type field is not present. If a type field is present I think it makes sense for that to take precedent.

tactilis:
@artemiogr97, to reinforce what @whalley has said...

Regardless of the column being called Type, you don't actually have a Type data field in your test csv file.

Lines 4 and 5 both have a Type of 'test', which is unrelated to whether the Amount value is a deposit or a withdrawal. You **must** therefore set column 4 to Don't Care, so that MMEX Import ignores it.

![image](https://user-images.githubusercontent.com/4060092/188330639-e58635be-6171-4a1b-9926-39ffa7297526.png)

artemiogr97:
I just marked that column as "Don't care" and the behavior is what I expected, thanks for the help.
[...]

PR #5041, opened 2022-09-04
fix(#5039) error when amount field is selected

artemiogr97:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/5041)


whalley:
As advised against https://github.com/moneymanagerex/moneymanagerex/issues/5039 the behaviour is correct

Issue #4631, opened 2022-06-17
Regular Expressions in transaction filter notes


gabriele-v:
Hello, considering that it has already been implemented for payees/categories, would be possible to have the lookup with Regular Expressions also for the notes?

I know that there are already some wild cards, but it would be useful to have RegEx as well.
![image](https://user-images.githubusercontent.com/7106290/174260887-9cffe3b7-7917-46e2-8ba1-7284e41ec201.png)

I think we may consider as suggested by @whalley to have somethign specific for regex, like Regex[...] to distinguish between standard searches, or enable a checkbox on the side which means use regex.

gabriele-v:
Yes, additional gui element would be useful/clearer but it's not mandatory, we just need a pattern or text to define when is standard search and when is regex

whalley:
> I think we may consider as suggested by @whalley to have somethign specific for regex, like Regex[...] to distinguish between standard searches, or enable a checkbox on the side which means use regex.

Yes, my logic is that power users don't need a checkbox and can enter the text regex[...] (or whatever notation we choose) to distinguish it from a normal match. But 'casual' users don't have to worry about it.

vomikan:
I am in favor of regular expressions being hidden from the eyes of non-advanced users. And for not loading the UI with additional elements. This is how it is now implemented for the Payee and Category fields.
But for Notes, it must be determined in advance which algorithm is used by RegEx or Matches.

n-stein:
Not sure if we were ready to implement this yet as it hasn't had any action in a few months, but I had a go at it this using the regex[...] syntax discussed above. I have put the changes in PR #5142 for review.

n-stein:
New PR #5144

PR #5142, opened 2022-09-29
Fix(#4631): allow regex in notes filter

n-stein:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/5142)

### DIFF
--- a/src/reports/budget.cpp
+++ b/src/reports/budget.cpp
@@ -57,7 +57,8 @@ void mmReportBudget::SetBudgetMonth(wxString budgetYear, wxDateTime& startDate, 
         wxString monthStr = pattern_month.GetMatch(budgetYear, 2);
         int month = wxAtoi(monthStr) - 1;
         startDate.Add(wxDateSpan::Months(month));
-        endDate.Add(wxDateSpan::Months(month));
+        endDate = startDate;
+        endDate.Add(wxDateSpan::Months(1));
     }
 }
 


src/filtertransdialog.cpp
```diff
@@ -1244,9 +1246,19 @@ bool mmFilterTransactionsDialog::mmIsRecordMatches(const Model_Checking::Data &t
     else if (mmIsNumberChecked() && (mmGetNumber().empty() ? !tran.TRANSACTIONNUMBER.empty()
         : tran.TRANSACTIONNUMBER.empty() || !tran.TRANSACTIONNUMBER.Lower().Matches(mmGetNumber().Lower())))
         ok = false;
-    else if (mmIsNotesChecked() && (mmGetNotes().empty() ? !tran.NOTES.empty()
-        : tran.NOTES.empty() || !tran.NOTES.Lower().Matches(mmGetNotes().Lower())))
-        ok = false;
+    else if (mmIsNotesChecked()) {
```
whalley, line 1249: If mmIsNotesChecked() = TRUE then the checking against the remainder checkboxes will not be processed. That is why the convoluted if statement was there I suspect.

src/filtertransdialog.cpp
```diff
@@ -1244,9 +1246,19 @@ bool mmFilterTransactionsDialog::mmIsRecordMatches(const Model_Checking::Data &t
...
+        if (mmGetNotes().empty() && !tran.NOTES.empty()) ok = false;
+        else if (!mmGetNotes().empty()) {
+            if (tran.NOTES.empty()) ok = false;
+            else if (mmGetNotes().StartsWith("regex[")) {
+                std::string pattern = mmGetNotes().SubString(6, mmGetNotes().find_last_of(']') - 1).ToStdString();
```
whalley, line 1254: As it will allow regex[....    and regex[....].  Perhaps we just use regex:......

src/filtertransdialog.cpp
```diff
@@ -1244,9 +1246,19 @@ bool mmFilterTransactionsDialog::mmIsRecordMatches(const Model_Checking::Data &t
...
+                std::regex regex(pattern, std::regex_constants::icase);
+                if (!std::regex_search(tran.NOTES.ToStdString(), regex)) ok = false;
+                notesEdit_->SetValue("regex[" + pattern + "]");
```
whalley, line 1257: What's the reasoning behind this change to the field?

src/filtertransdialog.cpp
n-stein: Good catch. I moved this to the end of the checks so nothing else is left to process.

src/filtertransdialog.cpp
n-stein: The pattern is derived from whatever is between the [ ], but the user could type something like "regex[...]extratexthere". In this code the pattern is still the [...] and the "extratexthere" is simply ignored, but when you see the transaction report under "Filtering Details" it shows the filter is "Notes: regex[...]extratexthere" rather than "Notes: regex[...]", so this bit was to get the report display in sync with the applied filter.

n-stein: Switching to regex: would eliminate the need for this since we don't need to find the end of the pattern string.

src/filtertransdialog.cpp
n-stein: Changed to regex:

PR #5163, opened 2022-10-05
HiDPI Support in Toolbar

gerstrong:

-> Not sure how to update this pot file. Can someone give me an example?

By direct accessing SVG resources and using wxBitmapBundle it possible to get proper scaling and better looking icons on HiDPI displays.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/5163)


whalley:
@gerstrong Thanks for the PR. Some thoughts...  (from a quick read)

 I think it would be best to retain `static wxSharedPtr<wxBitmap> programIcons[numSizes][MAX_PNG]` but update for `wxBitmapBundle` . At present `static std::array<wxSharedPtr<wxBitmapBundle>, MAX_PNG> programIcons` only holds a single icon at a defined size.  We may use the same icon at different sizes within the program. I do however note your fallback. Don't need to legacy handling.

If we loaded up `programIcons` with the `wxBitMapBundle`'s at the various sizes then should we just continue to use `mmBitmap` in the rest of the application. Don't need to legacy handling.

gerstrong:
So shall I rename both? wxBitmapBundle is meant to be used for multiple resolutions, with SVG it directly derives whatever size is required. The given size is meant to be the default size, which additionally gets scaled depending on the DPI settings. It think that is the idea of the wxBundleBitmap Class.

I left the older array of wxBitmap Ptrs in for fallback purposes. I think it could be removed, but I am not 100% sure yet. Apparently all the themes use SVG, but the engine does not allow PNG directly as part of a theme package. At that point I need to study the code a bit closer. Just wanting to get sure I am not breaking anything, you know... :-)

whalley:
> wxBitmapBundle is meant to be used for different resolutions. You indicate a default size for your widget containers like for a toolbar and that size will be used.

Understood, but we can use the same icon for toolbar, buttons, etc. Each with different 'default' sizes. As we need to specify this at Bitmap bundle creation time I thought it best to mimic what we currently do and create/access based on SVG/default size.

whalley:
How about just loading the SVGs into memory and then creating the Bitmap Bundles on request at the default size?

gerstrong:
The icons are loaded when the program starts, that is the architecture, afaik.  Tbh I don't feel confident to apply such changes, and if I would do it, I think it should be part of a different ticket. For now I will check which icons and sizes will be loaded, load SVG directly and make this available through ```wxBitmapBundle```, which the current way how the programs loads icons loads considered.

I generally like the idea of loading on demand, but let's do that in another step.

whalley:
OK, I'm good with that. Can look to refactor later if we feel the need.

gerstrong:
Okay, I extended the code a bit. Let me know your thoughts.

whalley:
Looks good. A few questions....

1. Can we now ditch mmBitMap and replace all usage with mmBitMapBundle now?
2. I note you are using GetDPIScaling when dealing with wxImageList? I see that use of this and SetImageList is now not recommended - https://docs.wxwidgets.org/trunk/classwx_image_list.html. Don;t think we need to use GetDPIScaling if we were to refactor to use [SetImages](https://docs.wxwidgets.org/trunk/classwx_with_images.html#aa1da4c4d0642e3f88c830faa588a4865) instead?

gerstrong:
Thanks :-)
About your questions:
1. I don't think we are ready to ditch ```mmBitmap``` yet. There are still too many dependencies. 
2. I tried using ```SetImages``` but I got blurry Icons again. Using the ```mmBitmap``` function with my additions gives a better result. In the end ```mmBitmap``` is now also a wrapper that derives Bitmaps with higher resolutions from ```wxBitmapBundle```. To give an example ```wxBitmapBundle``` and ```mmBitmapBundle``` work directly well with ```wxAuiToolbar```, but with ```wxListBook```, although we can use ```wxVector<wxBitmapBundle>```, the end result still is blurry. I am not sure, if it is related to GTK, but I still see some differences here with wxWidgets for some reason.

Don't get me wrong, reducing code in general and throwing out what is not required, I am always for, but I still see some wx classes struggling with ```wxBitmapBundle```.

Issue #910, opened 2016-09-24
Notes in split categories


J316:
It would be very useful to add notes for each entry in split categories entries

vomikan:
https://sourceforge.net/p/moneymanagerex/feature-requests/340/

guanlisheng:
hi @J316  we might consider this in v2.0.

J316:
Would be awesome. Thanks

anandasudarshan:
+1 very useful to have this

whalley:
We can now add this feature due to the recent DB change.

Any proposals on how to add the notes to this?

Some suggestions:

- Another field under the Category/Amount?
- A button on the right that the user can press to add/edit notes, and can view when hovered over?
- any others?

<img width="502" alt="CleanShot 2022-09-30 at 13 34 54@2x" src="https://user-images.githubusercontent.com/487345/193270808-c2cd5153-899c-477d-b424-19e83c2391cd.png">

tactilis:
> Any proposals on how to add the notes to this?

Thinking about it. Will produce a mock-up for dicussion.
 
 
 

Do you intend including this capability too?

![image](https://user-images.githubusercontent.com/4060092/193356495-fba96509-23cf-48be-a9dd-acc8908c43a1.png)

whalley:
> Do you intend including this capability too?

I guess we could.

tactilis:
First attempt, for discussion:

![image](https://user-images.githubusercontent.com/4060092/193450947-d5c27a8b-cd92-4b65-aa39-118c0eedbadf.png)

1. The Notes field is always shown next to the Amount field but greyed out for not active rows.

2. The initial (default) width of the dialog is such that you can see ~25 characters, probably enough for most short notes.

3. The width of the Notes field increases when the user increases the width of the dialog.

4. Clicking in a Notes field enables the note to be edited.

5. If the entered note is too long for the field, the text will scroll left/right as the enters/deletes characters or moves the text caret.

6. When the text caret is in a Notes field, the **[...]** button is displayed next to that row.

7. Clicking the  **[...]** button allows one of the frequently used notes to be selected and entered (similar to the functionality on the Edit Transaction dialog).

8. If a multi-line note is entered using the  **[...]** button (or by typing; see below) then all the lines of the note are displayed on one line as is done in the Notes column of the Transaction list. A one, two, and a three line note is shown on the dialog above.

[...]

PR #5178, opened 2022-10-12
fix(#910): split notes

whalley:


whalley:
@vomikan Not sure why the Button image is not appearing on Windows - the button is there in teh right place as I can click on it and the tooltip works. Also the FocusChange does not seem to the be reacting on Windows causing the `row_num_` to be wrong (think that is what causes the crash). Seems to work fine on MacOS. If you get chance can you have a look, otherwise I'll have to get a development env up and running as I can't V things working on my Win 11 ARM VM yet.
CleanShot 2022-10-13 at 01 08 42

whalley:
Can't quite work out why a button with only an image will not display on Windows. Maybe something to do with the scrollable panel (wxScrollableWindow). Anyway adding text to the button makes it visible! So have tweaked the implementation to accommodate this. Maybe consider change if somebody can work out what is going on, or just leave as-is.

Issue #2963, opened 2021-02-24
Add a Deleted/Trash transaction feature


whalley:
# Description of the feature

As suggested by @vomikan in https://github.com/moneymanagerex/moneymanagerex/issues/2918#issuecomment-785143362

We should consider supporting a deleted transaction view from where the user could undone, should be strigh tforward now we have the All Transactions view.

1. Consider auto-delete after 'n' days, and/or delete now.
2. Perhaps color the translation view in light red so user knows he is in this view.
3. Only real option should be to move a transaction back to a given account, all otehr actiosn such as editing must be done once 'un-deleted'
3. Can use this icon
![image](https://user-images.githubusercontent.com/487345/109028792-126c1580-76ba-11eb-82d2-a683fdc978d0.png)
https://www.svgrepo.com/svg/316810/duotone-trash-2

n-stein:
Give PR #5202 and themes PR #3 a go, see what you think.

whalley:
Ecellent.

Couple of quick observations, will do more comprehensive comments when I get more time.

- The deleted transactions are still being be counted in the account balance
- It is possible to edit a deleted transaction, we should disable edit. The transaction dialog should be available but it should not be possible to edit.
- Need to decide what we do with deleted transactions when payee/categories are deleted. E.g. at present it will block deletion of a payee if a transaction using it exists in the deleted state. I think we should allow deletion and remove the deleted transaction.

whalley:
> Hold off this feature until the database changes in https://github.com/moneymanagerex/moneymanagerex/issues/4481 and MMEX v2.0.0 so that deletion can be flagged independently of status? It would then be possible to mark the date/time of deletion so that the deleted transactions can be appropriately ordered.

Good idea. I've already said we should hold off including this in v1.6.0 as we need to do a lot of regression testing and get this right. So we could update the DB in the next release. It is probably cleaner as you say to have a separate marker.

tactilis:
We should also consider whether Undelete Transaction(s) is going to be the full extent of the Undo feature that is periodically requested by users (without much thought about its ramifications).

If we are eventually going to have full-featured Undo, then we should think about how Undelete fits into this.

whalley:
> If we are eventually going to have full-featured Undo, then we should think about how Undelete fits into this.

Can't see that ever happening. Far too much of a re-write for that.

n-stein:
> It is probably cleaner as you say to have a separate marker.

[...]

PR #5202, opened 2022-10-19
feat(#2963): Deleted Transactions view

n-stein:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/5202)


whalley:
@n-stein  Great. I'll test this and feedback comments. **Suggest we leave for v1.6.1 as it will will require quite a bit of regression testing to make sure the new transaction status is handled/ignored by existing logic/GRM reports/etc.**

PR #5244, opened 2022-10-28
Recurring Transaction Check Fix

intcooper:
This is a fix (plus a small improvement) of the Recurring Transaction Check.
Currently, the check is performed for all recurring transactions, regardless of whether they will be applied.
This results in an incorrect calculation in case the sum of all the transactions, even the ones that would not be immediately applied, would bring the balance below the minimum balance.

In short:
- The check is now performed just before the transaction is applied.
- The message box shows some details to help the user decide

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/5244)

### DIFF
--- a/src/mmframe.cpp
+++ b/src/mmframe.cpp
@@ -505,11 +505,10 @@ void mmGUIFrame::OnAutoRepeatTransactionsTimer(wxTimerEvent& /*event*/)
     for (const auto& q1 : bills.all())
     {
         bills.decode_fields(q1);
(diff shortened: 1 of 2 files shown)


whalley (changes requested):
Looks good, but the strings need to be enclosed in _( ) to signify that language translation applies.

intcooper:
> Looks good, but the strings need to be enclosed in _( ) to signify that language translation applies.

Done.
I changed my approach to minimise the number of strings to be translated.

Issue #5290, opened 2022-11-01
New transaction - Incorrect minimum balance check


intcooper:
MMEX version:
 - [ ] 1.5.20
 - [ ] 1.5.21
 - [x] 1.6.0
 - [ ] 1.6.1 Beta
 - [ ] Other (please specify)

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [x] Windows
 - [ ] Mac OSX
 - [ ] Linux 

# Description of the bug

A transfer created from the target account page leads to an incorrect check for minimum balance (and credit limit).
In fact, the balance used for the calculation is the target account one.

# Reproduction

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:
- Make sure to define a minimum balance for an account (account1)
- Navigate to a different account page (account2)
- Create a new transaction (or duplicate an existing one) 
- Change the transaction type to transfer
- Change the source account to account1
- Change the target account to account2
- Apply

Expected result:
The minimum balance check is performed against the source account balance.

Actual result:
The minimum balance check is performed against the target account balance.

# Additional information

If you have a backtrace for a crash or a warning, paste it here.

intcooper:
I opened pull request #5291 with a fix for the issue.

intcooper:
@whalley should I wait until I verify with the 1.6.1 release to close the issue?
What is the general approach to bugs and features?

tactilis:
> should I wait until I verify with the 1.6.1 release to close the issue?
What is the general approach to bugs and features?

My approach is that I independently test the first Beta build that includes the fix and, if all is OK, I comment on the Issue along the lines of:

> Tested 1.6.1-Beta.1 64-bit commit: dd56bdd (2022-11-01)     _{info copied from the About dialog}_
>
> Confirmed Fixed

If I raised the issue, I then close it.

If someone else raised the issue, I wait for them, @whalley, or @vomikan to close it.

PR #5291, opened 2022-11-01
#5290 Fix new transaction data validation

intcooper:
Fixes the validation of a new transaction against the minimum balance and the credit limit for the target account.

- The check is performed only for withdrawals and transfers
- The reference account is specifically the source one; this fixes an issue when the transaction is created from the target account page (the balance was incorrectly the target account's).

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/5291)

### DIFF
--- a/src/transdialog.cpp
+++ b/src/transdialog.cpp
@@ -702,32 +702,32 @@ bool mmTransDialog::ValidateData()
     //Checking account does not exceed limits
     if (m_new_trx || m_duplicate)
     {
-        bool abort_transaction = false;
-        double new_value = m_trx_data.TRANSAMOUNT;
-
-        if (m_trx_data.TRANSCODE == Model_Checking::all_type()[Model_Checking::WITHDRAWAL])
+        if ((m_trx_data.TRANSCODE == Model_Checking::all_type()[Model_Checking::WITHDRAWAL]) ||
+            (m_trx_data.TRANSCODE == Model_Checking::all_type()[Model_Checking::TRANSFER]))
         {
-            new_value *= -1;
-        }
+            const auto fromAccount{ Model_Account::instance().get(m_trx_data.ACCOUNTID) };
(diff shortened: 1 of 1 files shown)


src/transdialog.cpp
```diff
@@ -702,32 +702,32 @@ bool mmTransDialog::ValidateData()
     //Checking account does not exceed limits
     if (m_new_trx || m_duplicate)
     {
-        bool abort_transaction = false;
-        double new_value = m_trx_data.TRANSAMOUNT;
-
-        if (m_trx_data.TRANSCODE == Model_Checking::all_type()[Model_Checking::WITHDRAWAL])
+        if ((m_trx_data.TRANSCODE == Model_Checking::all_type()[Model_Checking::WITHDRAWAL]) ||
```
vomikan, line 705: m_trx_data.TRANSCODE != Model_Checking::all_type()[Model_Checking::DEPOSIT]

src/transdialog.cpp
```diff
@@ -702,32 +702,32 @@ bool mmTransDialog::ValidateData()
...
+            (m_trx_data.TRANSCODE == Model_Checking::all_type()[Model_Checking::TRANSFER]))
         {
-            new_value *= -1;
-        }
+            const auto fromAccount{ Model_Account::instance().get(m_trx_data.ACCOUNTID) };
```
whalley, line 708: The fromAccount is already available in `account` (see line 594)

whalley, line 708: We don't need this as we already have the account pointer in `account`.

intcooper: I agree. I will just use `account`.

src/transdialog.cpp
intcooper: I figured that Model_Checking defines constants for the types; see my last commit.

Issue #5727, opened 2023-03-03
Buttons in Language are showing using the system language, instead of the language set in MMEX


ovari:
Money Manager Ex
• Version: 1.6.4-Beta.2 64-bit
• Built on: Mar 3 2023 19:53:26
• Database version: 17 (aes128cbc)

# Description of the bug
Buttons in `Language` are showing using the system language, instead of the language set in MMEX.

# Reproduction

1. The operating system language is set to Hungarian.
![image](https://user-images.githubusercontent.com/17465872/222853449-c74d1d27-e4f4-4151-b339-f1c0bbb1b6a5.png)
2. Open MMEX and set the language to English.
3. Restart MMEX and click `User Interface Language`
4. Should the buttons in the `Language` dialog show in English, i.e. `Cancel`, instead of Hungarian, i.e. `Mégse`?
![image](https://user-images.githubusercontent.com/17465872/222853630-d66ae621-d887-48e0-9c8f-a5b062354cbc.png)

Is the bug reproducible? Always

Thank you

ovari:
Another test with:
* Operating system language = English, Australia
* MMEX language = Hungarian

Should the buttons in MMEX show in Hungarian the MMEX is set to Hungarian, instead of the operating system language?

![image](https://user-images.githubusercontent.com/17465872/222855255-a12288d6-bbf0-4fff-81c0-89ce3b828449.png)

Thank you

ovari:
It seems that the `&OK` is not translated, as in Hungarian it should show as `&Rendben`

vomikan:
This can be fixed. An example of the code is here:

https://github.com/moneymanagerex/moneymanagerex/blob/master/src/mmSimpleDialogs.cpp#L848-L865

ovari:
Thank you for the link.

Can https://github.com/moneymanagerex/moneymanagerex/blob/master/src/mmSimpleDialogs.cpp#L848-L865 be copied and pasted into the file at https://github.com/moneymanagerex/moneymanagerex/blob/master/src/appstartdialog.cpp#L185 using a web browser?

n-stein:
> Can https://github.com/moneymanagerex/moneymanagerex/blob/master/src/mmSimpleDialogs.cpp#L848-L865 be copied and pasted into the file at https://github.com/moneymanagerex/moneymanagerex/blob/master/src/appstartdialog.cpp#L185 using a web browser?

[...]

PR #5735, opened 2023-03-08
Update appstartdialog.cpp

ovari:

https://github.com/moneymanagerex/moneymanagerex/issues/5727#issuecomment-1455247420

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/5735)

### DIFF
--- a/src/appstartdialog.cpp
+++ b/src/appstartdialog.cpp
@@ -154,6 +154,10 @@ void mmAppStartDialog::CreateControls()
     {
         mmToolTip(itemButton61, wxString::Format(_("Open the previously opened database : %s"), val));
     }
+    wxButton* ok = static_cast<wxButton*>(FindWindow(wxID_OK));
+    if (ok) ok->SetLabel(_("&OK "));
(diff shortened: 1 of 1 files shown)


src/appstartdialog.cpp
```diff
@@ -154,6 +154,10 @@ void mmAppStartDialog::CreateControls()
     {
         mmToolTip(itemButton61, wxString::Format(_("Open the previously opened database : %s"), val));
     }
+    wxButton* ok = static_cast<wxButton*>(FindWindow(wxID_OK));
```
vomikan, line 157: This fix does not make sense, because the corresponding code is already a little higher.

```
    m_buttonClose = new wxButton(this, wxID_OK, _("&OK "));
    m_buttonExit = new wxButton(this, wxID_EXIT, _("E&xit "));
```

n-stein: Agreed. This is not the issue. Its here https://github.com/moneymanagerex/moneymanagerex/blob/master/src/appstartdialog.cpp#L212 where wxGetSingleChoice doesn't translate 'OK'.

PR #5952, opened 2023-05-30
Remove charset=utf-8 from ContentType in WebApp POST request 

### LINKED ISSUE #5951: WebApp Interaction broken - 400 Bad Request
MMEX version:
 - [x] 1.6.3
 - [x] current master 

Operating System:
 - [ ] Windows
 - [ ] Mac OSX
 - [x] Linux 

# Description of the bug

After my webhoster migrated internally, the interaction between MMEX and the WebApp stopped working. After some investigation with WireShark, it turns out the nginx is sending error code 400 Bad Request and WireShark is saying that the POST request from MMEX is a malformed packet.

Further investigation with curl revealed that the culprit was the `charset=utf-8` in the ContentType. Removing this fixes the issue.

# Reproduction

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:
Click "Tools/Refresh WebApp..."

Expected result:
WebApp is updated with Accounts, Payees and Categories

Actual result:
Error message displayed in MMEX to check WebApp settings and connection + WebApp completely cleared

moritzhoewer:
Fixes #5951

### DIFF
--- a/src/webapp.cpp
+++ b/src/webapp.cpp
@@ -150,7 +150,7 @@ Content-Disposition: form-data; name="MMEX_Post"
 --Custom_Boundary_MMEX_WebApp--
 )";
 
-    const auto ContentType = "Content-Type: multipart/form-data; charset=utf-8; boundary=Custom_Boundary_MMEX_WebApp";
+    const auto ContentType = "Content-Type: multipart/form-data; boundary=Custom_Boundary_MMEX_WebApp";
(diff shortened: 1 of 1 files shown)


moritzhoewer:
This is my first pull request in this repo. I tried to follow the contributing guidelines, but please notify me if I did something wrong so I can do better next time.

Issue #5881, opened 2023-05-02
Remove `Dialog` and `Window` from source strings


ovari:
Please remove `Dialog` and `Window` from en source strings.

The words `Dialog` and `Window` are superfluous.

Please see comments in translation at https://crowdin.com/translate/moneymanagerex/470/en-pt?filter=basic&value=0#784842

What do you think?

Thank you

tactilis:
> Please remove Dialog and Window from en source strings.

Yes please.

- Currency Dialog
- Relocate Category Dialog
- Relocate Payee Dialog

Are there any others?

ovari:
Examples:

1. `Relocate Category Dialog` → `Relocate Category`
https://crowdin.com/translate/moneymanagerex/470/en-hu#784814
1. `Relocate Payee Dialog` → `Relocate Payee`
https://crowdin.com/translate/moneymanagerex/470/en-hu#784842
1. `Transactions Dialog` → `Transactions`
https://crowdin.com/translate/moneymanagerex/470/en-hu?filter=basic&value=0#q=dialog
1. `Tip: You can modify some runtime behavior of MMEX by changing the options in the Options Dialog. ` → `Tip: Change MMEX Options to modify some runtime behaviors. `
Change first-person to third-person.
https://crowdin.com/translate/moneymanagerex/470/en-hu#783808
1. `Organize Categories Dialog Tip: Pressing the h key will cycle through all categories starting with the letter h` → `Organize Categories Tip: Pressing the h key will cycle through all categories starting with the letter h`
https://crowdin.com/translate/moneymanagerex/470/en-hu#783814
1. `Organize Categories Dialog Tip: Pressing 2 key combination will cycle through all categories starting with that key combination. Example: Pressing ho will select Homeneeds, Home, House Tax, etc...` → `Organize Categories Tip: Pressing 2 key combination will cycle through all categories starting with that key combination. Example: Pressing ho will select Homeneeds, Home, House Tax, etc…`
Change 3-character ellipsis `...` to a 1-character ellipsis `…`
https://crowdin.com/translate/moneymanagerex/470/en-hu#783816
1. `Show the Options Dialog` → `Show Options`
https://www.microsoft.com/en-us/language/Search?&searchTerm=show%20options&langID=382&Source=true&productid=undefined
https://crowdin.com/translate/moneymanagerex/470/en-hu#784152
1. `&Show App Start Dialog` → `&Show Options menu at Startup`
[...]

PR #5980, opened 2023-06-04
Some cleanups

gerstrong:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/5980)

### DIFF
--- a/src/transdialog.h
+++ b/src/transdialog.h
@@ -49,8 +49,7 @@ public:
         , int transaction_id
         , double current_balance
         , bool duplicate = false
-        , int type = Model_Checking::WITHDRAWAL
-        , const wxString& = "");
+        , int type = Model_Checking::WITHDRAWAL);
 
     bool Create(wxWindow* parent
         , wxWindowID id = wxID_ANY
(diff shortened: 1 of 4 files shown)


src/transdialog.h
```diff
@@ -44,15 +44,13 @@ class mmTransDialog : public wxDialog
...
-        wxWindow* parent
+    mmTransDialog(wxWindow* parent
         , int account_id
         , int transaction_id
         , double current_balance
         , bool duplicate = false
         , int type = Model_Checking::WITHDRAWAL
-        , const wxString& name = "Transactions Dialog"
-    );
+        , const wxString& = "Transactions Dialog");
```
ovari, line 53: 1. Should this be translatable?
2. If so, should `Dialog` be removed? https://github.com/moneymanagerex/moneymanagerex/issues/5881

Thank you

tactilis: The string "Transactions Dialog" does not appear in the UI.

It is shown to the user as "New Transaction" or "Edit Transaction"

gerstrong: The variable name is not used at all by the constructor, so it can be removed.

Issue #5963, opened 2023-06-01
Online update in Currency Dialog doesn't work


Jerry23011:
MMEX version:
 - [ ] 1.6.0
 - [ ] 1.6.1
 - [x] 1.6.2
 - [ ] 1.6.3 Beta
 - [ ] Other (please specify)

Operating System:
 - [ ] Windows
 - [x] macOS 13
 - [ ] Linux 

# Description of the bug

When I click `online update` in the Currency Dialog, the popup says "Error Nothing to update"
I can update each currency individually by selecting them and clicking the update button on below the currency history view.

# Reproduction

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:
1. Open the Currency Dialog
2. Click `Online update`
Expected result:
The currency will update
Actual result:
The popup says there's an error
# Additional information
<img width="699" alt="截屏2023-06-01 21 13 27" src="https://github.com/moneymanagerex/moneymanagerex/assets/89069957/b1a8ddf8-8fc8-45d3-a7c2-09aec57cb00c">

n-stein:
Same issue as #5894. Fetching rates for all currencies uses the Yahoo Finance v7 API which no longer functions in any released version of MMEX while history download uses the v8 API which still works.

This has already been fixed for 1.6.4.

Jerry23011:
Thank you, @n-stein

Jerry23011:
> This has already been fixed for 1.6.4.

Hi, I downloaded the release build of 1.6.4, yet it's still not working?

n-stein:
For those outside the US there were additional fixes required which were corrected after the official 1.6.4 release. Try the 1.6.5 beta from June 26th here: https://link.ipx.co.uk/mmex-mac-betas

n-stein:
I'm not able to reproduce this issue on Windows. @whalley any suggestions?

Heny67:
> For those outside the US there were additional fixes required which were corrected after the official 1.6.4 release. Try the 1.6.5 beta from June 26th here: https://link.ipx.co.uk/mmex-mac-betas

Hi n-stein!

Is there such an improved beta for linux?  Because it doesn't throw an error on 1.6.4, but the exchange rate doesn't update, it doesn't download the real exchange rate.

n-stein:
Linux build here:

[...]

Issue #5953, opened 2023-05-31
MMEX cannot build an aarch64 due to cmake flags being set


joshua-stone:
MMEX version:
 - [ ] 1.6.0
 - [ ] 1.6.1
 - [ ] 1.6.2
 - [X] 1.6.3
 - [ ] Other (please specify)

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [ ] Windows
 - [ ] Mac OSX
 - [X] Linux 

# Description of the bug

A patch is currently needed to enable MMEX to build on aarch64:

https://github.com/flathub/flathub/pull/4184

```diff
diff --git a/3rd/CMakeLists.txt b/3rd/CMakeLists.txt
index b49403c6c..b9a349bd0 100644
--- a/3rd/CMakeLists.txt
+++ b/3rd/CMakeLists.txt
@@ -225,7 +225,7 @@ if(NOT wxSQLite3_FOUND OR NOT (WXSQLITE3_HAVE_CODEC OR MMEX_ENCRYPTION_OPTIONAL)
     endif()

     if(NOT MSVC) # wxSQLite3 security ext doesn't require C 99 for MSVC
-        set (CMAKE_C_FLAGS "-msse4.2 -maes")
+        set (CMAKE_C_FLAGS "")
         if(";${CMAKE_C_COMPILE_FEATURES};" MATCHES ";c_std_99;")
             target_compile_features(SQLite3 PRIVATE c_std_99)
         elseif(";${CMAKE_C_COMPILE_FEATURES};" MATCHES ";c_restrict;")
```

# Reproduction

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:

Run a flatpak build for aarch64 as shown in the [app submission pull request](https://github.com/flathub/flathub/pull/4184).

Expected result:

MMEX builds on aarch64 as well as x86_64 using the same build flags.

Actual result:

MMEX fails to build on aarch64 with the following output:

```
cc: error: unrecognized command-line option ‘-msse4.2’
cc: error: unrecognized command-line option ‘-maes’
make[2]: *** [3rd/CMakeFiles/Lua.dir/build.make:188: 3rd/CMakeFiles/Lua.dir/lua/ldebug.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
[...]

PR #6023, opened 2023-06-16
fix(#5963): add arm64

whalley:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6023)

### DIFF
--- a/3rd/CMakeLists.txt
+++ b/3rd/CMakeLists.txt
@@ -225,7 +225,7 @@ if(NOT wxSQLite3_FOUND OR NOT (WXSQLITE3_HAVE_CODEC OR MMEX_ENCRYPTION_OPTIONAL)
     endif()
 
     if(NOT MSVC) # wxSQLite3 security ext doesn't require C 99 for MSVC
-        if(";i386;i686;amd64;x86_64;" MATCHES ";${CMAKE_HOST_SYSTEM_PROCESSOR};")
+        if(";i386;i686;amd64;x86_64;arm64;" MATCHES ";${CMAKE_HOST_SYSTEM_PROCESSOR};")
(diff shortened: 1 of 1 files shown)


3rd/CMakeLists.txt
```diff
@@ -225,7 +225,7 @@ if(NOT wxSQLite3_FOUND OR NOT (WXSQLITE3_HAVE_CODEC OR MMEX_ENCRYPTION_OPTIONAL)
     endif()
 
     if(NOT MSVC) # wxSQLite3 security ext doesn't require C 99 for MSVC
-        if(";i386;i686;amd64;x86_64;" MATCHES ";${CMAKE_HOST_SYSTEM_PROCESSOR};")
+        if(";i386;i686;amd64;x86_64;arm64;" MATCHES ";${CMAKE_HOST_SYSTEM_PROCESSOR};")
```
joshua-stone, line 228: Adding `aarch64` to the list of architectures nullifies the patch that was put in place for addressing https://github.com/moneymanagerex/moneymanagerex/issues/5953

The idea is that passing `-msse4.2 -maes` will cause aarch64 builds on Linux to fail. That's why the flatpak build patches CMakeLists to get multi-arch support:

https://github.com/flathub/org.moneymanagerex.MMEX/blob/e1f3b4f4f61a5360ed5d50f4b043a4b7a6683a13/cmakelists-3rd.patch

joshua-stone: A better solution may be to try to add `-msse4.2 -maes` *unless* it detects that it's building for aarch64 on Linux.

joshua-stone:
This change should add fix the regression:

https://github.com/moneymanagerex/moneymanagerex/pull/6024

Issue #6040, opened 2023-06-22
Refresh Stock Prices from Yahoo error "url using bad/illegal format or missing url"


therealstretch:
MMEX version:
 - [x] 1.6.4

Operating System:
 - [x] Windows 11

# Description of the bug

When I click "Refresh Stock Prices from Yahoo"  in Stock Portfolio view
I get the error "url using bad/illegal format or missing url"

# Reproduction

Is the bug reproducible? 
 - [x] Always 

Reproduction steps:
click "Refresh Stock Prices from Yahoo" in Stock Portfolio view

Expected result:
Stock prices are updated

Actual result:
above Error, no stock price updated

![yahoo](https://github.com/moneymanagerex/moneymanagerex/assets/131329434/37d6216e-2ad1-48c5-9ac2-075d69f82a1a)

Heny67:
> I am unable to reproduce this issue. Can you please share the stock symbols that you are trying to refresh?

Hi n-stein!

It works the same way for me.
Ubuntu mate 22.04 + https://github.com/moneymanagerex/moneymanagerex/issues/6038

![Névtelen1](https://github.com/moneymanagerex/moneymanagerex/assets/80578702/57e881db-2886-457d-9523-4996770a01e0)

n-stein:
@Heny67 need more information as everything is working for me.

Can you provide:
* What symbols you are refreshing
* The YAHOO_FINANCE_COOKIE & YAHOO_FINANCE_CRUMB values from your mmexini.db3 file

Heny67:
2b76.de
d5bk.de
iuse.l
iwda.as
vuaa.mi
xdwh.de

n-stein:
With an SQL browser you can view your mmexini.db3 settings database. There are two keys which store the Yahoo Finance Cookie & Crumb values:

![image](https://github.com/moneymanagerex/moneymanagerex/assets/16599929/db81daed-c19d-4cf3-b1d7-49e1aa702dbb)

Heny67:
![kép](https://github.com/moneymanagerex/moneymanagerex/assets/80578702/448e3f85-aa1d-4758-b43a-5255adc7fab5)

Galdor83:
Same error here and in the mmexini.db3 file I've no entry for YAHOO settings (171 total rows)

n-stein:
Don't know what I was thinking.. of course the YAHOO settings will be blank -- they are only stored if the fetch is successful (which clearly isn't the case here).

[...]

PR #6052, opened 2023-06-26
fix(#6040): fix stock refresh for GDPR countries

n-stein:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6052)


n-stein:
Hang on... there's a memory leak somewhere that needs to be plugged.

Issue #1349, opened 2017-10-17
Consolidate README.md and README.TXT


slodki:
[README.md](/moneymanagerex/moneymanagerex/blob/master/README.md) and [README.TXT](/moneymanagerex/moneymanagerex/blob/master/README.TXT) are very similar but different and should be consolidated into one file.

ovari:
Where is [README.TXT](https://github.com/moneymanagerex/moneymanagerex/blob/master/README.TXT) used?

Can README.TXT be deleted?

vomikan:
> Where is [README.TXT](https://github.com/moneymanagerex/moneymanagerex/blob/master/README.TXT) used?
> 
> Can README.TXT be deleted?

I was sure that this file is used in the About dialog. But in place of the application description, there is information about the components.

At the moment there is only a link to the file, but it is not used.

vomikan:
let's think again. This file is located in the archive of the portable build, as well as after installation.
This is a good practice. Is it really worth deleting this file? It is static and you can not change it for 100 years.

![image](https://github.com/moneymanagerex/moneymanagerex/assets/6836805/0fd4e7dc-f530-4187-9c2d-03a62bfc6c70)

tactilis:
The Supported Platforms information in README.TXT is significantly out of date compared to the up to date information here  https://moneymanagerex.org/docs/downloads/compatibility/

For example on Windows: 95, 98, ME, 2000, NT, XP are not supported platforms. Not that anyone is likely to be running these now.

vomikan:
> Windows: 95, 98, ME, 2000, NT, XP are not supported platforms

Money Manager Ex is a free, open-source, cross-platform, easy-to-use personal 
finance software. It primarily helps organize one's finances and keeps track 
of where, when and how the money goes.

MMEX includes all the basic features that 90% of users would want to see in a
personal finance application. The design goals are to concentrate on simplicity
and user-friendliness - something one can use everyday.


Let's leave only the description.

Issue #1028, opened 2017-01-18
-==New Features collection==-


vomikan:
In case somebody want to develop one of this feature, please take and reopen one of them.
#302 - Refactor home page (phase 2)
#303 - Crash report collection
#359 - Custom autocomplete ComboBox
#386 - OFX Import
#396 - automatic assignment
#528 - Pagination support for all item list
#628 - New feature: insert new operations directly
#637 - Remove Historic Items without disturbing Current Balances
#707 - Pre-loaded data translation
#735 - Backup location
#842 - User agent refine
#847 - Expanded Assets section on Home screen, similar to Stocks
#893 - Make secure connection for web server
#910 - Notes in split categories
#925 - Compound interest in assets
#926 - Auto update current market values for goods in assets
#952 - Migrate/import from Quicken

J316:
In case somebody want to develop one of this feature, please take and reopen one of them.
#1025 - Auto linking of transactions cntinuation of #560 
#628 - New feature: insert new operations directly
#302 - Refactor home page (phase 2)
#303 - Crash report collection
#735 - Backup location
#626 - New feature: be able to remove unused reports
#791 - Feature Request - Charts for monthly budget performance
#952 - Migrate/import from Quicken
#925 - Compound interest in assets
#910 - Notes in split categories
#847 - Expanded Assets section on Home screen, similar to Stocks
#742 - "Budget Performance Summary" - Enhancement Needed
#926 - Auto update current market values for goods in assets
#842 - User agent refine
#893 - Make secure connection for web server
#637 - Remove Historic Items without disturbing Current Balances
#713 - YQL
#396 - automatic assignment
#386 - OFX Import
#359 - Custom autocomplete ComboBox
#629 - New feature: trend/evolution of the account should be a standard report
#707 - Pre-loaded data translation
#528 - Pagination support for all item list
#478 - User interface scaling too tiny on high DPI devices

FTFY. I can't see any reason for closing everything though. Anyway you can update your post and delete mine.

vomikan:
@J316  thank you for update. I'll update case nàmes.

siena123:
I have already started working on some changes to reports in general that include some of the report items noted above. I will work on the others as well while I am doing them. I will be submitting just small framework changes to start with working toward the end goal of overall improvement of the reports.

#626 - New feature: be able to remove unused reports
#791 - Feature Request - Charts for monthly budget performance
#742 - "Budget Performance Summary" - Enhancement Needed
[...]

PR #6134, opened 2023-07-22
Update README.TXT

ovari:
https://github.com/moneymanagerex/moneymanagerex/issues/1349

https://github.com/moneymanagerex/moneymanagerex/issues/1028#issuecomment-1646639533

@tactilis is this what you had in mind?

Thank you

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6134)


tactilis:
> @tactilis is this what you had in mind?

Sorry, I have no idea.

I am just working my way through all the old issues that are still open with a view to recommending to @vomikan whether he should close them or not.

I was unsure about #1349 as I don't know where README.TXT is used. Maybe it can be deleted but it needs investigating.

Issue #1983, opened 2019-02-03
Change name of Recurring Transactions to Transaction Planner


slodki:
There are different names in MMEX for BILLSDEPOSITS table functionality:
* main menu, nav tree and right panel title &rarr; *Recurring Transactions*
* tooltips and right mouse button context menu in right panel &rarr; *Bills & Deposits Series*

Both existing names are misleading:
* not only recurring transactions can be stored in BILLSDEPOSITS but one time events too
* transactions not related to bills payments or deposits can be managed there like transfers etc.

Transaction Planner name is clearly describing the purpose of BILLSDEPOSITS table: place for storing planned, not confirmed transactions to be entered into ledger in the future.

vomikan:
Please after 1.4 release

vomikan:
@ovari could you please sort this out?

ovari:
> could you please sort this out?

1. I don't have a problem with `Recurring Transactions`.
2. Quicken uses `Scheduled Transaction List`.
3. `Transaction Planner` seems fine too.
4. What do other financial software programs use?

Can a vote with reasons be taken somewhere online before this is changed?

tactilis:
> I don't have a problem with `Recurring Transactions`

I have no problem with that name.

and this:

> tooltips and right mouse button context menu in right panel → _Bills & Deposits Series_

is no longer the case. They say Recurring Transactions now.

ovari:
`Scheduled Transactions` may be better as there may be a one-time payment/deposit which is not technically a recurring transaction.

ovari:
Seems like there are 26 strings to update https://crowdin.com/translate/moneymanagerex/470/en-hu?filter=basic&value=0#q=recur

ovari:
https://github.com/moneymanagerex/moneymanagerex/pull/6156

ovari:
@vomikan can this issue please be closed? Thank you

PR #6156, opened 2023-08-04
fix(#1983): rename `Recurring` to `Scheduled`

ovari:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6156)


src/model/Model_Attachment.cpp
n-stein, line 1: Please revert these. The Model_Attachment strings can't be changed without an accompanying database upgrade or else all attachments and tags will be disconnected from scheduled transactions.

src/billsdepositsdialog.h
n-stein, line 1: Since you've changed the "name" string here, you must also update it here https://github.com/moneymanagerex/moneymanagerex/blob/1af83dc97b5713c79e8cb05683d43ced27cc3e03/src/util.cpp#L1855

n-stein (changes requested):

src/billsdepositsdialog.h
ovari: https://github.com/moneymanagerex/moneymanagerex/pull/6156/commits/93be80dd14df7cd1bd7ec2e76ac3da5fedb59587

src/model/Model_Attachment.cpp
ovari: Done. Thank you

ovari:
Thank you @n-stein for your comments. All suggestions should be done. If not, please advise. Thank you once again.

src/model/Model_Attachment.cpp
```diff
@@ -25,10 +25,10 @@ const std::vector<std::pair<Model_Attachment::REFTYPE, wxString> > Model_Attachm
     {Model_Attachment::STOCK, wxTRANSLATE("Stock")},
     {Model_Attachment::ASSET, wxTRANSLATE("Asset")},
     {Model_Attachment::BANKACCOUNT, wxTRANSLATE("BankAccount")},
-    {Model_Attachment::BILLSDEPOSIT, wxTRANSLATE("ScheduledTransaction")},
+    {Model_Attachment::BILLSDEPOSIT, wxTRANSLATE("RecurringTransaction")},
     {Model_Attachment::PAYEE, wxTRANSLATE("Payee")},
     {Model_Attachment::TRANSACTIONSPLIT, wxTRANSLATE("TransactionSplit")},
-    {Model_Attachment::BILLSDEPOSITSPLIT, wxTRANSLATE("ScheduledTransactionSplit")}
+    {Model_Attachment::BILLSDEPOSITSPLIT, wxTRANSLATE("RecurringTransactionSplit")}
```
vomikan: I don't like this piece of code.
translators will ask questions why there are no spaces.

there are none, it seems that there are no spaces in the attachments in the file path.
where it's just an algorithm to change spaces to _.

ovari: There has been no change with this patch as it has been reverted.

Issue #5213, opened 2022-10-23
Add `User Interface Language` to `About` dialog


ovari:
Should the `About` dialog contain the system `locale` as it could help with debugging?

```
$ locale
```

https://github.com/moneymanagerex/moneymanagerex/issues/3394#issuecomment-1288182280

What do you think?

Thank you

ovari:
@joshua-stone can this please be added? Thank you

joshua-stone:
How exactly would you like this to be achieved? Embed the equivalent to the following code?

```c++
#include <iostream>
#include <locale>

int main() {
    std::cout << "Locale: " << std::locale("").name() << std::endl;
}
```

```
$ g++ locale.cpp -o locale
$ ./locale 
Locale: en_US.UTF-8
```

joshua-stone:
It's also possible for someone to share locale output from the flatpak environment for debugging purposes:

```
$ flatpak run --command=locale org.moneymanagerex.MMEX 
LANG=en_US.UTF-8
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_PAPER="en_US.UTF-8"
LC_NAME="en_US.UTF-8"
LC_ADDRESS="en_US.UTF-8"
LC_TELEPHONE="en_US.UTF-8"
LC_MEASUREMENT="en_US.UTF-8"
LC_IDENTIFICATION="en_US.UTF-8"
LC_ALL=
```

ovari:
Sorry for the misunderstanding.

Whatever the `User Interface Language` is set in MMEX using `Menu Bar` > `View` > `User Interface Language` > `hu-HU magyar (Magyarország)`

![image](https://github.com/moneymanagerex/moneymanagerex/assets/17465872/6443f87b-017e-4236-918e-92f5470136a6)

At the moment, the MMEX About dialog does not show the `User Interface Language`, i.e. `hu-HU magyar (Magyarország)`.

![image](https://github.com/moneymanagerex/moneymanagerex/assets/17465872/8274ad38-2d3b-4c6a-afd5-bd08b960358c)

When writing bug reports, it may be useful to copy and paste the information in the `About` dialog like so:

```
Money Manager Ex
[...]

PR #6265, opened 2023-10-27
Show system locale and user application language in About dialog

joshua-stone:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6265)

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -43,6 +43,7 @@
 #include <wx/fs_mem.h>
 #include <fmt/core.h>
 #include <cwchar>
+#include <locale>
 
 using namespace rapidjson;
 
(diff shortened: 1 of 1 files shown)


joshua-stone:
@whalley This should help address https://github.com/moneymanagerex/moneymanagerex/issues/5213

image

joshua-stone:
It should have broader desktop environment detection now:

image

n-stein:
Windows has an extra bullet point as well:

mmex_JRZKAJLvr1

whalley:
Same on MacOS
CleanShot 2023-10-30 at 20 05 56@2x

joshua-stone:
@n-stein @n-stein This result is consistent with the way [wxPlatformInfo::Get().GetDesktopEnvironment()](https://docs.wxwidgets.org/latest/classwx_app_traits.html#ad333680437446304014c087d2fffedcf) behaves.

Since it doesn't appear that this never worked outside of Linux/X11, it would stand to reason that this should be moved into an #ifdef block like other OS-specific codepaths.

PR #6374, opened 2023-12-16
Update transactions.cpp

ovari:

https://crowdin.com/translate/moneymanagerex/470/en-pl?filter=basic&value=0#798745

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6374)


whalley:
I would question whether this is needed. The concise use of "Rate" would seem sufficient to me without extending to a log title.

Issue #6481, opened 2024-01-24
`MMB Files (*.mmb)`: 1. Add a space ` `; and, 2. allow translation


ovari:
# Description of the bug

1. Menu Bar, File, New Database.
2. Please add a space ` ` to change `MMB Files(*.mmb)` to `MMB Files (*.mmb)`.
3. Please allow `MMB Files (*.mmb)` to be translatable.

Thank you

![image](https://github.com/moneymanagerex/moneymanagerex/assets/17465872/5c959468-2bf2-4b91-8d04-443f5c39faa4)

```

Money Manager Ex
• Version: 1.7.1-Beta.1 64-bit
• Built on: Jan 20 2024 03:16:22
• Database version: 19 (aes128cbc)
• Git commit: dbfb7dacd (2024-01-20)
• Git branch: master
MMEX is using the following support products:
• wxWidgets 3.2.4 (wxGTK 3.24)
• wxSQLite3 4.9.1 (SQLite 3.40.0)
• RapidJSON 1.1.0
• Lua 5.3.5
• libcurl/8.5.0-DEV OpenSSL/3.1.4 zlib/1.3 nghttp2/1.56.0
• gettext 0.22
• apexcharts.js
Built with:
• CMake 3.28.1
• GNU Make 4.4.1
• GCC 13.2.0
Running on:
• GNOME.45 (Flatpak runtime)
• GNOME 45 (Flatpak runtime) ""
• Linux 5.15.0-91-generic x86_64
• X-Cinnamon (X11)
• System Locale: en_AU.UTF-8
• User Interface Language: en_US.UTF-8
• 1368x768 24-bit 96x96ppi
```

whalley:
@ovari If we are making these changes perhaps better wording would be.... "MMEX Database" rather than "MMB Database" it seems clearer.

n-stein:
Also, this text needs to be wrapped in `_( )` if it should be available for translation.

https://github.com/moneymanagerex/moneymanagerex/blob/98fcf343f0ae12fbb5fbda71e5e4c3d6ff34e39f/src/mmframe.cpp#L2296

rmelillo76:
Do you guys think this is a risky string to allow translation? Allowing it to be controlled externally could break the open file dialog if the placement of the pipe and wildcards are not done 100% perfect. If the goal is to allow "database "and "file" to be translated, maybe only a substring should be wrapped.

whalley:
[...]

PR #6486, opened 2024-01-26
fix(#6481) feedback

ovari:

@whalley https://github.com/moneymanagerex/moneymanagerex/issues/6481#issuecomment-1908327658

@n-stein https://github.com/moneymanagerex/moneymanagerex/issues/6481#issuecomment-1908408876

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6486)

### DIFF
--- a/src/appstartdialog.cpp
+++ b/src/appstartdialog.cpp
@@ -106,7 +106,7 @@ void mmAppStartDialog::CreateControls()
     itemBoxSizer5->Add(itemButton6, 0, wxGROW | wxALL, 5);
 
     wxButton* itemButton7 = new wxButton(this, wxID_OPEN, _("Open &Existing Database"));
-    mmToolTip(itemButton7, wxString::Format(_("Open %s database (*.mmb)"),
+    mmToolTip(itemButton7, wxString::Format(_("Open %s database (*.mmb)")
         , mmex::getProgramName()));
     itemBoxSizer5->Add(itemButton7, 0, wxGROW | wxALL, 5);
 
--- a/src/appstartdialog.cpp
+++ b/src/appstartdialog.cpp
@@ -106,7 +106,8 @@ void mmAppStartDialog::CreateControls()
     itemBoxSizer5->Add(itemButton6, 0, wxGROW | wxALL, 5);
 
     wxButton* itemButton7 = new wxButton(this, wxID_OPEN, _("Open &Existing Database"));
-    mmToolTip(itemButton7, _("Open MMEX database (*.mmb)"));
+    mmToolTip(itemButton7, wxString::Format(_("Open %s database (*.mmb)"),
+        , mmex::getProgramName()));
     itemBoxSizer5->Add(itemButton7, 0, wxGROW | wxALL, 5);
 
     wxButton* itemButton8 = new wxButton(this, wxID_SETUP , _("User Interface &Language"));
--- a/src/appstartdialog.cpp
+++ b/src/appstartdialog.cpp
@@ -106,7 +106,7 @@ void mmAppStartDialog::CreateControls()
     itemBoxSizer5->Add(itemButton6, 0, wxGROW | wxALL, 5);
 
     wxButton* itemButton7 = new wxButton(this, wxID_OPEN, _("Open &Existing Database"));
-    mmToolTip(itemButton7, _("Open an already created database file with extension (*.mmb)"));
+    mmToolTip(itemButton7, _("Open MMEX database (*.mmb)"));
     itemBoxSizer5->Add(itemButton7, 0, wxGROW | wxALL, 5);
 
     wxButton* itemButton8 = new wxButton(this, wxID_SETUP , _("User Interface &Language"));
--- a/src/constants.cpp
+++ b/src/constants.cpp
@@ -102,7 +102,7 @@ const wxString mmex::DEFDELIMTER = ",";
 
 const wxString mmex::getProgramName()
 {
-    return wxString("Money Manager Ex");
+    return wxString("MMEX");
 }
 const wxString mmex::getTitleProgramVersion()
 {
--- a/src/import_export/univcsvdialog.cpp
+++ b/src/import_export/univcsvdialog.cpp
@@ -2214,7 +2214,7 @@ void mmUnivCSVDialog::OnFileBrowse(wxCommandEvent& WXUNUSED(event))
     long flags = IsImporter() ? wxFD_FILE_MUST_EXIST | wxFD_OPEN : wxFD_SAVE;
     const wxString defaultWildcard = IsXML()
         ? wxString() << _("XML Files (*.xml)") << "|*.xml;*.XML|" << _("All Files") << "|" << wxFileSelectorDefaultWildcardStr
-        : wxString() << _("CSV Files (*.csv)") << "|*.csv;*.CSV";
+        : wxString() << _("Comma-separated values (*.csv)") << "|*.csv;*.CSV";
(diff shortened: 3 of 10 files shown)


ovari:
@rmelillo76 does this patch satisfy your https://github.com/moneymanagerex/moneymanagerex/issues/6481#issuecomment-1908486239 comment?

rmelillo76:
> @rmelillo76 does this patch satisfy your https://github.com/moneymanagerex/moneymanagerex/issues/6481#issuecomment-1908486239 comment?

Hi @ovari yes I believe so. Thank you for those tweaks

src/appstartdialog.cpp
ovari, line 1: Could someone please advise how to change
`_("Open MMEX database (*.mmb)")`
to
`_("Open "+mmex::getProgramName()+" database (*.mmb)")`
?

ovari: Attempt 1:
```
wxString::Format(_("Open %s database (*.mmb)"),
        , mmex::getProgramName())
```

src/constants.cpp
whalley, line 1: Not sure we really want to change the name here. "Money Manager Ex" is the formal name. "MMEX" is a shortend version. Would be better, perhaps, to add a getProgramShortName().

src/mmchecking_list.cpp
```diff
@@ -649,28 +649,28 @@ void TransactionListCtrl::OnMouseRightClick(wxMouseEvent& event)
         : wxPLURAL("&Permanently delete selected transaction...", "&Permanently delete selected transactions...", selected));
     if (is_nothing_selected) subGlobalOpMenuDelete->Enable(MENU_TREEPOPUP_DELETE2, false);
     subGlobalOpMenuDelete->AppendSeparator();
-    subGlobalOpMenuDelete->Append(MENU_TREEPOPUP_DELETE_VIEWED, !m_cp->isTrash_ ? _("Delete all transactions in current view...") : _("Permanently delete all transactions in current view..."));
+    subGlobalOpMenuDelete->Append(MENU_TREEPOPUP_DELETE_VIEWED, !m_cp->isTrash_ ? _("Delete &all transactions in current view...") : _("Permanently delete &all transactions in current view..."));
     if (!m_cp->isTrash_) {
-        subGlobalOpMenuDelete->Append(MENU_TREEPOPUP_DELETE_FLAGGED, _("Delete Viewed \"Follow Up\" Transactions..."));
-        subGlobalOpMenuDelete->Append(MENU_TREEPOPUP_DELETE_UNRECONCILED, _("Delete Viewed \"Unreconciled\" Transactions..."));
+        subGlobalOpMenuDelete->Append(MENU_TREEPOPUP_DELETE_FLAGGED, _("Delete Viewed “&Follow Up” Transactions..."));
```
whalley, line 654: Miust admit, I don't like using the alternatoive double quotes here and prefer escaping the 'normal' quotes.

ovari: Is there an operating system which breaks using these quotation marks?

Mozilla standard is to use these quotes now, which an example shown below:
https://transvision.mozfr.org/?recherche=Send+websites+a+%E2%80%9CDo+Not+Track%E2%80%9D+signal+that+you+don%E2%80%99t+want+to+be+tracked&repo=all_projects&sourcelocale=en-US&locale=en-US&search_type=strings

It is understood that Apple, Google and Microsoft uses these new quotes too.

Furthermore, they also use the 1-character ellipsis `…`.

n-stein: If you use double quotation marks you MUST leave the escape character `\` to indicate to the compiler that the inner marks do not terminate the string. It will not compile as it is currently written.

`_("Delete Viewed \"Follow Up\" Transactions...")`

Using single quotes doesn't require any escape character:

`_("Delete Viewed 'Follow Up' Transactions...")`

src/mmchecking_list.cpp
```diff
@@ -649,28 +649,28 @@ void TransactionListCtrl::OnMouseRightClick(wxMouseEvent& event)
...
+        subGlobalOpMenuDelete->Append(MENU_TREEPOPUP_DELETE_UNRECONCILED, _("Delete Viewed “&Unreconciled” Transactions..."));
```
n-stein, line 655: Same as above, you can't remove the `\` escape character prior to the inner double quotation marks.

src/constants.cpp
n-stein: I agree with @whalley on this one. Leave the getProgramName alone and we can add a new getProgramShortName.

Although I'm not sure we even need to replace all of the instances of "MMEX" with a `Format("%s", getProgramShortName())`. Why can't we just leave "MMEX" in the string? Is this just to prevent users from changing the name during translation?

src/mmTips.h
```diff
@@ -19,51 +19,61 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
...
+    wxTRANSLATE("Software tip: Replace closed-source software with free/libre open-source software. Visit https://alternativeto.net/"),
+    wxTRANSLATE("Software tip: Share MMEX Money Manager with your friends. Share https://moneymanagerex.org/"),
+    wxTRANSLATE("Software tip: Share, freely and privately with “Jami” a free/libre, end-to-end encrypted, and private communication software. Visit https://jami.net/"),
+    wxTRANSLATE("Software tip: Take handwritten notes with ease using “Xournal++”. Visit https://xournalpp.github.io/"),
+    wxTRANSLATE("Software tip: Use “Qalculate!” the ultimate desktop calculator – a powerful and versatile, yet simple to use, calculator with useful tools for everyday needs (such as currency conversion and percent calculation). Visit https://qalculate.github.io/"),
+    wxTRANSLATE("Software tip: Watch free and live TV, movies and series from around the world with “Hypnotix”. Visit https://github.com/linuxmint/hypnotix"),
+    wxTRANSLATE("Software tip: Watch YouTube more conveniently with “FreeTube”. Visit https://freetubeapp.io/"),
```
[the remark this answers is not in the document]
renato-mmex: @whalley @n-stein @vomikan 
I have great difficulty translating all the text changes / tips made by @ovari   There are a lot of pointless texts that I can't expect German-speaking users to read, so I leave such texts blank. In addition, we now have the problem that the translators of the different languages have to do a lot of unnecessary translation work.
**I request that the text changes be reversed.**

ovari:
Thank you for your feedback. Will send new patch.

src/import_export/univcsvdialog.cpp
```diff
@@ -2213,8 +2213,8 @@ void mmUnivCSVDialog::OnFileBrowse(wxCommandEvent& WXUNUSED(event))
 
     long flags = IsImporter() ? wxFD_FILE_MUST_EXIST | wxFD_OPEN : wxFD_SAVE;
     const wxString defaultWildcard = IsXML()
-        ? wxString() << _("XML Files (*.xml)") << "|*.xml;*.XML|" << _("All Files") << "|" << wxFileSelectorDefaultWildcardStr
-        : wxString() << _("CSV Files (*.csv)") << "|*.csv;*.CSV";
+        ? wxString() << _("Extensible Markup Language (*.xml)") << "|*.xml;*.XML|" << _("All Files") << "|" << wxFileSelectorDefaultWildcardStr
```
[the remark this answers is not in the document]
ovari: [LibreOffice](https://www.libreoffice.org/) expands the strings as shown in the screenshot below.

LibreOffice

@n-stein would "PDF - Portable Document Format (*.pdf)" be better?

Thank you

PR #6532, opened 2024-02-25
switch default cipher to SQLCipher4(AES256) from AES128

guanlisheng:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6532)

### DIFF
--- a/src/dbwrapper.cpp
+++ b/src/dbwrapper.cpp
@@ -42,7 +42,9 @@ wxSharedPtr<wxSQLite3Database> mmDBWrapper::Open(const wxString &dbpath, const w
 
     int err = SQLITE_OK;
     wxString errStr=wxEmptyString;
-    wxSQLite3CipherAes128 cipher;
+    wxSQLite3CipherSQLCipher cipher;
+    cipher.InitializeVersionDefault(4);
+    cipher.SetLegacy(true);
     try
     {
         if (debug)
--- a/src/dbwrapper.cpp
+++ b/src/dbwrapper.cpp
@@ -42,13 +42,14 @@ wxSharedPtr<wxSQLite3Database> mmDBWrapper::Open(const wxString &dbpath, const w
 
     int err = SQLITE_OK;
     wxString errStr=wxEmptyString;
+    wxSQLite3CipherAes128 cipher;
     try
     {
         if (debug)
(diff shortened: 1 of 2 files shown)


whalley:
Plan is get v1.7.1 out in next few weeks so we should hold of with any non essential bugs fixes until then please.

guanlisheng:
this PR changes nothing... as the subject, it is to set the default cipher explicitly.

src/dbwrapper.cpp
```diff
@@ -42,13 +42,16 @@ wxSharedPtr<wxSQLite3Database> mmDBWrapper::Open(const wxString &dbpath, const w
 
     int err = SQLITE_OK;
     wxString errStr=wxEmptyString;
+    wxSQLite3CipherSQLCipher cipher;
+    cipher.InitializeVersionDefault(4);
```
n-stein, line 46: What is the reason for using version 4 here? The API documentation for wxSQLite3Cipher suggests 3 is the max.

https://utelle.github.io/wxsqlite3/docs/html/classwx_s_q_lite3_cipher_s_q_l_cipher.html#a8df5f826a6582c238ff83de491a852a2

> Currently version 1, 2 and 3 of SQLCipher are 
supported. If the version number is invalid, version 3 is assumed.

n-stein: It seems the documentation is out of date. Looking at the code I see 4 is now the default:

https://github.com/utelle/wxsqlite3/blob/1e387a4bea65ada2fdb1b7c95621dc84afb48cb7/src/wxsqlite3.cpp#L6257C1-L6269C13

guanlisheng:
@vomikan @whalley @renato-mmex and @n-stein , can you help to download binary and perform some basic test?

n-stein:
Seems to work OK. I encrypted my database in the current 1.8.0 build, then built this branch and it opened successfully. Tried saving, opening, removing password and had no issues.

Only thing I noticed (not directly related to these changes) is that this "Error" dialog should just be a simple wxMessageDialog. It makes it look like the conversion failed when removing the encryption.

https://github.com/moneymanagerex/moneymanagerex/blob/e046d191d00bf1f62537981cdac59f7f8d21204b/src/mmframe.cpp#L2375

image

Issue #6503, opened 2024-02-01
Move transactions


xdodo49:
# Description of the feature

Move same-dated transactions up or down to maintain chronological order

# Use cases

If not obvious, explain the use cases or problems to solve.

n-stein:
Seems like a reasonable request. I know we have had some complaints about the Date+ID sort order lock, so this could help.

Have you thought about the mechanics of how it would work? Maybe Ctrl + Up/Down arrow to move the transaction? Do we allow a transaction to be moved any time, or only when sorted by date?

whalley:
Would this not be better acheived by implementing this:  https://github.com/moneymanagerex/moneymanagerex/issues/2729. the use can then better order chronologically.

n-stein:
> Would this not be better acheived by implementing this: #2729. the use can then better order chronologically.

Good point. Two birds, one stone. Let me work on #2729.

DevelopmentAnything:
#5873 제가 원하던 기능입니다.
MMEX v2에서는 거래 날짜 필드 속성을 datetime로 만들면 좋겠습니다.
😍😍😍

n-stein:
> 제가 원하던 기능입니다. 
> MMEX v2에서는 거래 날짜 필드 속성을 datetime로 만들면 좋겠습니다.

> Translation:
> This is a feature I've been wanting.
> In MMEX v2, it would be nice to make the transaction date field attribute a datetime.

This is completed for 1.7.1. Just go into Tools -> Options -> View Options and enable `Use 'Time' in transaction recording/reporting`

![image](https://github.com/moneymanagerex/moneymanagerex/assets/16599929/dbe1b72c-28fc-4bd9-abba-a57e76eafd4d)

![image](https://github.com/moneymanagerex/moneymanagerex/assets/16599929/78e3db6e-399f-4ba0-9555-3631e1ee08d4)

![image](https://github.com/moneymanagerex/moneymanagerex/assets/16599929/f63a4d7b-a978-4b6b-b30d-ec087433c0b8)


@xdodo49 This allows you to reorder the same-dated transactions by changing the times.

DevelopmentAnything:
[...]

PR #6555, opened 2024-03-05
fix(#6503): parse date time

vomikan:
Stage 1

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6555)

### DIFF
--- a/src/filtertransdialog.cpp
+++ b/src/filtertransdialog.cpp
@@ -1423,20 +1423,10 @@ template <class MODEL, class DATA> bool mmFilterTransactionsDialog::mmIsRecordMa
 {
     bool ok = true;
 
-    wxString refType;
-    // Check the Data type to determine the tag RefType
-    if (typeid(tran).hash_code() == typeid(Model_Checking::Data).hash_code())
-        refType = Model_Attachment::reftype_desc(Model_Attachment::TRANSACTION);
-    else if (typeid(tran).hash_code() == typeid(Model_Billsdeposits::Data).hash_code())
-        refType = Model_Attachment::reftype_desc(Model_Attachment::BILLSDEPOSIT);
-
     // wxLogDebug("Check date? %i trx date:%s %s %s", getDateRangeCheckBox(), tran.TRANSDATE, getFromDateCtrl().GetDateOnly().FormatISODate(),
-    wxDate date;
-    date.ParseDateTime(tran.TRANSDATE) || date.ParseDate(tran.TRANSDATE);
-    wxString strDate = date.FormatISOCombined();
     if (mmIsAccountChecked() && m_selected_accounts_id.Index(tran.ACCOUNTID) == wxNOT_FOUND && m_selected_accounts_id.Index(tran.TOACCOUNTID) == wxNOT_FOUND)
         ok = false;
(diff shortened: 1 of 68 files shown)


src/filtertransdialog.cpp
```diff
@@ -1423,20 +1423,10 @@ template <class MODEL, class DATA> bool mmFilterTransactionsDialog::mmIsRecordMa
...
-        refType = Model_Attachment::reftype_desc(Model_Attachment::BILLSDEPOSIT);
-
     // wxLogDebug("Check date? %i trx date:%s %s %s", getDateRangeCheckBox(), tran.TRANSDATE, getFromDateCtrl().GetDateOnly().FormatISODate(),
-    wxDate date;
-    date.ParseDateTime(tran.TRANSDATE) || date.ParseDate(tran.TRANSDATE);
-    wxString strDate = date.FormatISOCombined();
     if (mmIsAccountChecked() && m_selected_accounts_id.Index(tran.ACCOUNTID) == wxNOT_FOUND && m_selected_accounts_id.Index(tran.TOACCOUNTID) == wxNOT_FOUND)
         ok = false;
-    else if ((mmIsDateRangeChecked() || mmIsRangeChecked()) && (strDate < m_begin_date || strDate > m_end_date))
+    else if ((mmIsDateRangeChecked() || mmIsRangeChecked()) && (tran.TRANSDATE < m_begin_date || tran.TRANSDATE > m_end_date))
```
n-stein, line 1429: We can't compare straight TRANSDATE here because 'YYYY-mm-dd' < 'YYYY-mm-ddT00:00:00', so all transactions without a time are always filtered out even if they fit the date range. We need to format TRANSDATE as ISOCombined prior to checking against the specified filter values.

vomikan: > 'YYYY-mm-dd' < 'YYYY-mm-ddT00:00:00'

image

Yes, it's just that dates without specifying the time will always be at the top of the list.

vomikan: https://github.com/moneymanagerex/moneymanagerex/pull/6555/commits/7482dd1e22a2d0a4cacf4108855dcf036a754489

Here is the solution: if the time is not set, then we assume that it is any time of the day and not 00:00:00

n-stein: That works, though I feel it limits the usefulness of the filter -- there is no way to isolate transactions with a specific time, the report shows everything from that day with no time as well. You can't filter to show transactions on 2024-03-05 with a time between 13:00 and 16:00.

Personally I prefer treating no time the same as 00:00:00. If the user is modifying the time field on the date filter then they are not looking for all transactions on that date, otherwise they would leave the default range 00:00:00 - 23:59:59.

Maybe others want to weigh in here.

whalley: > Personally I prefer treating no time the same as 00:00:00. If the user is modifying the time field on the date filter then they are not looking for all transactions on that date, otherwise they would leave the default range 00:00:00 - 23:59:59.

I would concur. In fact I suspect that most of the users (me included) don't even need to record the transaction time.

whalley:
What is the current state of this PR?

vomikan:
> What is the current state of this PR?

Take the liberty to accept these changes. Tuning may still be required after more thorough testing of the application.

PR #6575, opened 2024-03-10
Some cleanups and C++ modernisations

gerstrong:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6575)


src/billsdepositsdialog.cpp
```diff
@@ -1348,14 +1348,18 @@ void mmBDDialog::OnsetPrevOrNextRepeatDate(wxCommandEvent& event)
     switch (repeatType)
     {
     case INXDAYS:
+        [[fallthrough]];
```
vomikan, line 1351: for me VS 2022
error C2429: attribute 'fallthrough' requires compiler flag '/std:c++17'

src/dbupgrade.cpp
```diff
@@ -157,7 +157,7 @@ void dbUpgrade::BackupDB(const wxString& FileName, int BackupType, int FilesToKe
     if (BackupType != BACKUPTYPE::VERSION_UPGRADE)
     {
         wxSortedArrayString backupFileArray;
-        const auto fileSearch = wxString::Format("%s%s????-??-??.bak", FileName, BackupName[BackupType]);
+        const auto fileSearch = wxString::Format("%s%s\?\?\?\?-\?\?-\?\?.bak", FileName, BackupName[BackupType]);
```
vomikan, line 160:
```R"(%s%s????-??-??.bak)"

src/billsdepositsdialog.cpp
gerstrong: You are right and it needs to be C++11 compatible.

Issue #6596, opened 2024-03-20
Translated Help (index) pages link to English sub-pages (grm, budget, stocks_and_shares) even when translations are available


DevelopmentAnything:
MMEX version:
 - [ ] 1.6.4
 - [ ] 1.7.0
 - [x] 1.7.1 RC1
 - [ ] 1.7.2
 - [ ] Other (please specify)

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [x] Windows
 - [ ] Mac OSX
 - [ ] Linux 

# Description of the bug

Please describe your issue with details.
Add screenshot or other files if needed.

![image](https://github.com/moneymanagerex/moneymanagerex/assets/110871727/80a6eaac-b5c8-4d90-bc68-74547223bf17)

# Reproduction

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:

1. click any language (no English)
![image](https://github.com/moneymanagerex/moneymanagerex/assets/110871727/c9ba5e49-1e7b-4b9c-92d3-e66bd494f3ed)

2. go to 14. Stocks and Shares section
![image](https://github.com/moneymanagerex/moneymanagerex/assets/110871727/674aa412-f45f-4222-b10c-41ae26152c3b)

3. click Further details refer to Stocks, Shares and Share Dividend Management in MMEX 

You can only see the English page

Expected result:

load ..language directory/stocks_and_shares.html

Actual result:

only loaded english page.

# Additional information

If you have a backtrace for a crash or a warning, paste it here.

n-stein:
Same applies to any of the subpages -- Budgeting, Reports, etc.

All the link tags look like `<a href="../budget.html?lang=ru">` but they should have the full locale `<a href="../budget.html?lang=ru_RU">`

DevelopmentAnything:
@n-stein 

The content you explained has already been applied.
Because if don't modify it, it won't even appear on the language list.

What I'm asking is sub page. It's not just Korean but all languages are the same. Only the English page appears.

DevelopmentAnything:
![image](https://github.com/moneymanagerex/moneymanagerex/assets/110871727/0868bdc1-a808-4aa8-9f60-fde2937cdb78)

?lang=ko
has already been applied.
Code does not do parameter parsing.

n-stein:
> The content you explained has already been applied.

[...]

PR #6613, opened 2024-03-26
fix(#6596): correct links in translated help pages

n-stein:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6613)


docs/ru_RU/index.html
```diff
@@ -953,7 +953,7 @@ <h2>Периодические транзакции</h2>
     <h2>Инвестиции</h2>
     MMEX allows you to track Stocks and Shares. The total Stocks are added to
     your total financial worth.<br>Further details refer to
-    <a href="../stocks_and_shares.html?lang=ru">Stocks, Shares and Share Dividend Management in MMEX</a>
+    <a href="../stocks_and_shares.html?lang=en_GB">Stocks, Shares and Share Dividend Management in MMEX</a>
```
ovari, line 956: Unless it never existed, should this be `ru_RU`? Thank you

n-stein: There is no stocks_and_shares.html for the Russian language, so I sent it back to en_GB. However, I think it would be better to just copy the en_GB file to the ru_RU directory then have the links point to files within the directory. The stocks_and_shares file can be translated into Russian later by other users.

n-stein: Done

PR #6654, opened 2024-04-13
Attempt to show captions on Flathub

ovari:

Perhaps having the image before the caption will enable captions to be displayed at https://flathub.org/apps/org.moneymanagerex.MMEX

Image captions show at https://flathub.org/apps/io.github.endless_sky.endless_sky. Is this due to https://github.com/endless-sky/endless-sky/blob/master/io.github.endless_sky.endless_sky.appdata.xml?

@joshua-stone will this work?

Thank you

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6654)

### DIFF
--- a/resources/dist/linux/share/metainfo/org.moneymanagerex.MMEX.metainfo.xml.in
+++ b/resources/dist/linux/share/metainfo/org.moneymanagerex.MMEX.metainfo.xml.in
@@ -288,63 +288,63 @@
   <screenshots>
     
     <screenshot type="default">
+      <image>https://github.com/moneymanagerex/moneymanagerex/assets/17465872/01cc2202-9d5a-4bba-a816-5868512196cb</image>
       <caption>Dashboard</caption>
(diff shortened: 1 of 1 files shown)


joshua-stone:
@ovari I don't think position of the fields matter, but you can try.

Something to consider is that the caption field might not work with additional translations, so you may have to make them English-only. Like so:

```xml
    <screenshot type="default">
      <image>https://github.com/moneymanagerex/moneymanagerex/assets/17465872/01cc2202-9d5a-4bba-a816-5868512196cb</image>
      <caption>Dashboard</caption>
    </screenshot>
```

ovari:
@joshua-stone is it possible to merge this patch? If this doesn't work, then another patch without the additional translations can be merged to check? Thank you

joshua-stone:
I don't have the permissions to merge this.

@vomikan would you merge this?

PR #6660, opened 2024-04-17
Attempt to show captions on Flathub

ovari:


ovari:
Note: <image> MUST come before <caption> tag.

Verified fixed. Thank you

image

Issue #6607, opened 2024-03-25
Change to calendar-based numbering scheme (YY.MM) for releases


ovari:
Should MMEX versioning be changed to a calendar-based numbering scheme (`YY.MM`)? If released in March 2024, then the version would be `24.03`.

Other software which generally have releases with version numbers based on the calendar are Ubuntu, (now) LibreOffice, Quicken, MS Office (desktop).

What do you think?

Thank you

n-stein:
It's hard to remember when versions such as 1.6.4 and 1.7.0 were released, so I often find myself going back to the "Releases" page to check. This would certainly make it much easier.

ovari:
@whalley would calendar-based versioning be possible for the upcoming release? Thank you

ovari:
Suggested updates for the [Milestones](https://github.com/moneymanagerex/moneymanagerex/milestones) labels:

[v1.7.1](https://github.com/moneymanagerex/moneymanagerex/milestone/42) → `v+0.1`
[v1.7.2](https://github.com/moneymanagerex/moneymanagerex/milestone/44) → `v+0.2`
[v1.7.3](https://github.com/moneymanagerex/moneymanagerex/milestone/16) → `v+0.3`
[v2.0](https://github.com/moneymanagerex/moneymanagerex/milestone/4) → `v+1.0`
[v2.1](https://github.com/moneymanagerex/moneymanagerex/milestone/5) → `v+1.1`

`v+x` → latest released `v`ersion `+` future version `x`

Prior to release the milestone label can be updated to `YY.MM`.

What do you think?

Thank you

tosim91:
IMHO , the calendar based versioning, as above post, should be used.

gabriele-v:
Money Manager uses Semantic Versioning.
IMHO considering the releases lifecycle is better than CalVer because we cannot guarantee that each year will have a version and users may be confused because they misses version of a specific year.
It's also unclear to distinguish between major and minor changes, major ones generally implies a DB upgrades so they are recognizable with something like 1.6.X -> 1.7.X but not with CalVer.

ovari:
Thank you for your comment @gabriele-v

> users may be confused because they misses version of a specific year

Not having a MMEX release in a year is fine as the version indicates when the software was released.

Microsoft Office for Windows has not had releases every year with versions 95, 97, 2000, 2003, 2007, 2010, 2013, 2016, 2019, 2021, 2024 (planned).
[...]

PR #6667, opened 2024-04-22
fix(#6607): change to calendar-based numbering scheme (`YY.0M.n`) for releases

ovari:

https://github.com/moneymanagerex/moneymanagerex/issues/6607

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6667)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,4 @@
-Version 9.0.0 RC 2
+Version 18.0.0 RC 2
 ~~~~~~~~~~~~~~~~~~~~
 Released: 2024-04-23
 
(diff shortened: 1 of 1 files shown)


whalley:
The version numbering validation has failed, as you know we currently use semantic version numbering, and IT IS validated. You cannot just change to a new version number format.

ovari:
> The version numbering validation has failed, as you know we currently use semantic version numbering, and IT IS validated. You cannot just change to a new version number format.

Thank you @whalley for your comment.

Updated back to SemVer as commented at https://github.com/moneymanagerex/moneymanagerex/issues/6607#issuecomment-2071655904

Thank you once again

Issue #6733, opened 2024-05-08
Total Net worth line on Dashboard hide when scrolling down


spillaip:
MMEX version:
 - [ ] 1.7.0
 - [X] 1.8.0
 - [ ] 1.8.1
 - [ ] 1.8.2
 - [ ] Other (please specify)

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [x] Windows
 - [ ] Mac OSX
 - [ ] Linux 

# Description of the bug
On the dashboard the summary line with Total Net worth is not floating while scrolling. This is benefial to view the accouns and relate to the summary line of all the accounts.

Please describe your issue with details.
Add screenshot or other files if needed.

# Reproduction

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:

Expected result:

Actual result:

# Additional information

If you have a backtrace for a crash or a warning, paste it here.

ovari:
Net Worth covers data if it is not 1-line and fully scrolled to the top.

![image](https://github.com/moneymanagerex/moneymanagerex/assets/17465872/cd8cc821-0394-4b1c-8de8-b7fb3767d3f6)

![image](https://github.com/moneymanagerex/moneymanagerex/assets/17465872/4212018b-5c0e-45a5-a285-83b0ec2f741b)

Thank you

n-stein:
I've turned off wrapping.

ovari:
![image](https://github.com/moneymanagerex/moneymanagerex/assets/17465872/bf9425f5-7dc4-4224-a946-b4349610673d)

> I've turned off wrapping.

Can wrapping stay and the height of the first row be variable? Thank you

PR #6754, opened 2024-05-17
fix(#6733): float dashboard net worth table

n-stein:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6754)

### DIFF
--- a/resources/home_page.htt
+++ b/resources/home_page.htt
@@ -25,7 +25,7 @@
         <tr valign="top">
         <td colspan=2>
             <!--Grand total-->
-            <table class='table'>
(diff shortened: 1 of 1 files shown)


resources/home_page.htt
```diff
@@ -47,6 +47,7 @@
                 document.getElementById('bal_value').innerHTML=grand_json['BALVALUE'];
             </script>
         </td>
+        <tr height='25px'></tr>
```
ovari, line 50: Does this work correctly if the font size is bigger?

image

Thank you

Issue #6827, opened 2024-08-04
The model of scheduled transactions (billsdeposits) is confusing


georgeef:
MMEX version:
 - [ ] 1.7.0
 - [X] 1.8.0
 - [ ] 1.8.1
 - [ ] 1.8.2
 - [X] 5a67e5d (2024-08-04)

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [ ] Windows
 - [X] Mac OSX
 - [ ] Linux 

# Description

In the following
- `entry` means a scheduled transaction (a record in table `BILLSDEPOSITS_V1`)
- `panel` means the `Scheduled Transactions` panel,
- `edit form` means the `Edit Scheduled Transaction` dialog,
- `REPEAT_NONE entry` means an entry with `REPEAT_TYPE` equal to `REPEAT_NONE`,
- `REPEAT_IN entry` means `REPEAT_TYPE` is `REPEAT_IN_X_*`,
- `REPEAT_EVERY entry` means `REPEAT_TYPE` is `REPEAT_EVERY_X_*`,
- `REPEAT_countdown entry` means any other entry,
- `N` is abbreviation for `NUMOCCURRENCES`. 

A `REPEAT_NONE` entry is executed once (it is equivalent to a `REPEAT_countdown` entry with `N=1`).
The label `None` in the interface and the enum name `REPEAT_NONE` are confusing; they should better be `Once` and `REPEAT_ONCE`, resp.
The panel column `Repetitions` shows `∞` (infinity); it should show `1` (as for `REPEAT_countdown` entries with `N=1`).
The edit form accepts a positive number in `Payments Left`, and this number (`N>0`) also appears in the panel column `Repetitions`,  which doesn't make sense to the user; the field `Payments Left` should be reset to empty (i.e., `N` should be reset to `-1`).

`REPEAT_EVERY` entries with `N>0` are executed infinite times (until canceled), e.g., `REPEAT_EVERY_X_DAYS` with `N=1` is equivalent to `REPEAT_DAILY` with `N=-1`.
The panel column `Repetitions` shows the value of `N`; it should show `∞` (infinity).
The value of `N` in `Repetitions` is redundant anyway; it appears also in column `Frequency`.

`REPEAT_EVERY` entries with empty `Period` (`N=-1`) are more confusing.
[...]

PR #6829, opened 2024-08-06
fix(#6827), part 1: Repetitions, special values of NUMOCCURRENCES

georgeef:
Change column `Repetitions` in scheduled transactions (billsdeposits) panel, as proposed in #6827.

Also change the special codes of `NUMOCCURRENCES` from {`REPEAT_INACTIVE`, `REPEAT_NONE`} to the corresponding numeric values {-1, 0}, so they don't interfere with the codes of repeat frequency.  The enum `REPEAT_TYPE` has not been changed; I just added a comment that the case `REPEAT_INACTIVE` can be removed, since it is not used.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6829)

### DIFF
--- a/src/billsdepositspanel.cpp
+++ b/src/billsdepositspanel.cpp
@@ -481,10 +481,15 @@ wxString mmBillsDepositsPanel::getItem(long item, long column)
     case COL_FREQUENCY:
         return GetFrequency(&bill);
     case COL_REPEATS:
-        if (bill.NUMOCCURRENCES == -1)
-            return L"\x221E";
+    {
+        int numRepeats = GetNumRepeats(&bill);
+        if (numRepeats > 0)
+            return wxString::Format("%i", numRepeats).Trim();
+        else if (numRepeats == -1)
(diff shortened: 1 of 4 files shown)


guanlisheng:
thanks @georgeef, 
i think an ENUM value would be more readable than `-1`.  what do you think?

georgeef:
@guanlisheng,
yes i agree. i added enum `REPEAT_NUM` with cases for -1 (infinity) and 0 (unknown). it is separate from enum `REPEAT_TYPE`.

i checked that `REPEAT_TYPE::REPEAT_INACTIVE` appears only in its definition, but i leave it for now (in case i missed something). i'll remove it in a next PR.

georgeef:
I close this PR, since it has been merged as part of #6836.

georgeef:
I close this PR, since it has been merged as part of #6836.

guanlisheng:
thanks, and looking forward to more

PR #6836, opened 2024-08-17
fix(#6827), part 2: inactive billsdeposits; None -> Once

georgeef:
Change billsdeposits model as proposed in #6827

- Change `None` to `Once` and `REPEAT_NONE` to `REPEAT_ONCE`.
- An empty `Period` in the edit form for type `REPEAT_IN_*` or `REPEAT_EVERY_*`, means `1`.
- After the first execution, `REPEAT_IN_*` entries are converted to `REPEAT_ONCE`.
- All entries created by the edit form (with new/edit/duplicate) are active.

## Backwards compatibility

The old inactive entries (of type `REPEAT_IN_*`, `REPEAT_EVERY_*` and with `NUMOCCURRENCES == -1`) are handled as follows:

- They are not executed automatically.
- If they are executed manually (by pressing the `Enter` button), they are converted to `REPEAT_ONCE`, therefore they are removed after execution. (In MMEX 1.8.0 they can be executed manually, as described in #6827, probably due to a bug.)
- If they are edited, they are converted to `REPEAT_ONCE`. The user can change the type again in the edit form, but the new entry is always active.
- If they are duplicated, they are converted to `REPEAT_ONCE`.
- In the Scheduled Transactions panel, only the old inactive entries have `Repetitions` equal to `-` (horizontal bar). All new/updated entries have `Repetitions` equal to a positive number or infinity. 

## User's perspective

Apart from the old inactive entries (which are converted to active at the first chance), all entries in the Scheduled Transactions are active. If the user wants to have a dummy entry as a template, it can be added as type `Once` (or any other type) with a paid/due date in the far future (so it does not appear in the reports of the following years).

The type `REPEAT_IN_*` can be used in order to schedule two transactions. It can also be used to schedule a sequence of transactions with unpredictable paid/due dates, as follows: the user inserts a scheduled transaction of type `REPEAT_IN_*` with the known date of the first transaction, and with a period  according to the expected next paid/due date. After the first execution and before the second date, the user may edit the entry (which is now of type `REPEAT_ONCE`) and change it again to type `REPEAT_IN_*`, with a period according to the third expected paid/due date. This method is anyway of limited value; the user can simply duplicate a previous transaction in the Transactions panel.

The behavior of `REPEAT_IN_*` entries in MMEX 1.8.0 is not clear. If they are meant to be executed only once, then they are no different than `REPEAT_ONCE` entries (which replaces the misleading term `REPEAT_NONE`). Their manual execution is quite strange, since they move backwards in time.

## Additional changes

In `src/billsdepositsdialog.cpp`, the enum in the beginning of the file has been removed and the enum values have been replaced by the corresponding `Model_Billsdeposits::REPEAT_*` values. The old enum was redundant and error-prone (if e.g., by mistake the order of the symbols becomes inconsistent with the order in `Model_Billsdeposits::REPEAT_*`).

The order of repeat types in the gui (defined in `mmBDDialog::BILLSDEPOSITS_REPEATS`), is the same as the order in `Model_Billsdeposits ::REPEAT_*`, however a different order is also supported.

## Branching

This PR (part 2) is a follow-up based on #6829 (part 1). If needed, I can rebase after part 1 is merged. I think that it is easier to review/merge smaller and self-contained parts separately, instead of one big chunk. I plan to submit one more part for the same issue, based on parts 1 and 2.

## New translation strings

> Please do not forget to update the mmex.pot file and write information about the fixed bug in the prerelease [page](https://github.com/moneymanagerex/moneymanagerex/releases).

There are a few new translation strings (e.g., "Once"), but I don't know how to register them.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6836)

### DIFF
--- a/src/billsdepositsdialog.cpp
+++ b/src/billsdepositsdialog.cpp
@@ -39,33 +39,26 @@
 #include "model/Model_CurrencyHistory.h"
 #include <wx/valnum.h>
 
-enum { NONE, WEEKLY, FORTNIGHTLY, MONTHLY
-    , BIMONTHLY, QUARTERLY, HALFYEARLY
-    , YEARLY, FOURMONTHS, FOURWEEKS
-    , DAILY, INXDAYS, INXMONTHS
-    , EVERYXDAYS, EVERYXMONTHS, MONTHLYLASTDAY, MONTHLYLASTBUSINESSDAY
-};
-
+// the order in gui may be different than the database encoding order.
 const std::vector<std::pair<int, wxString> > mmBDDialog::BILLSDEPOSITS_REPEATS =
 {
-    { NONE, wxTRANSLATE("None")},
-    { WEEKLY, wxTRANSLATE("Weekly")},
-    { FORTNIGHTLY, wxTRANSLATE("Fortnightly")},
-    { MONTHLY, wxTRANSLATE("Monthly")},
-    { BIMONTHLY, wxTRANSLATE("Every 2 Months")},
-    { QUARTERLY, wxTRANSLATE("Quarterly")},
-    { HALFYEARLY, wxTRANSLATE("Half-Yearly")},
-    { YEARLY, wxTRANSLATE("Yearly")},
-    { FOURMONTHS, wxTRANSLATE("Four Months")},
-    { FOURWEEKS, wxTRANSLATE("Four Weeks")},
-    { DAILY, wxTRANSLATE("Daily")},
-    { INXDAYS, wxTRANSLATE("In (n) Days")},
-    { INXMONTHS, wxTRANSLATE("In (n) Months")},
-    { EVERYXDAYS, wxTRANSLATE("Every (n) Days")},
-    { EVERYXMONTHS, wxTRANSLATE("Every (n) Months")},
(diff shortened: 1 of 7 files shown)


guanlisheng:
hi @georgeef , i like the idea of renaming `None` to `Once`, which also consists of Android (https://github.com/moneymanagerex/android-money-manager-ex/blob/master/app/src/main/java/com/money/manager/ex/recurring/transactions/Recurrence.java#L27)

I suggest we make the change pure on the rename to limit the impact, and also making the PR more clear.

georgeef:
hi @guanlisheng,

yes, of course i can re-partition the changes. please let me know which of the changes in #6827 you want to have (or which you don't want to have). if you want all the changes already in PRs, let me know how to re-partition them.

guanlisheng:
All items from #6827 are great, it can be applied to both desktop and android.

I think we can implement them gradually, e.g. in multi-small steps.

for this particular PR, can we just include the renaming change? and separate the refactor (no behavior impact) ones.

guanlisheng (approved):
LGTM

PR #6839, opened 2024-08-21
fix(#6827): encapsulate auto execute values

georgeef:
The auto execution mode of scheduled transactions is multiplexed into the `REPEATS` field. The choices are defined in enum `Model_Billsdeposits::REPEAT_AUTO`.

Changes:
- Multiplex/demultiplex the `REPEATS` field into/from auto execution mode and repeat type, using `BD_REPEATS_MULTIPLEX_BASE`, but without any other assumption about the encoding of auto execution mode.
- Use the symbols in `Model_Billsdeposits::REPEAT_AUTO` instead of raw values for the encoding of auto execution mode.
- Use the symbols in `Model_Billsdeposits::REPEAT_TYPE` instead of raw values for repeat type.
- Ignore inactive scheduled transactions (of type `REPEAT_IN_*`, `REPEAT_EVERY_*` with `NUMOCCURRENCES == -1`) in `src/mmhomepage.cpp` and `src/reports/cashflow.cpp`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6839)

### DIFF
--- a/src/billsdepositsdialog.cpp
+++ b/src/billsdepositsdialog.cpp
@@ -232,39 +232,37 @@ void mmBDDialog::dataToControls()
     field_date.ParseDate(m_bill_data.NEXTOCCURRENCEDATE);
     m_date_due->SetValue(field_date);
 
-    // Have used repeatSel to multiplex auto repeat fields.
-    if (m_bill_data.REPEATS >= 2 * BD_REPEATS_MULTIPLEX_BASE)
-        autoExecuteSilent_ = true;
-    else if (m_bill_data.REPEATS >= BD_REPEATS_MULTIPLEX_BASE)
-        autoExecuteUserAck_ = true;
-    m_bill_data.REPEATS %= BD_REPEATS_MULTIPLEX_BASE;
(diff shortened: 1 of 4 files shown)


guanlisheng (approved):
lgtm. @whalley and @n-stein as FYI

georgeef:
there is a small inefficiency in the demultiplexing of `REPEATS`, because it makes an integer division instead of integer multiplication/shift and comparison in the original code. i think a cleaner interface is better, despite this inefficiency.

as commented in #6827, in a future update of the database schema, it would be good to split `REPEATS` into separate fields for auto mode and repeat type. or if the current multiplexing scheme is kept, it would be good to change `BD_REPEATS_MULTIPLEX_BASE` to a power of 2, such that integer division can be implemented as bit-shift (at the cost of human readability of the multiplexed encoding, but anyway the multiplexing is not good for human readability).

guanlisheng:
Readability cares. I will merge the PR shortly.  
btw, i think we should try the best to keep the schema stable

PR #6841, opened 2024-08-23
fix(#6827): Model_Billsdeposits::AllowTransaction

georgeef:
## Current status

`Model_Account::balance(a)`, which calculates the balance of account `a`, has the following properties:
- Transactions with `STATUS == VOID` or with non-empty `DELETEDTIME` are ignored.
- `TRANSAMOUNT` is added to the balance of `ACCOUNTID` if `TRANSCODE == DEPOSIT`.
- `TRANSAMOUNT` is subtracted from the balance of `ACCOUNTID` if `TRANSCODE == WITHDRAWAL` or `TRANSCODE == TRANSFER`.
- `TOTRANSAMOUNT` is added to the balance of `TOACCOUNTID` if `TRANSCODE == TRANSFER`.

`htmlWidgetAccounts::get_account_stats()`, which calculates and stores the sum of transactions per account, is consistent with `Model_Account::balance()`.

## Problems

`mmTransDialog::ValidateData()`, which checks the limits of new or duplicated transactions, is inconsistent with `Model_Account::balance()`:
- It aborts transactions with `STATUS == VOID`, although these transactions do not affect the account balance.

`Model_Billsdeposits::AllowTransaction()`, which maintains a cache of the balance per account, is highly inconsistent with `Model_Account::balance()`:
- It updates the balance also for transaction with `STATUS == VOID`.
- It adds `TRANSAMOUNT` to the balance of `ACCOUNTID` with the wrong sign if `TRANSCODE == TRANSFER`.
- It does not add `TOTRANSAMOUNT` to the balance of `TOACCOUNTID` if `TRANSCODE == TRANSFER`.

Additional (minor) problems in `Model_Billsdeposits::AllowTransaction()`:
- If an account has both a `MINIMUMBALANCE` and a `CREDITLIMIT` (normally this does not happen), the abort message may wrongly show the `CREDITLIMIT` as the exceeded limit.
- If an account has negative `MINIMUMBALANCE` and `CREDITLIMIT` (normally this does not happen), the abort message does not show the exceeded limit.
- It is not a pure function, as implied by its name. Its main functionality is mostly to update the balance cache; in addition it indicates if an account limit is exceeded.

The balance cache avoids repeated calls to `Model_Account::balance()`, which can be expensive for accounts with too many transactions. However it is quite tedious to make it consistent with `Model_Account::balance`. E.g., `Model_Billsdeposits::AllowTransaction()` is called by `mmGUIFrame::OnAutoRepeatTransactionsTimer()` for a sequence of auto executed transactions, either in manual or in silent mode. In manual mode the user can change the amount before pressing OK. Therefore, the balance cache is updated based on the amount in the original scheduled transaction, but the actual transaction may exceed the account limit (without warning), and a change in the amount invalidates the balance cache.

## Changes

In `mmTransDialog::ValidateData()`:
- Do not abort transactions with `STATUS == VOID`.
- As an optimization, `Model_Account::balance()` is called only for accounts with limit.

In `Model_Billsdeposits::AllowTransaction()`:
- Do not cache the accounts balance (this also makes the function pure).
- As an optimization, `Model_Account::balance()` is called only for accounts with limit.
- The calculation of `new_balance` is fully consistent with `Model_Account::balance()`.

## User's perspective

The overhead of repeated calls to `Model_Account::balance()` in `Model_Billsdeposits::AllowTransaction()` is typically small. Usually only a few scheduled transactions are executed in the same batch. The source code is greatly simplified without the balance cache, and consistency with `Model_Account::balance()` is much more important than a minor optimization.

If nevertheless `Model_Account::balance()` becomes very expensive and it is highly desirable to avoid it, the user can disable the limits in accounts with too many transactions, at the cost of missing warnings on exceeded limits for these accounts.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6841)

### DIFF
--- a/src/billsdepositsdialog.cpp
+++ b/src/billsdepositsdialog.cpp
@@ -908,9 +908,7 @@ void mmBDDialog::OnOk(wxCommandEvent& WXUNUSED(event))
     bill_data.TRANSAMOUNT = m_bill_data.TRANSAMOUNT;
     bill_data.TRANSCODE = m_bill_data.TRANSCODE;
 
-    Model_Billsdeposits::AccountBalance bal;
-
-    if (!Model_Billsdeposits::instance().AllowTransaction(bill_data, bal)) return;
+    if (!Model_Billsdeposits::instance().AllowTransaction(bill_data)) return;
     if (!textAmount_->checkValue(m_bill_data.TRANSAMOUNT)) return;
 
     m_bill_data.TOTRANSAMOUNT = m_bill_data.TRANSAMOUNT;
--- a/src/mmframe.cpp
+++ b/src/mmframe.cpp
@@ -509,14 +509,13 @@ void mmGUIFrame::OnAutoRepeatTransactionsTimer(wxTimerEvent& /*event*/)
     //Auto scheduled transaction
     bool continueExecution = false;
 
-    Model_Billsdeposits::AccountBalance bal;
     Model_Billsdeposits& bills = Model_Billsdeposits::instance();
     for (const auto& q1 : bills.all())
     {
         bills.decode_fields(q1);
         if (bills.autoExecuteManual() && bills.requireExecution())
         {
-            if (bills.allowExecution() && bills.AllowTransaction(q1, bal))
+            if (bills.allowExecution() && bills.AllowTransaction(q1))
             {
                 continueExecution = true;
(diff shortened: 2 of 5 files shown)


guanlisheng (approved):
I like the code,  the idea of simplification,  and the principle of design.

PR #6843, opened 2024-08-25
fix(#6827): mmBDDialog::OnOk

georgeef:
`mmBDDialog::OnOk()` calls `Model_Billsdeposits::.AllowTransaction()` in order to check if the execution of a scheduled transaction would exceed an account limit.

## Problems
- The check is performed before `m_bill_data.TRANSAMOUNT` is updated. If the user changed the amount in the dialog form and then immediately pressed OK (without changing focus in between), the old amount in the dialog form is checked.
- This check makes sense only when the scheduled transaction is executed, i.e., only if the dialog form was created by the Enter button. For all other actions (New, Edit, Duplicate), a scheduled transaction is just created/updated in the `BILLSDEPOSITS_V1` table, but it is not executed, therefore no check is needed. Actually a premature check is misleading, e.g., the scheduled transaction may exceed an account limit if executed now, but not if executed as scheduled.

## Change
- Move the check of account limits just before the schedule transaction is executed. Do not check the account limits in other actions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6843)

### DIFF
--- a/src/billsdepositsdialog.cpp
+++ b/src/billsdepositsdialog.cpp
@@ -903,12 +903,6 @@ void mmBDDialog::OnOk(wxCommandEvent& WXUNUSED(event))
     m_bill_data.ACCOUNTID = cbAccount_->mmGetId();
     Model_Account::Data* acc = Model_Account::instance().get(m_bill_data.ACCOUNTID);
 
-    Model_Billsdeposits::Data bill_data;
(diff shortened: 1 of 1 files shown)


guanlisheng:
+ @whalley , @n-stein and @vomikan , any thoughts?

guanlisheng (approved):

PR #6845, opened 2024-08-27
fix(#6827): cashflow (REPEATS, NUMOCCURRENCES)

georgeef:
This PR fixes the repetition of scheduled transactions in `mmReportCashFlow::getTransactions()`, in accordance with `Model_Billsdeposits::completeBDInSeries()`.

The functions in `cashflow.cpp` have several more problems, e.g.,
- The initial balance in `mmReportCashFlow::getTransactions()` ignores `foreignTransactionAsTransfer` (trasfters to `AS_TRANSFER`), which is inconsistent with `Model_Account::balance()`.
- `mmReportCashFlow::trueAmount()` ignores transfer transactions if both `ACCOUNTID` and `TOACCOUNTID` are in `m_account_id`, probably assuming that the net amount is zero, but this is not always true.
- The scheduled transaction dates in `mmReportCashFlow::getTransactions()` are based on `NEXTOCCURRENCEDATE` instead of `TRANSDATE`.
- Residual scheduled transaction (with date earler than today, but not yet executed), are added into `m_forecastVector`, and therefore they appear as past entries in the cashflow reports; they should be added to the initial balance.
- The currencies conversion in `mmReportCashFlow::getTransactions()` and `mmReportCashFlow::trueAmount()` is inconsistent; all conversions should be based on today's rates.

Should I send more fixes for cashflow in the billsdeposits issue (#6827), or should I create another issue? Also, should I send them before or after the upcoming desktop release?

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6845)

### DIFF
--- a/src/reports/cashflow.cpp
+++ b/src/reports/cashflow.cpp
@@ -77,7 +77,6 @@ void mmReportCashFlow::getTransactions()
     wxDateTime endDate = mmDateRange::getDayEnd(m_today.Add(wxDateSpan::Months(getForwardMonths())));
 
     // Get initial Balance as of today
-
     for (const auto& account : Model_Account::instance().find(
         Model_Account::ACCOUNTTYPE(Model_Account::all_type()[Model_Account::INVESTMENT], NOT_EQUAL)
         , Model_Account::STATUS(Model_Account::CLOSED, NOT_EQUAL)))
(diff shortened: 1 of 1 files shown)


guanlisheng (approved):

georgeef:
@guanlisheng
i'll create a new issue for the problems in `cashflow.cpp`. anyway it is good to have some documentation about the changes.

PR #6848, opened 2024-08-30
refactor choices in Model_Checking

georgeef:
This PR improves a few design weaknesses in `TYPE` and `STATUS` choices in `Model_Checking`. If accepted, other PRs will implement similar improvements in other model choices.

## Design weaknesses

- The order of the entries in vector `TYPE_CHOICES` and `STATUS_CHOICES` is not checked. If the order is wrong (by mistake) and this becomes unnoticed, the database may be corrupted, since the choice values presented by the gui are different than the choice values processed by the program.

- The cases in enum `TYPE` and `STATUS_ENUM` pollute the `Model_Checking` class namespace, especially if they have a common short name like `VOID`. Since they are not prefixed, they are sometimes confusing when used outside of `Model_Checking`.

- The single-letter keys of `STATUS` appear as raw values in several places, i.e., their values are not encapsulated.

- `status_key` (previously called `toShortStatus`) may generate an invalid status key (single-letter code), if its input argument is wrong (by mistake).

## Changes

- Rename enum `TYPE` -> `TYPE_ID`, `STATUS_ENUM` -> `STATUS_ID`.

- Add prefix `TYPE_ID_` in the cases of `TYPE_ID`. Similar for `STATUS_ID`.

- Add the status keys (single-letter) in `STATUS_CHOICES`, together with the long names for each case.

- At initialization phase (only once), create wxArrayString `TYPE_STR`, `STATUS_KEY`, `STATUS_STR`, and at the same time check the order of `TYPE_CHOICES`, `STATUS_CHOICES`.

- At initialization phase, create wxString `TYPE_STR_*`, `STATUS_KEY_*`, `STATUS_STR_*` for all cases in `TYPE_ID`, `STATUS_ID`. These symbols simplify the code outside of `Model_Checking`.

- Rename `all_type()` -> `type_str_all()` and make it private. Similar for `all_status()` -> `status_str_all()` and `all_status_key()` -> `status_key_all()`.

- Rename `type()` -> `type_id()`, in order to clarify that this function returns a `TYPE_ID`. Similar for `status()` -> `status_id()`.

- `status_key()` (previously called `toShortStatus`) converts a long name to a key (single-letter). Do not rely on the first letter of the input argument (it may be wrong by mistake). Instead, convert the long name to `STATUS_ID` using `status_id()` and then to key using wxArrayString `STATUS_KEY`.

- Replace all raw values of status keys to `Model_Checking::STATUS_KEY_*`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6848)

### DIFF
--- a/src/billsdepositsdialog.cpp
+++ b/src/billsdepositsdialog.cpp
@@ -152,7 +152,7 @@ mmBDDialog::mmBDDialog(wxWindow* parent, int bdID, bool duplicate, bool enterOcc
         }
     }
 
-    m_transfer = (m_bill_data.TRANSCODE == Model_Checking::TRANSFER_STR);
+    m_transfer = (m_bill_data.TRANSCODE == Model_Checking::TYPE_STR_TRANSFER);
 
     int ref_id = m_dup_bill ?  -bdID : (m_new_bill ? 0 : -m_bill_data.BDID);
     m_custom_fields = new mmCustomDataTransaction(this, ref_id, ID_CUSTOMFIELDS);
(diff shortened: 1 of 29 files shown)


guanlisheng (approved):

guanlisheng:
+ @whalley and @n-stein fyi

Issue #6860, opened 2024-09-09
Add executions of scheduled transactions in transaction panels


georgeef:
This is a proposal to add an option to include scheduled transactions in transaction panels.

## Motivation

The transaction panels (`mmCheckingPanel`) currently show only entries in the `CHECKINGACCOUNT_V1` table, while the scheduled transactions panel (`mmBillsDepositsPanel`) shows only entries in the `BILLSDEPOSITS_V1` table. This simplifies the implementation, since each panel needs to deal with only one type of transactions.

From the user's point of view, a single list containing both executed and scheduled transactions, in a selected period from past to future, gives a better overview. It also makes the processing of transactions more convenient, since a single interface can provide the functionality which is curretly separated in two interfaces.

The transaction panels are most suitable for the display and processing of each individual transaction, either executed or scheduled, either in the past or in the future. They should give options to select any desirable combination. On the other hand, the report panels (mostly) present aggregate information, and they should also give options to select any combination between executed and scheduled transactions, or between past and future.

This issue proposes a fusion between (already executed) transactions and future executions of scheduled transactions, in one panel. Another issue will propose a similar fusion in the report panels.

## Implementation

- Define a new class, which represents either a transaction or an execution of a scheduled transaction. A single scheduled transaction unrolls to one or more executions, which differ only in the execution time (`TRANSDATE`). This class contains a member field which represents the repetition number, starting from 1 for executions of a scheduled transaction, and set to 0 for transactions. This field discriminates between the two cases. 
 
- A database typically contains only a limited number of scheduled transactions, however the list of their executions may be large, depending on the frequency and the selected period. In order to improve efficiency, an execution of a scheduled transactions is represented by the execution date, the repetition number, and a link to the original scheduled transaction.

- The transactions and the scheduled transactions are fused together as follows: a list of transactions is generated as before; a list of scheduled transactions is generated, if this options has been selected; a list of executions is generated and sorted by date. The two lists (transactions and executions of scheduled transactions) are then processed in parallel; in each iteration the one with the smaller date (`TRANSDATE`) is selected and stored.

- Once a single list of fused transactions is available, most of the interface functionality is the same for all, however several actions must differentiate between a transaction and an execution of a scheduled transaction.

## Task list
- [X] #6861
- [X] #6873
- [X] #6878
- [X] #6880
- [X] #6981
- [X] #6983
- [X] #7081
- [X] #7084

georgeef:
related issue: #2225

[...]

PR #6861, opened 2024-09-09
new(#6860): show scheduled transactions in mmCheckingPanel

georgeef:
This PR adds an option to include scheduled transactions in `mmCheckingPanel`, as proposed in #6860.

## Implementation

- Declare struct `Fused_Transaction::Data` derived from `Model_Checking::Data`, which represents either a transaction or a single execution of a scheduled transaction. It contains the additional fields `m_bdid` and `m_repeat_num`. For a transaction, `m_bdid` and `m_repeat_num` are set to 0. For an execution of a scheduled transaction, `m_bdid` is the `BDID` of the original scheduled transaction, `TRANSID` is set to 0, `m_repeat_num` is the repetition number (starting from 1), `TRANSDATE` is the date of execution (it may be different than the `TRANSDATE` of the original scheduled transaction), and all other data fields are as if the scheduled transaction has been executed.
 
- Declare struct `Fused_Transaction::Full_Data` derived from `Model_Checking::Full_Data`, with the additional fields `m_bdid` and `m_repeat_num`. For an execution of a scheduled transaction, `m_tags` is set to the tags of the original scheduled transaction, while `m_splits` contains the execution of the original scheduled transaction splits (with `TRANSID` set to 0).

- Define `Model_Billsdeposits::unroll(r, end_date)`, which returns the list (vector) of execution dates for the scheduled transaction `r`, until `end_date`. An optional argument `limit` can be used in order to limit the number of executions.

- Extend the main loop in `mmCheckingPanel::filterTable()` to also process scheduled transactions. If scheduled transactions are disabled, the loop behavior and efficiency are the same as before. The vector of transactions and the executions of scheduled transactions are processed in parallel (they are both sorted by date), choosing the entry with the smaller date in each iteration (or choosing a transaction, if both have the same date).

# Interface changes

- In the header (top) area of transactions panels, show the balance information above (closer to the title) and the controls below (closer to the main action area).

- If all transactions are reconciled, then hide `Reconciled Bal` and `Diff`, since they are redundant. If at least one transaction is not reconciled, show `Reconciled Bal` and `Diff`, even if the difference is zero (this also serves as a warning to the user that there are unreconciled transactions).

## Additional changes
- Rename enum cases `MENU_VIEW_*` to `FILTER_ID_*` in `mmcheckingpanel.h`.
- Remove `VIEW_TRANS_*_STR` from `constants`. Add `FILTER_STR` in `mmcheckingpanel`.
- Refactor `Model_Checking::Full_Data` constructors.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6861)


georgeef:
replaced the "Schduled Transactions" checkbox (which had quite long text), with a wxBitmapToggleButton containing the RECURRING bitmap (same as in nav tree). i tested on macOS, according to wxWidgets manual this button is not available in all systems. please check.

georgeef:
there is an error in static code analysis, but i don't understand what kind of error.

georgeef:
new commit:
- fix issue in static analysis
- change `FILTER_ID_NONE` to `FILTER_ID_ALL`
- set min width of the filter button to 200

georgeef:
new commit: add dynamic tooltip on scheduled button (depending on its state).

georgeef:
new commit: hide reconciled and diff in header, if all transactions are reconciled (see also updated description).

georgeef:
Replaced by #6981.

PR #6873, opened 2024-09-13
new(#6860): fix Model_Checking::(amount, balance)

georgeef:
This PR depends on #6861.

## Terminology

The usage of terms "amount" and "balance" at the interface and in different places in the source code is confusing.

### Proposed terminology:

- The "amount" of a transaction corresponds to the `TRANSAMOUNT` and `TOTRANSAMOUNT` fields in `Model_Checking::Data`, and it is (normally) a positive quantity. It means the amount (of money) moved out of an account or into an account.

- The "flow" of a transaction is a signed quantity. It is (normally) positive for a deposit and negative for a withdrawal. The "inflow" is a non-nagetive quantity, as defined in `Model_Checking::deposit()`. The "outflow" is a non-nagetive quantity, as defined in `Model_Checking::withdrawal()`.

- The "balance" of an account is a signed quantity, as defined in `Model_Account::balance()`. Transactions do not have a "balance".

All the above terms refer to a specific account, which also determines the currency. There is no such thing as the "amount of a transaction", unless a specific account is defined or implied. At a higher level, where currency conversions are handled, the above terms can be extended, using the base currency as unit, e.g., one can talk about the "flow/inflow/outflow of a set of transactions on a set of accounts". But such extensions are not possible at the level of `Model_Checking`.

## Model_Checking

`amount(r, account_id)` returns the flow of transaction `r` for the account `account_id`. It is a low level function, in the sense that it only changes the sign of the amounts found in `r`, without any attempt to perform currency conversions.

If `account_id` is a valid account and `r` is transaction of `account_id` (i.e., `account_id` is equal to either `r.ACCOUNTID` or `r.TOACCOUNTID`), then its return value is as expected.

If `account_id` is equal to -1 (which in `mmCheckingPanel` means "any account"), and if `r` is a withdrawal or deposit, then the return value is the flow on the single account in `r`. But if `r` is a transfer, then the return value is the flow on `TOACCOUNTID`, which is arbitrary (why `TOTRANSAMOUNT` is preferred to `-TRANSAMOUNT`?).

If `account_id` is a valid account but `r` is not a transaction of `account_id`, then the return value is rather unexpected, for any type of transaction. `amount()` assumes, but does not check, that `r` is a transaction of `account_id`.

`Model_Checking` provides two functions, `amount()` and `balance()`, for almost the same purpose and with confusing names. `balance()` ignores void and deleted transactions, which is what is needed in almost all calculations. `amount()` considers also void and deleted transactions, which is rarely needed.

### Changes:

- In struct `Full_Data`, add the fields `ACCOUNTID_W` (withdrawal account_id), `ACCOUNTID_D` (deposit account_id), `TRANSAMOUNT_W` (withdrawal amount), `TRANSAMOUNT_D` (deposit amount), and rename the fields `AMOUNT` -> `ACCOUNT_FLOW`, `BALANCE` -> `ACCOUNT_BALANCE`. These fields are used as cache by `mmCheckingPanel`.

- Remove `amount()` in order to simplify the `Model_Checking` interface, and move its functionality in `balance()` and where else it is needed.

- In `balance(r, account_id)`, check that `r.ACCOUNTID` or `r.TOACCOUNTID` is equal to `account_id`, otherwise return 0.

- In the declaration of `balance()`, `withdrawal()`, remove the default value for `account_id`.

- Rename `balance()` -> `account_flow()`, `withdrawal()` -> `account_outflow()`, `deposit()` -> `account_inflow()`, `reconciled()` -> `account_recflow()`.

## mmCheckingPanel

The transactions panel (`mmCheckingPanel`) for a specific `account_id`, shows the withdrawal/deposit amounts for `account_id` only, as numeric values, since the currency is implied. For transfers, only the amount for `account_id` is shown. For void transactions, the amount is shown but does not affect the account balance; this amount should better be stricken through. Exactly one of the columns "Widthdrawal" and "Deposit" has an amount, which is the "amount" of this transaction for this account.

The "All Transactions" panel shows the withdrawal/deposit amounts together with the currency of the corresponding account. For transfers, only the withdrawal side is shown, although the deposit amount may be different (or in different currency); it should show both sides. On the other hand, the right-click menu on a transfer amount, shows the (hidden) deposit amount, which is confusing. For void transactions, zero amount is shown; it should show the amount as in an individual account panel. The term "amount" in the context of "All Transactions" is ill-defined; in general a transfer transaction has two (probably different) amounts, in two (probably different) currencies. In this context, the terms "withdrawal amount" and "deposit amount" should be used instead. Since there is no single implied currency, each amount is shown together with the corresponding currency.

The "Deleted Transactions" panel should be consistent with the "All Transactions" panel, since it also does not have a single implied currency. All amounts are shown without currency; they should have a currency. For transfers, it shows only the deposit side; it should show also the withdrawal side.

### Changes:

- Rename `m_filteredBalance` -> `m_account_flow`.

### Interface changes:

- In the panel header for a single account, change `Filtered View Bal` -> `Filtered Flow`.

- Add a marker in front of the amount of void transactions (strikethrough would be better).

- In the "All Transactions" panel, add missing deposit amount in transfers.

- In the "All Transactions" panel, show the amount of void transactions.

- In the "Deleted Transactions" panel, add missing withdrawal amount in transfers.

- In the "Deleted Transactions" panel, add the currency in each amount.

- Fix the amount shown in the right-click menu.

## Backtracking and derived changes

(The names in this section are the original names, before any renaming.)

`Model_Checking::amount()` is called directly or indirectly by:

- `balance()`, `withdrawal()`, `deposit()`, `reconciled()`.

- `mmCheckingPanel::filterTable()`, with `m_AccountID` that may be -1: the return value is stored in `m_account_balance`, `m_reconciled_balance`, `m_filteredBalance`, `full_tran.AMOUNT`, `full_tran.BALANCE`.

  - `m_account_balance`, `m_reconciled_balance`, `m_filteredBalance` are used by `mmCheckingPanel::setAccountSummary()` only if `!isAllAccounts_`.

  - `m_account_balance` is used by `TransactionListCtrl::On(Duplicate|Edit|New)Transaction()` -> `mmTransDialog` constructor: its value is stored in `mmTransDialog m_current_balance`, but not used.

  - `full_tran.AMOUNT` is used (mostly in a wrong way) by `TransactionListCtrl::getItem()` and `TransactionListCtrl::OnMouseRightClick()`. It is also used by `Model_Checking::SorterBy(DEPOSIT|WITHDRAWAL)`, which are wrong if the list of transactions contains transfers.

  - `full_tran.BALANCE` is used by `TransactionListCtrl::getItem()`, `TransactionListCtrl::OnMouseRightClick()`, `Model_Checking::SorterByBALANCE`, only in the panel of an individual account.

- `mmCheckingPanel::updateExtraTransactionData()`: the return value is stored in double `balance`, but it is not used if `m_AccountID` is equal to -1.

- `Model_Category::getCategoryStats()`, for withdrawal or deposit, and without account id.

- `mmReportForecast::getHTMLText()`, for withdrawal or deposit, with account id equal to -1.

In all other direct or indirect calls to `amount(r, account_id)`, `account_id` is a valid account and `r` is a transaction of `account_id`.

### Changes:

- Remove `mmTransDialog m_current_balance`. Remove `current_balance` argument in `mmTransDialog` constructor.

- In `TransactionListCtrl::getItem()`, fix the calculation of withdrawal and deposit amounts.

- In `TransactionListCtrl::OnMouseRightClick()`, fix the calculation of amounts in the menu.

- Fix `Model_Checking::SorterBy(DEPOSIT|WITHDRAWAL)`.

- In `mmCheckingPanel::updateExtraTransactionData()`, do not call `balance()` if `m_AccountID` is equal to -1.

- In `Model_Category::getCategoryStats()`, add `transaction.ACCOUNTID` as second argument to `balance()`.

- In `mmReportForecast::getHTMLText()`, add `trx.ACCOUNTID` as second argument to `withdrawal()` and `deposit()`.

## Notes

(Names after all changes above.)

- `account_flow(r, account_id)` assumes that `r.ACCOUNTID != r.TOACCOUNTID`. This invariant is enforced by all functions which create or update transactions and it is not checked (it is tedious to check it in all places where it is important). It would be good to have a centralized check for this and all other database invariants.

- If `TRANSAMOUNT` and `TOTRANSAMOUNT` are non-negative (as it should normally happen), then the inflow is one of 0, `TRANSAMOUNT_D`, while the outflow is one of 0, `TRANSAMOUNT_W`. However, for generality, `account_inflow()` and `account_outflow()` do not assume anything about the sign of `TRANSAMOUNT` and `TOTRANSAMOUNT`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6873)


georgeef:
Replaced by #6981.

Issue #6935, opened 2024-10-23
"Find all transactions" not working from checking account panels


n-stein:
MMEX version:
 - [X] 1.8.1

In any account panel other than the "All Transactions" view, right clicking and selecting "Find all transactions with ..." jumps to the "All Transactions" page, but doesn't automatically apply the filter. The filer is saved correctly, the panel is just defaulting to "View All Transactions" instead of "View Transaction Report..."

n-stein:
Seems it was broken by e88c3e58df3584ab19ab09c30a3100f3d2aa7eab

n-stein:
Fixed by using the defined constant instead of hard coding.

ovari:
Thank you for fixing what we broke. Is it correct to assume it is fixed for user interfaces set to different locales too? Thank you

georgeef:
Is #6874 related to this?

n-stein:
> Is #6874 related to this?

Yup, same issue. Sorry I didn't see yours was already logged. Should be fixed now.

PR #6936, opened 2024-10-23
fix(#6935): use constant for right click filter type

n-stein:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6936)


whalley:
@n-stein Think we can ditch the wxString(.....)

n-stein:
> @n-stein Think we can ditch the wxString(.....)

Done

PR #6939, opened 2024-10-27
Cleanups: Dangling pointers, unused variables, signness

gerstrong:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6939)

### DIFF
--- a/src/import_export/univcsvdialog.cpp
+++ b/src/import_export/univcsvdialog.cpp
@@ -198,7 +198,7 @@ void mmUnivCSVDialog::CreateControls()
     m_text_ctrl_->Connect(ID_FILE_NAME
         , wxEVT_COMMAND_TEXT_ENTER, wxCommandEventHandler(mmUnivCSVDialog::OnFileNameEntered), nullptr, this);
 
-    const wxString& file_button_label = _("&Browse...");
+    const auto file_button_label = wxString("&Browse...");
     wxButton* button_browse = new wxButton(itemPanel6, wxID_BROWSE, file_button_label);
     itemBoxSizer7->Add(button_browse, g_flagsH);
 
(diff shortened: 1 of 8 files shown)


src/import_export/univcsvdialog.cpp
```diff
@@ -198,7 +198,7 @@ void mmUnivCSVDialog::CreateControls()
     m_text_ctrl_->Connect(ID_FILE_NAME
         , wxEVT_COMMAND_TEXT_ENTER, wxCommandEventHandler(mmUnivCSVDialog::OnFileNameEntered), nullptr, this);
 
-    const wxString& file_button_label = _("&Browse...");
+    const auto file_button_label = wxString("&Browse...");
```
vomikan, line 201: why?

src/mmSimpleDialogs.cpp
```diff
@@ -1017,7 +1017,7 @@ void mmErrorDialogs::InvalidFile(wxWindow *object, bool open)
 
 void mmErrorDialogs::InvalidAccount(wxWindow *object, bool transfer, TOOL_TIP tm)
 {
-    const wxString& errorHeader = _("Invalid Account");
+    const auto errorHeader = wxString("Invalid Account");
```
vomikan, line 1020: Should be  wxGetTranslation("Invalid Account") ?
or just "Invalid Account"

src/import_export/univcsvdialog.cpp
whalley: Good point, I don't see an issue with... `const wxString file_button_label = _("&Browse..."); `

src/mmSimpleDialogs.cpp
whalley: Good point, I don't see an issue with... `const wxString errorHeader = _("Invalid Account"); `

Issue #4690, opened 2022-06-21
Secondary 'Sort By' Column


vomikan:
https://forum.moneymanagerex.org/viewtopic.php?t=10552

When sorting by transactions by 'Payee' in Account View, transactions are sorted by 'Payee', then seemingly by ID as a secondary value. It would be great if we could select the secondary sorting to be via Date (or any other value desired).

The issue with the current solution is that when future transactions are entered for the same payee, but then you add additional transactions in a shorter timeframe, there is no way to have them sort in ascending date ordered by AND payee, without deleting/recreating them in the desired order, to get ascending IDs.

whalley:
How about having the secondary sort set to the last sorte column.
E.g. 
1. sort by Date
2. sort by Payee (secondary sort is now Date)
3. sort by ID (secondary sort is now Payee)
4. sort by Payee (secondary sort is now by ID)

Just a thought. Avoids any UI changes,

tactilis:
> How about having the secondary sort set to the last sort column.

I have sometimes wished for this myself but I can see good and less good consequences of this.

1. At present, if you enter transactions manually as you actually make them or, more likely, import the transactions from a bank account, then simply clicking on the Date column to sort by Date>ID will always match the order of transactions as entered (i.e. same order as on your bank statement).

    If we implement 'last sort column becomes the secondary sort', then simply clicking on the Date column will not necessarily give the same order as in the bank. We are likely to get more people asking questions such as _"Why doesn't the MMEX balance match my bank balance"_, or _"Why can't MMEX add up the balances correctly"_, which we have often seen on the forum.

2. Because of the above, does it make sense to show visually which column is being used as the secondary sort?

3. In the case of click to sort by Date, then click to sort by Payee (for example), should the secondary sort by Date remember the order (ascending / descending) that the user had set before clicking on Payee? If so, and we provide a visual indicator of the secondary sort column, then should this show whether the sort is ascending / descending?

4. Following on from 3, if the user clicks on Payee again, should this just reverse the Payee order but leave the secondary date order as it was?

5. At present, the user selected sort column _and_ direction of sort is saved (in mmexini.db3 I presume) so that this is remembered for the next MMEX run. If this change is implemented, then we should also remember the user-selected secondary sort order.
  
   
 
>    
> Avoids any UI changes
 
Maybe.
To be debated.

tactilis:
> The only thing that is needed is for the primary and secondary sorting icon to differ.

For example:
 
[...]

PR #6964, opened 2024-11-25
SUID support

guanlisheng:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6964)

### DIFF
--- a/src/db/DB_Table.h
+++ b/src/db/DB_Table.h
@@ -12,7 +12,7 @@
  *      @brief
  *
  *      Revision History:
- *          AUTO GENERATED at 2024-11-30 08:58:01.982619.
+ *          AUTO GENERATED at 2024-12-04 15:54:58.326993.
  *          DO NOT EDIT!
  */
 //=============================================================================
@@ -20,6 +20,7 @@
 
 #include <vector>
 #include <map>
+#include <random>
 #include <algorithm>
 #include <functional>
 #include <cwchar>
@@ -67,6 +68,19 @@ struct DB_Table
     void drop(wxSQLite3Database* db) const
     {
         db->ExecuteUpdate("DROP TABLE IF EXISTS " + this->name());
+    }
+
+    static wxLongLong newId()
+    {
+        // Get the current time in milliseconds as wxLongLong
+        wxLongLong ticks = wxDateTime::UNow().GetValue();
+        // Generate a random 3-digit number (0 to 999)
+        std::random_device rd;
+        std::mt19937 gen(rd());
+        std::uniform_int_distribution<int> dist(0, 999);
+        int randomSuffix = dist(gen);
+        // Combine ticks and randomSuffix
+        return (ticks * 1000) + randomSuffix;
     }
 };
 
--- a/src/db/DB_Table_Accountlist_V1.h
+++ b/src/db/DB_Table_Accountlist_V1.h
@@ -12,7 +12,7 @@
  *      @brief
  *
  *      Revision History:
- *          AUTO GENERATED at 2024-11-30 08:58:01.982619.
+ *          AUTO GENERATED at 2024-12-04 15:54:58.326993.
  *          DO NOT EDIT!
  */
 //=============================================================================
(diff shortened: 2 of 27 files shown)


whalley:
@guanlisheng I presume we no longer need this PR?

guanlisheng:
hi @vomikan, @whalley , @n-stein ,can you take a look 

we can step to SUID now after #6965 is done

georgeef:
The SUID is of little interest to the user because it is too long to remember, however SUIDs are monotonic in time, so sorting by SUID is the same as sorting by time of creation. Old ID's are small numbers and have been created before the introduction of SUID, so they are also sorted by time of creation.

**Proposal:** instead of the ID (SUID), show the sequence number, starting from 1 for the first item in the list and incremented by 1 for each next item, with a label like `#` or `S/N`.

whalley:
> **Proposal:** instead of the ID (SUID), show the sequence number, starting from 1 for the first item in the list and incremented by 1 for each next item, with a label like `#` or `S/N`.
> 
> 
> 
> The ID can be a separate column, which is hidden by default.

That would work, but suspect we don't need to display the ID as we don't display for any other tables and may confuse. Perhaps we just replace display of ID by the sequence number.

whalley:
Can change DB SELECT queries to do an SQL RANK to generate the sequence number...

E.g. `SELECT *,  RANK () OVER (ORDER BY TRANSID ASC) SEQNO FROM CHECKINGACCOUNT_V1;`

I'm not very good with Python and it will need tweaks to the `sqlite2cpp.py` to add this for the tables.

georgeef:
Maybe you don't need DB support to assign sequence numbers, you can do it on the fly when you generate the list of items to be displayed. If they are stored in a table, the sequence number is the table index plus 1. If there are more items in the table (like executions of scheduled transactions), you can add an additional field for the sequence number in an auxiliary table and fill it in on the fly. 

Or you may assign sequence numbers also to scheduled executions. But you need to distinguish between transactions and scheduled executions in the panels, I did this by showing an empty ID field for scheduled executions.

The sequence number is not a sticky information applied to a transaction. Different filters or even a redraw may assign different sequence numbers. It is just a visual aid to the users and a handle to sort by time of creation, but its value is actually opaque.

georgeef:
My proposal was to assign sequence numbers when the list is generated, and to preserve them until a new list is generated. Sorting by another column of course does not change the sequence numbers. So in this respect a sequence number behaves like a temporary ID.

Sorry for not being very clear in my first proposal.

georgeef:
By the way, for almost all (I think for all) other tables, there is no need to display the ID or a sequence number, because they have either a unique key which is more important, or other fields which are sufficient for ordering.

In transaction panels, the ID is essentially used as a substitute for a missing time-of-creation field. There is a date field, but it is not enough for sorting by time of creation.

georgeef:
I think you need to store the sequence number in a separate field, you cannot just overwrite the ID (SUID) with a sequence number in the list of transactions. If the user wants to edit or process an item in the list, then you need the ID of this item in order to interact with DB.

whalley:
> I think you need to store the sequence number in a separate field, you cannot just overwrite the ID (SUID) with a sequence number in the list of transactions. If the user wants to edit or process an item in the list, then you need the ID of this item in order to interact with DB.

Agreed, my original thought was to generate a sequence number in addition to all the 'real' DB fields (including ID) hence the SQL RANK(). But as you note we only really need this for the one table view so perhaps we just code for it there.

georgeef:
@whalley 
Since the display of ID is under revision, I have a relevant question/request.

`TransactionListCtrl::OnColClick` forces the secondary sort column to be ID when the primary is Date. This is sometimes annoying, because there is no way to sort by Date/Number or any pairing with Date.

Could we remove this restriction, so the user can select any secondary together with Date, like in all other cases?  Or if Date/ID is often needed, could we assign it to a modifier-key + click on Date?

whalley:
> @whalley Since the display of ID is under revision, I have a relevant question/request.
> 
> `TransactionListCtrl::OnColClick` forces the secondary sort column (`prev_g_sortcol`) to be ID when the primary (`g_sortcol`) is Date. This is sometimes annoying, because there is no way to sort by Date/Number or any other pairing with Date.
> 
> Could we remove this restriction, so the user can select any secondary together with Date, like in all other cases? Or if Date/ID is often needed, could we assign it to a modifier-key + click on Date?

There is discussion on this here: https://github.com/moneymanagerex/moneymanagerex/issues/4690

guanlisheng:
guys, how about this PR itself?

guanlisheng:
I'm going to merge it if there is no other objection. 

Large id would be from mobile devices as sync data is common.

georgeef:
@whalley, @guanlisheng
I can implement sequence numbers in transaction panels, unless you already work on it. I estimate about 1 day.

whalley:
> @whalley, @guanlisheng I can implement sequence numbers in transaction panels, unless you already work on it. I estimate about 1 day.

Go for it. I've got used to the large transaction IDs now :-) But it would be a good usablility feature to have the sequence number.

n-stein:
Is there a way the sequence number can be derived based on transaction data? It would be good to have a calculated sequence number that provides consistency, otherwise it might be confusing if the same transaction had a different sequence number in the account view, account group view, all transactions view, and transaction report.

georgeef:
The idea is that the sequence number is only a placeholder for sorting purposes, and its value is of no interest. It is not any kind of identification and it is not sticky (it may change at every refresh). In my proposal above it is created on the fly each time a panel is created or refreshed.

Essentially it is used instead of the creation datetime. In other panels and reports the ID is also not included.

The main problem in the identification of transactions is that they don't have a unique key. Even if time is enabled, the DB does not guarantee unique datetime. ID is now way too long to be useful for this purpose.

For cross reference between panels/reports, it is maybe more convenient to have some kind of hyperlinks, instead of asking users to memorize IDs.

georgeef:
I think it is better to have both the ID and SN as separate columns. The users can choose which one to keep and which to hide.

rmelillo76:
@georgeef - thoughts on changing the column header text from "SN" to "Sequence #" or just "#" instead? I realize "SN" is Sequence Number, but it may not be obvious to most users. The # is generally known as the number sign.

For example, we do this on the Import dialog to number the transactions:
image

georgeef:
This was one of the options, it also has the big advantage that it doesn't need translation. On the other hand # doesn't look so nice in the sort indicator and SN is more symmetric to ID. # could be confusing, it might be interpreted as ID, maybe it is used already with different meaning somewhere else.
At the end I decided to use the more descriptive name SN. The meaning can be explained in a tooltip.

georgeef:
Added a tip for SN in #7095.

PR #6965, opened 2024-11-29
C++ Int64/wxLongLong to represent SQLite3 INTEGER

guanlisheng:

https://www.sqlite.org/autoinc.html

> In SQLite, a column with type INTEGER PRIMARY KEY is an alias for the [ROWID](https://www.sqlite.org/lang_createtable.html#rowid) (except in [WITHOUT ROWID](https://www.sqlite.org/withoutrowid.html) tables) which is always a 64-bit signed integer.

Let's use wxLongLong as int64 for the best compatibility 
1. `typedef wxLongLong int64`
2. use `GetInt64` to replace `GetInt` for [wxSQLite3ResultSet](https://utelle.github.io/wxsqlite3/docs/html/classwx_s_q_lite3_result_set.html#a4b80559aac5a6d78a965aa9580d53684) 
3. use `std::vector<int64>` to replace `wxArrayInt`
4. `json_writer.Int64(xxx.GetValue())` to replace `json_writer.Int(xxx)`
5.  `row(L"xxxID") = xxxID.GetValue()` to replace `row(L"xxxID") = xxxID`
6. Non-ID integer, just extra call `xxx.GetValue()`, e.g.  `REPEATS`

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6965)


guanlisheng:
hi @vomikan , @whalley and @n-stein 

here you go the change to use int64 to represent the SQLite Integer field.  can you take a look ?

src/db/DB_Table.h
```diff
@@ -35,6 +35,8 @@ using namespace rapidjson;
 #include "html_template.h"
 using namespace tmpl;
 
+typedef wxLongLong int64;
```
whalley, line 38: Not that it really matters, but should we be using wxLongLong as this is indicated as obsolete.  https://docs.wxwidgets.org/3.2/classwx_long_long.html

whalley: Have built on MacOS and confirm it builds and runs fine.

guanlisheng: Thanks for trying. 

a major reason is wxSQLite3 still uses this in its GetInt64 

https://utelle.github.io/wxsqlite3/docs/html/classwx_s_q_lite3_result_set.html#a4b80559aac5a6d78a965aa9580d53684

guanlisheng: we may consider the switch along with wxSQLite3

@vomikan , @n-stein , can you help on win and linux platforms?

guanlisheng: + @georgeef and @wolfsolver as FYI 
then all mobile and desktop mmex would work on large IDs

guanlisheng:
Guys, 
I'm merging this, then we will have a better ID handling.

guanlisheng:
this is addressing #6992

n-stein:
It crashes when running the wxWidgets debug builds. The format specifiers for wxString::Format are still using "%i" but they need to be using "%lld" for anything passed as an int64.

n-stein:
I think I got all of them in #6994

guanlisheng:
thank you, @n-stein

guanlisheng:
#7007

PR #6967, opened 2024-11-29
budget report link.

giongy:
To me, budget performance report is the most useful section of money manager ex. 
I use it to understand where I'm missing my goal and where I have some room to play; maybe that month I was good in spending for a category and want to "move" part of that budget to next month. 
Having the whole vision of the budget (year+months) it's easier to see what is wrong, what is fine, what is adjustable.
So, I'm trying to make a link on this report in order to access directly to the budget entry dialog of that category and that month to edit it. 
It works, even if there are still some work to do. It's my first time on c++ , I think this result is my max, maybe some of you can refine it if you think is valuable.
Any way, from here:
image
estimate value is now a link, and if you click on that link , budget entry dialog open:
image
then you can change value in the dialog and... it works!
I trying to refresh the report once an entry was edited..

What do you think? could be useful also for other, in your opinion? if yes, I'm afraid that I need some help because I'm not sure about the code. and things are missing, like estimate and actual in the dialog:
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6967)


giongy:
there is need also to check if monthly budget is created or is missing when you click.

CMakeLists.txt
whalley, line 1: Is the rest of the applicstion fine with wxUSE_WEBVIEW_EDGE? I note the wxWidgets document notes some limitayion around this, i.e. lack of VFS support. Perhaps a developer on Windows can review and comment further.

CMakeSettings.json
whalley, line 1: I don't use VS so can't comment. Perhaps a developer on Windows can review and comment further.

src/reports/budgetingperf.cpp
```diff
@@ -26,7 +26,12 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 #include "model/Model_Budget.h"
 #include "model/Model_Category.h"
 #include "reports/mmDateRange.h"
-
+#include <string>
+namespace tags
+{
+static const wxString TABLE_CELL_LINK = R"(<a href="%s" target="_blank">%s</a>)";
```
whalley, line 32: Think we already have mmHTMLBuilder::addTableCellLink is it posible to re-use this in some way?

CMakeLists.txt
giongy: EDGE was just a test, it's not needed, sorry, forgot to remove

src/reports/budgetingperf.cpp
giongy: I can try to reuse that, but I still need to have the TABLE_CELL_LINK_RED.

whalley: Perhaps include TABLE_CELL_LINK_RED in mmHTMLBuilder also? Or add a colour option to addTableCellLink? Makes it usable elsewhere.

giongy: since that cell contain a span:
```
hb.startSpan((budgetDeductMonthly && estimate != 0 && round(estimateTotal[12] / budgetStats[catID][12] * 100) / 100 > 1)
                 ? hb.getFormattedLink("red",editBudgetEntry,estimateVal)
                 : hb.getFormattedLink("", editBudgetEntry, estimateVal),
             wxString::Format(" style='text-align:right;%s' nowrap", ""));
hb.endSpan();
```
I cannot use the existent `addTableCellLink`. 
I create another method `getFormattedLink` in mmHTMLBuilder and a new TABLE_CELL_LINK_COLOR that can manage color.


```
static const wxString TABLE_CELL_LINK = R"(<a href="%s" target="_blank">%s</a>)";
static const wxString TABLE_CELL_LINK_COLOR = R"(<a style="color: %s;" href="%s" target="_blank">%s</a>)";
```

should be flexible and reusable

giongy: @whalley  thank  you very much for accepting my proposal

giongy:
@whalley Thank you very much for accepting my proposal

whalley:
@giongy Having merged I've noticed a few issues (on MacOS, not sure about other platforms) regarding URL escapes characters like | to %7F and how this is handled. Also I wonder whether using the string formats for currency (with separators, etc) is the right thing to do. I'll do some more investigation later.

PR #6995, opened 2024-12-04
fix navtree

georgeef:
In Navigator tree, align Dashboard to other top-level items.

## Implementation:
- Add a non-visible root item.
- Dashboard is the first child of the root item (it was the root before).
- Adjust spacing of open/close symbols (a little more compact layout, the symbols still show clearly the hierarchy).

## Test
- Tested on macOS.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/6995)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -9,8 +9,8 @@ assignees: ''
 
 MMEX version:
  - [ ] 1.7.0
- - [X] 1.8.0
- - [ ] 1.8.1
+ - [ ] 1.8.0
+ - [x] 1.8.1
  - [ ] 1.8.2
  - [ ] 1.9.0
  - [ ] Other (please specify)


n-stein:
> Add a non-visible root item.
> Dashboard is the first child of the root item (it was the root before).

This causes a crash whenever the nav tree is reset (#6998). I've put fixes under PR #6999.

> Adjust spacing of open/close symbols (a little more compact layout, the symbols still show clearly the hierarchy).

I think this is making it impossible to collapse first-level nodes on Windows.

georgeef:
Thank you @n-stein 

On macOS I don't see any of these issues. I compiled wxWidgets-3.2.6 with `--enable-debug` and MMEX with `-DCMAKE_BUILD_TYPE=Debug`. I don't get any alert at startup, the top-level items can expand and collapse, and I could delete a transaction without crash.

@whalley Do you see the issues above?

georgeef:
This is how it looks on macOS:

809C955C-DE72-45A8-A12A-22AC851D532C

georgeef:
I tested #6999 on macOS, it looks good (I don't see any difference).

ovari:
Update [screenshots](https://github.com/moneymanagerex/moneymanagerex?tab=readme-ov-file#screenshots) required before next release?

ovari:
Right-clicking on "Dashboard" brings up the "Theme Manager". Is that the desired behavior? Thank you

georgeef:
Right-click on "Dashboard" did not change with this PR. I don't know the history of this behavior, but if there are no other settings for "Dashboard", it is ok to bring something else which is useful.

It could open [Options -> Dashboard] instead, which is more relevant.

PR #7016, opened 2024-12-13
fix(#6967): budget link fixes?

whalley:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7016)

### DIFF
--- a/src/reports/budgetingperf.cpp
+++ b/src/reports/budgetingperf.cpp
@@ -205,7 +205,7 @@ wxString mmReportBudgetingPerformance::getHTMLText()
                         wxString estimateVal = Model_Currency::toString(estimate, Model_Currency::GetBaseCurrency());
                         //make the href string parameters
                         std::stringstream ss;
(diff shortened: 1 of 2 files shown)


src/reports/budgetingperf.cpp
```diff
@@ -205,7 +205,7 @@ wxString mmReportBudgetingPerformance::getHTMLText()
                         wxString estimateVal = Model_Currency::toString(estimate, Model_Currency::GetBaseCurrency());
                         //make the href string parameters
                         std::stringstream ss;
-                        ss << "budget: " << estimateVal << "|" << Model_Currency::toString(actual, Model_Currency::GetBaseCurrency()) << "|" << catID << "|"
+                        ss << "budget:" << estimate << "|" << actual << "|" << catID << "|"
```
giongy, line 208: @whalley  Thank you for fixing budget link. I noticed that decimal on the dialog doesn't work fine after your commit:
image
could be caused by changing estimate and actual in the passed string?

whalley: Ooops. Does it work fine if you just revert line 208?

Issue #7030, opened 2024-12-15
Import CSV with new lines


blueberry-bytes:
I'd like to import a CSV file with a note made in a few lines.

2024-02-09;Payee;1.00;Category;"line1
line2
line3
line4";2024-01;

In Excel it looks properly:
![image](https://github.com/user-attachments/assets/df18d5a6-9639-4920-a493-818c9bde8518)

Unfortunately import divides it into more rows.
![image](https://github.com/user-attachments/assets/a3c32b16-4060-453b-8e5d-69568580952e)

n-stein:
#1643

n-stein:
That change takes a file like this:

```
acct,12/1/2024,Test,100,"Line1
Line2
Line3"
acct,12/2/2024,Test,200,"Line1, Line2, Line3"
acct,12/2/2024,Test,200,"""Line1
Line2"", Line3"
acct,12/3/2024,Test,300,"Line1,""Line2""
Line3"
```

And imports as:

![image](https://github.com/user-attachments/assets/fb1004f1-7b33-4616-82bd-46d7a63f4f4f)

![image](https://github.com/user-attachments/assets/65d7cf0d-4975-448f-ae5b-ce9fe84b7974)

Newlines are preserved:

![image](https://github.com/user-attachments/assets/84a6e2ff-3ea5-427f-8501-810a496e30c5)

n-stein:
@blueberry-bytes if you have the time it would be great if you could give the beta a quick test, see if it handles the newlines as you are expecting.

blueberry-bytes:
Thanks a lot, it works perfectly. Sorry for the delay, I've just had time to use it.

PR #7038, opened 2024-12-18
fix(#7030): fix regex for double quote at start of field

n-stein:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7038)

### DIFF
--- a/src/import_export/parsers.cpp
+++ b/src/import_export/parsers.cpp
@@ -50,11 +50,11 @@ bool FileCSV::Load(const wxString& fileName, unsigned int itemsInLine)
     // Parse rows
     wxString line;
     int row = 0;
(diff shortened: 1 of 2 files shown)


n-stein:
1. Minor fix for regex. Old pattern didn't correctly handle when a field containing a line break began with a double quote, i.e. the CSV contained `,"""`.

2. Switched from the unicode replacement character `\ufffd` to the substitution control character `\x1a` to conform to UTF-8 for wxRegEx compile.

3. Corrected the path separator detection. Separator detection was failing because the `skip` boolean that tells whether or not the current character is inside of double quotes was not reset properly between loops over the string. When the CSV file had only one line that had hanging double quotes, e.g.:

```
acct,12/2/2024,Test,200,"""Line1
Line2"", Line3"
```
it caused the separator to be detected as `\t` when the file was opened.

PR #7045, opened 2024-12-20
cleanup

ovari:
be able → enable
could not → unable to
history → historical
needed → required
not available → unavailable
webapp → the WebApp

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7045)

### DIFF
--- a/src/maincurrencydialog.cpp
+++ b/src/maincurrencydialog.cpp
@@ -658,7 +658,7 @@ void mmMainCurrencyDialog::OnHistoryUpdate(wxCommandEvent& WXUNUSED(event))
     }
 
     wxString base_currency_symbol;
-    wxASSERT_MSG(Model_Currency::GetBaseCurrencySymbol(base_currency_symbol), "Could not find base currency symbol");
+    wxASSERT_MSG(Model_Currency::GetBaseCurrencySymbol(base_currency_symbol), "Unable to find base currency symbol");
 
     int msgResult = wxMessageBox(_("Do you want to add also dates without any transaction?")
         , _("Currency Manager")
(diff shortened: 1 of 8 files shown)


n-stein:
The 3-to-1 ellipsis character affects so many strings, we might want to consider doing the ellipses conversion in our PO files as well then uploading the translations back to Crowdin so user's don't have to retranslate just for an ellipses character change.

ovari:
Reverted the 3-to-1 ellipsis character conversion. Unsure how to:

> doing the ellipses conversion in our PO files as well then uploading the translations back to Crowdin so user's don't have to retranslate just for an ellipses character change.

n-stein:
Just a procedural method to retain the existing translations with the character change. In fact, it would be better to do the ellipses change in a separate PR that only has those changes.

ovari:
Can this patch please be merged and then another patch with ellipses can be done? Thank you

georgeef:
You can also create a new PR depending on this PR. Create a new local branch based on this PR, commit the changes, push the new branch to your fork, create new PR. I did it a few times. For clarification, in the beginning of the new PR description I write something like: This PR depends on #...

src/maincurrencydialog.cpp
```diff
@@ -432,8 +432,8 @@ void mmMainCurrencyDialog::OnListItemSelected(wxDataViewEvent& event)
                 else if (Model_CurrencyHistory::instance().find(Model_CurrencyHistory::CURRENCYID(m_currency_id)).size() > 0)
                 {
                     if (wxMessageBox(wxString::Format(_(
-                        "Historic rates for %1$s found, but \"Use currency history\" in options is disabled:\n"
-                        "click no and enable it or click yes to remove all historic rates for %2$s"),
+                        "Historical rates for %1$s found, but “Use historical currency” in options is disabled:\n"
```
n-stein, line 435: I'm a bit surprised the removal of the escape character here didn't cause any issues.

ovari: Does the smart quote require an escape character?

n-stein: Ah, I didn't see the standard `"` character was replaced with the left/right quotation mark unicode symbols. Makes sense.

PR #7048, opened 2024-12-22
cleanup: {cannot, can't} → unable to

ovari:
{cannot, can't} → unable to

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7048)


n-stein:
Most of these sound odd to me. Not quite sure why, but it seems like the focus of these sentences changes from 'you are prevented from doing this due to a restriction' to 'this object has an inability to do what you are asking'.

georgeef:
@ovari 
Why do you change "Cannot" to "Unable to"? In almost all cases the new text is bad english, and in some cases it is less precise technically. What is the problem with "cannot"?

PR #7057, opened 2024-12-24
fix(#7055): fix translations with unicode

n-stein:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7057)

### DIFF
--- a/src/assetspanel.cpp
+++ b/src/assetspanel.cpp
@@ -97,19 +97,19 @@ void mmAssetsListCtrl::OnMouseRightClick(wxMouseEvent& event)
     }
     m_panel->updateExtraAssetData(m_selected_row);
     wxMenu menu;
-    menu.Append(MENU_TREEPOPUP_NEW, _("&New Asset…"));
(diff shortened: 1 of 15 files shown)


n-stein:
@ovari fyi -- including unicode symbols in translatable strings doesn't work unless you wrap them properly.

`_("text with unicode…")` 

should be changed to

`wxGetTranslation(wxString::FromUTF8(wxTRANSLATE("text with unicode…")))`

georgeef:
@n-stein 
Can we add a macro to make it short, e.g., `_u("text with unicode…")`?

n-stein:
Yes, good idea. I'll work on that.

ovari:
> > @n-stein Can we add a macro to make it short, e.g., `_u("text with unicode…")`?
> 
> I've added a macro in #7121
> 
> `wxGetTranslation(wxString::FromUTF8(wxTRANSLATE(unicode_string)))` can now be written `_u(unicode_string)`
> 
> @ovari please use this shortened form in #7072

Done.

Issue #6847, opened 2024-08-29
about default column seq.


DevAnyKR:
MMEX version:
 - [ ] 1.7.0
 - [ ] 1.8.0
 - [x] 1.8.1 bata
 - [ ] 1.8.2
 - [ ] Other (please specify)

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [x] Windows
 - [ ] Mac OSX
 - [ ] Linux 

# Description of the bug
The positions of the columns are not saved and their order changes depending on the language.

1. by Language
a. Korean v1.8.1 beta(appveyor)  a month ago.
![image](https://github.com/user-attachments/assets/849ed9be-41a6-40bf-9675-967849b1a0e3)
column: ID, DATE, number, payee, status, category, tag, withdrawal, deposit, note, UCF1, UCF2, lastdate, **balance**, 
balance column is fixed at the end.

a. Korean v1.8.1 beta(appveyor)  today.
![image](https://github.com/user-attachments/assets/e6980042-2ec0-4269-a03b-ab880459e98c)
column: ID, DATE, number, payee, status, category, tag, withdrawal, deposit, balance, note, **UCF1, UCF2**, lastdate

b. English
![image](https://github.com/user-attachments/assets/d50f5ac4-b134-41a8-904d-9452cd441fe9)
column: ID, **UCF1, UCF2** , DATE, number, payee, status, category, tag, withdrawal, deposit, balance, note, lastdate

![image](https://github.com/user-attachments/assets/a2dd520b-194e-4181-9084-31448f034c7b)
All transactions page
account column is fixed at the end.

2. settings
![image](https://github.com/user-attachments/assets/e2752319-f68e-4dd7-9469-2b62067145c9)
Look at the order of the UCF columns. They don't appear at the beginning.
The order of columns is different in Korean and English.

# Reproduction

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:

open column order manager

Expected result:

Actual result:

# Additional Requests

Column Order manager
- allow resizing
- Column names are duplicated by language.
-- UCF columns does not appear in the English list.

jensb:
Thanks for reporting this. This feature is my addition to MMEX, it's not finished yet (beta :-) ).
I'm working on making the order of columns not depend on the language, this turned out to be more difficult than I expected.

[...]

PR #7080, opened 2025-01-02
new: add sequence number in transactions panel

georgeef:
This PR adds a sequence number in transaction panels.

## Motivation

The transaction ID used to be assigned by SQLite and it was essentially a sequence number, indicating the order of creation. With the introduction of SUID (see #6964), the transaction ID is a long number and its appearance in transaction panels becomes less convenient.

The ID column in transaction panels is useful for sorting (by order of creation), although the exact value of a transaction ID is essentially opaque (i.e., of little interest to the user).

This PR adds one more column in transaction panels, containing a sequence number (SN) for each transaction, which is assigned sequentially (starting with 1) when a transactions list is generated. The order of SN is equivalent to the order of Date/ID, i.e., transactions are sorted by Date and ID, and then an SN is assigned to them sequentially.

From user's perspective, either the SN or the ID can be used for sorting. Users may hide one of the two columns, according to their preference. The SN has the advantage that it is a shorter number, but it may be confusing since it is assigned dynamically. On the other hand, the ID is static, but it is a longer number and may look random and confusing.

## Design

The displayed SN for executions of scheduled transactions in a transactions panel, follows the same rules as the displayed ID. The SN of scheduled transactions is empty, while the SN of scheduled splits appears as ".", in order to distinguish the scheduled transactions from the rest of transactions, when either the SN or the ID column is used.

## Implementation

- In `Model_Checking::Full_Data`: add long `SN`, wxString `displaySN`.
- In `Fused_Transaction`: add struct `SorterByFUSEDTRANSSN`.
- Refine `mmCheckingPanel::filterTable()`. Fix bug with wrong filtering when "Ignore Future Transactions" is set.
- In `TransactionListCtrl::EColumn`: add `COL_SN` at the end of the list (in order to preserve already assigned column numbers).
- Refine `TransactionListCtrl`.
- In `TransactionListCtrl::OnColClick()`, break the association between Date and ID.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7080)

### DIFF
--- a/src/fusedtransaction.h
+++ b/src/fusedtransaction.h
@@ -61,12 +61,14 @@ public:
         explicit Full_Data(const Model_Checking::Data& t);
         Full_Data(const Model_Checking::Data& t,
             const std::map<int64 /* TRANSID */, Split_Data_Set>& splits,
-            const std::map<int64 /* TRANSID */, Taglink_Data_Set>& tags);
+            const std::map<int64 /* TRANSID */, Taglink_Data_Set>& tags
+        );
         Full_Data(const Model_Billsdeposits::Data& r);
         Full_Data(const Model_Billsdeposits::Data& r, wxString date, int repeat_num);
         Full_Data(const Model_Billsdeposits::Data& r, wxString date, int repeat_num,
             const std::map<int64 /* BDID */, Budgetsplit_Data_Set>& budgetsplits,
-            const std::map<int64 /* BDID */, Taglink_Data_Set>& tags);
+            const std::map<int64 /* BDID */, Taglink_Data_Set>& tags
+        );
         ~Full_Data();
 
         int64 m_bdid;
@@ -80,6 +82,15 @@ public:
         bool operator()(const DATA& x, const DATA& y)
         {
             return (!x.m_repeat_num && (y.m_repeat_num || x.TRANSID < y.TRANSID));
+        }
+    };
+
+    struct SorterByFUSEDTRANSSN
+    { 
+        template<class DATA>
+        bool operator()(const DATA& x, const DATA& y)
+        {
+            return (!x.m_repeat_num && (y.m_repeat_num || x.SN < y.SN));
         }
     };
 
--- a/src/mmchecking_list.cpp
+++ b/src/mmchecking_list.cpp
@@ -911,7 +911,8 @@ void TransactionListCtrl::OnColClick(wxListEvent& event)
     m_sortCol = toEColumn(ColumnNr);
     g_sortcol = m_sortCol;
 
-    // If primary sort is DATE then secondary is always ID in the same direction
+    // disabled: If primary sort is DATE then secondary is always ID in the same direction
+    // decouple DATE and ID, since SN may be used instead of ID (see #7080)
     if (false && ColumnNr == COL_DATE) {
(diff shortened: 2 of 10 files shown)


georgeef:
@n-stein 
I added a new column code (COL_SN) at the end of `TransactionListCtrl::EColumn`. I understand that code numbers are important for the saved column settings, so the new code should not interfere with previously assigned codes.

georgeef:
@whalley 
See last bullet in Implementation above. I had to break the association between Date and ID, since users may use SN instead of ID.

giongy:
Hello,
I'm just reading your updates, I'm not a developer but I'm very interested about this software.
Your update will be affect this kind of link?
image
I mean, which number will be showed here? the "fake" SN or the real one?

georgeef:
The ID will be longer (about 16 decimal digits) for new entries in 1.9.0. Only the transaction panels that can open directly from navigation tree at left will show an SN in addition to ID. The ID will still be available and it will be the real ID, as before.

rmelillo76:
I know this still a work in progress but wanted to point out that Transaction Report hyperlinks seem to be broken and are not opening the Edit Transaction dialog, (nothing appears to happen when clicking the link). It still does work for the older, shorter ID's:

image

georgeef:
@rmelillo76 
Thank you for reporting. Could you file a separate issue, including the version and how to reproduce?  Please check first if there is another recent issue for the same problem.

@guanlisheng as FYI

rmelillo76:
> @rmelillo76 Thank you for reporting. Could you file a separate issue, including the version and how to reproduce? Please check first if there is another recent issue for the same problem.
> 
> @guanlisheng as FYI

Thanks, happy to help. Issue https://github.com/moneymanagerex/moneymanagerex/issues/7082 has been opened.

n-stein:
I'm having two issues:

1. SN is displayed in the wrong column:

image

2. The All Transactions page crashes because it is attempting to sort on the "Deleted Time" column which doesn't exist in that panel.

n-stein:
For issue 2 we should verify that `g_sortcol` and `prev_g_sortcol` exist in `m_real_columns` at the beginning of `TransactionListCtrl::sortTable`. If not, they should be reset to `COL_def_sort` and `COL_def_sort2` respectively.

n-stein:
I found issue 1 occurs when the stored column order contains UDFC entries which don't exist in CUSTOMFIELD_V1. This may occur since SETTING_V1 is cross-database -- you can open db1 and the UDFC columns will be stored in the order, but when you open db2 which doesn't have any UDFC fields it still loads them into m_real_columns which throws off the order.

n-stein:
I'll create a new issue for these and fix them.

georgeef:
@n-stein 
Thank you for the observations and for the debugging. I didn't notice in my tests.

n-stein:
#7099 

> I didn't notice in my tests.

To be fair, these bugs are unrelated to your SN change. They are an artifact of the column order save/restore changes made for #6847, so mostly my fault... I simply noticed them when I glanced at the SN column. In hindsight I should have debugged further before commenting in this thread, so my apologies.

georgeef:
No problem, good that you noticed and you know how to fix it.

PR #7095, opened 2025-01-06
add tip for SN in transactions panel

georgeef:
The keyword SN for Sequence Number introduced in #7080, may be confusing or cryptic to users in the beginning (see also the discussion in #6964).

This PR shows the following tip in `m_info_panel` (bottom area of transaction panels), when SN is selected as the primary sorting column: "SN (Sequence Number) has the same order as Date/ID."

Note: A tooltip on the SN header would be more suitable, however wxWidgets does not support it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7095)

### DIFF
--- a/src/mmchecking_list.cpp
+++ b/src/mmchecking_list.cpp
@@ -246,9 +246,10 @@ void TransactionListCtrl::sortTable()
     );
     m_cp->m_header_sortOrder->SetLabelText(sortText);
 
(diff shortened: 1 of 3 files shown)


georgeef:
Added also a tip when ID is selected as the primary sorting column.

Issue #7097, opened 2025-01-06
CSV Import - attempting to load a preset saved for an account with an int64 ID results in error


n-stein:
MMEX version:
 - [x] 1.9.0

Operating System:
 - [x] Windows

If you create a new account (so that it has the new int64 ID) then save a preset for that account in CSV import by checking "Load preset when account is ...", subsequent attempts to open the CSV import dialog for that account will result in an stoi error due to this line:

https://github.com/moneymanagerex/moneymanagerex/blob/47b07a0234b55b23da46585ca3d5589af82b06ef/src/import_export/univcsvdialog.cpp#L225

![image](https://github.com/user-attachments/assets/2ae13ff8-ed68-4829-a7a3-dcf3edfa08d1)

n-stein:
Also same issue here if CATEGID is an int64.

https://github.com/moneymanagerex/moneymanagerex/blob/47b07a0234b55b23da46585ca3d5589af82b06ef/src/mmreportspanel.cpp#L712

https://github.com/moneymanagerex/moneymanagerex/blob/47b07a0234b55b23da46585ca3d5589af82b06ef/src/mmreportspanel.cpp#L719

Issue #7090, opened 2025-01-03
Import causes PAYEE_V1 Exception constraint failed[2067]


itlee23:
MMEX version:
 - [ ] 1.7.0
 - [ ] 1.8.0
 - [x] 1.8.1
 - [ ] 1.8.2
 - [ ] 1.9.0
 - [ ] Other (please specify)

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [x] Windows
 - [ ] Mac OSX
 - [ ] Linux 

# Description of the bug

Please describe your issue with details.
Add screenshot or other files if needed.

On import I use Payee pattern matching, in the Preview we see that it matches the pattern
![image](https://github.com/user-attachments/assets/672a0eeb-47ce-45d6-b629-eabf2b685301)

![image](https://github.com/user-attachments/assets/f8a005fc-3650-4a55-9d30-7f7c8e257a28)

but on import MMEX throws this error
```
23:15:13: PAYEE_V1: Exception constraint failed[2067]: UNIQUE constraint failed: PAYEE_V1.PAYEENAME, {
    "PAYEEID": -1,
    "PAYEENAME": "Tesco PFS 3719, H Hempstead ",
    "CATEGID": -1,
    "NUMBER": "",
    "WEBSITE": "",
    "NOTES": "",
    "ACTIVE": 1,
    "PATTERN": ""
}
```
and the transactions are imported as "Unknown".
![image](https://github.com/user-attachments/assets/91a2f9a7-0976-47c7-a702-2c5868434980)

I understand the error message, but this should really fail silently as the record already exists and that is fine. But this is causing the import to incorrectly assign Unknown as the payee.

Actually the problem is the matching pattern is not working, as it shouldnt be creating a new payee as we have assigned it to an existing one via pattern matching.

If I delete the payee first, the import works without error but still creates the payee instead of using the matched payee, which means the next time I import it throws the error again.

![image](https://github.com/user-attachments/assets/a201bf53-691c-434b-8733-cc74373be68a)

# Reproduction

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:
- import csv
- incoming payee is confirmed as matching pattern
- press import
- dialog indicates 3 records successfully imported
- error dialog pops up

Expected result:
- new transaction should have matched payee "Tesco PFS"
- no new payee

Actual result:
[...]

PR #7098, opened 2025-01-06
fix(#7097), fix(#7090): corrections for CSV import

n-stein:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7098)


n-stein:
Found a couple cases where we were using `atoi` and `stoi` to parse strings representing `int64` ids. Fixed all that I found.

PR #7121, opened 2025-01-13
tidy: unicode translation macro

n-stein:
A small refactor which adds a macro to simplify translation of strings containing unicode characters

`wxGetTranslation(wxString::FromUTF8(wxTRANSLATE(unicode_string)))` can now be written `_u(unicode_string)`

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7121)


georgeef:
This is a good addition.

`_u8` might be a little more descriptive, but `u8` usually means "unsigned 8-bit" (I see that both `u8` and `__u8` are used in wxWidgets), so better to avoid it.

Issue #7150, opened 2025-01-25
Date range in transaction panels


georgeef:
This proposal introduces a new specification for date ranges and applies it to transaction panels. Another proposal will apply a similar specification to reports.

## Motivation

Transaction panels provide a small set of pre-defined date ranges (e.g., Current Month, Last Month, etc.). Only a few of the available choices have a finite extension to the future, which is useful for scheduled executions (introduced by #6860). Many date ranges that might be useful to the users are missing (e.g., from start of current year to end of current week).

Adding more choices is not a satisfactory solution for the following reasons: (1) The number of useful date ranges is too big, since the extension into the future adds one more dimension. (2) A large number of date ranges requires some organization in sub-menus, which makes the user interface inconvenient and slow.

## Proposal

The date ranges offered in transaction panels are configurable by the user, using a dialog window designed for this purpose.

A simple range is specified by:
- int `so1` : start offset
- int `eo1` : end offset
- PERIOD_ID `sp1` : start period type 
- PERIOD_ID `ep1` : end period type

where PERIOD_ID is one of the following:
- `A` (all) 
- `Y` (year)
- `Q` (quarter)
- `M` (month)
- `W` (week)
- `T` (today)
- `S` (statement date)

The start of a simple range is determined by (`so1`, `sp1`):
- (x, `A`) means the first available date (past without limit)
- (0, `W`) means the first day of the current week
- (-1, `W`) means the first day of the previous week (subctact 1*7 days from the date of today, then round to the first day of the week)
- (-7, `T`) means 7 days before today (subctact 7 days from the date of today)

Similarly, the end of a simple range is determined by (`eo1`, `ep1`):
- (x, `A`) means the last available date (future without limit)
- (0, `W`) means the last day of the current week
- (1, `W`) means the last day of the next week (add 1*7 days to the date of today, then round to the last day of the week)
- (7, `T`) means 7 days after today (add 7 days to the date of today)

If a period type is `A`, then the correspoding offset is not applicable (it is set to 0 at writeand it is ignored at read). The statement date `S` is available only in single-account panels and, like `T`, its granularity is day:
- (0, `S`) means the statement date
- (1, `S`) means the next day after the statement date (add 1 day to the statement date)

For more flexibility, a second subranage can optionally be specified:
- int `so2` : start offset of the second subrange
- int `eo2` : end offset of the second subrange
- PERIOD_ID `sp2` : start period type of the second subrange
[...]

PR #7167, opened 2025-02-01
new(#7150), part 4: refine DateRange2 and add CHECKING_RANGE_DEFAULT

georgeef:
This PR depends on #7162.

## Refactor header files

- Move primitive types/functions from `util.{h,cpp}` to `primitive.{h,cpp}`.

- Move `mmTreeItemData` from `util.{h,cpp}` to `mmTreeItemData.{h,cpp}`.

This refactor aims to reduce include cycles. The definitions in `primitive.{h,cpp}` do not depend on any other file in `src/` (they depend on `wx` through `defs.h`). The definitions in `mmTreeItemData.{h,cpp}` depend on `reports/reportbase.h`, which in turn depends on many header files.

## Changes in `DateRange2`

- Add class `Spec`.

- Refactor spec functions to class `Spec`.

## Changes in `Option`

- Add `CHECKING_RANGE_DEFAULT`.

- Add `m_checking_range_a`, `m_checking_range_m`.

- Add `getCheckingRange{A,M}()`.

- Add `parseCheckingRange()` and refine `{load,set}CheckingRange()`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7167)


georgeef:
@whalley 
The AppVeyor build failed. There are a few warnings of the form "declaration hides class member", but this should not be the reason. Do you understand why it failed?

Anyway is it preferable to not hide class members in constructors?

Issue #7229, opened 2025-02-15
Error on using the credit limit feature


MartinX3:
MMEX version:
 - [x] 1.8.1
 - [x] 1.9.0

Operating System:
 - [ ] Windows
 - [ ] Mac OSX
 - [x] Linux 

# Description of the bug

An assertion failed.
`/usr/src/debug/wxwidgets/wxWidgets/src/gtk/gauge.cpp(108): assert ""pos <= m_rangeMax"" failed in SetValue(): invalid value in wxGauge::SetValue()`

# Reproduction

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:
Set a credit limit.
Be in negative balance over the limit.

Expected result:
No error.

Actual result:
Error.

# Additional information

```
ASSERT INFO:
/usr/src/debug/wxwidgets/wxWidgets/src/gtk/gauge.cpp(108): assert ""pos <= m_rangeMax"" failed in SetValue(): invalid value in wxGauge::SetValue()

BACKTRACE:
[1] wxGauge::SetValue(int)
[2] wxEvtHandler::ProcessEventIfMatchesId(wxEventTableEntryBase const&, wxEvtHandler*, wxEvent&)
[3] wxEvtHandler::SearchDynamicEventTable(wxEvent&)
[4] wxEvtHandler::TryHereOnly(wxEvent&)
[5] wxEvtHandler::ProcessEventLocally(wxEvent&)
[6] wxEvtHandler::ProcessEvent(wxEvent&)
[7] wxScrollHelperEvtHandler::ProcessEvent(wxEvent&)
[8] wxGenericTreeCtrl::DoSelectItem(wxTreeItemId const&, bool, bool)
[9] wxGenericTreeCtrl::OnMouse(wxMouseEvent&)
[10] wxEvtHandler::ProcessEventIfMatchesId(wxEventTableEntryBase const&, wxEvtHandler*, wxEvent&)
[11] wxEventHashTable::HandleEvent(wxEvent&, wxEvtHandler*)
[12] wxEvtHandler::TryHereOnly(wxEvent&)
[13] wxEvtHandler::ProcessEventLocally(wxEvent&)
[14] wxEvtHandler::ProcessEvent(wxEvent&)
[15] wxScrollHelperEvtHandler::ProcessEvent(wxEvent&)
[16] wxEvtHandler::SafelyProcessEvent(wxEvent&)
[...]

PR #7230, opened 2025-02-15
fix(#7229): set range max

whalley:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7230)


georgeef:
@whalley 
std::max may fail in windows, due to a macro in windows.h (see comment in #7167).

Also this function has been refactored in #7224, merge will fail but it is easy to fix manually.

whalley:
> @whalley std::max may fail in windows, due to a macro in windows.h (see comment in #7167).
> 
> Also this function has been refactored in #7224, merge will fail but it is easy to fix manually.

Seems to work OK in windows.

PR #7260, opened 2025-02-22
refine mmListCtrl (part 2)

georgeef:
## GUI changes

At right-click on a column header:

- "Hide/Show column" shows the first (previously empty) column as "Icon".

- "Move column left/right" appears only if applicable.

- Added "Move hidden column", which opens a menu of hidden columns. The selected column is added after the current column. This item is shown only if there are hidden columns.

Use case: users can move a column over multiple positions as follows

- At the original column: right-click -> "Hide this column"

- At (before) the final position: right-click -> "Move hidden column" -> select column.

Other changes

- The width of a column is saved when the column is hidden, and it is restored when the column is shown again (it was reset to default before). However, the current settings format encodes the hidden state as zero width, therefore the width preferences of hidden columns is not stored in settings (to be fixed in another PR.)

## Architectural changes

Column variables indicate either a column number (0-based physical position, like 1 for the second physical column), or a column id (the code of a logical column, like LISTCOL_ID_DATE). To avoid inconsistencies, all column variables/functions have a suffix (nr, id) in their name.

Derived classes may have 0 (budgetingListCtrl), 1 (all other), or 2 (TransactionListCtrl) sorting columns. The state variables `m_col_nr_id`, `m_sort_asc` are defined as vectors in order to cover all cases.

Sorting functions need both the column number and the column id. However, maintainance of column number is tedious due to column moves. The state variable `m_col_nr_id` stores the column ids, while the corresponding column numbers are cached into `c_sort_col_nr`.

## Changes in mmListCtrl

- Rename and refactor `m_columns` -> `m_col_id_info`. Change the index from column number to column id. `m_col_id_info` is configured by the constructor and it is not updated at every column move. This simplifies the code in several functions.

- Rename `m_column_order` -> `m_col_nr_id`. The new name indicates that this is a map from column number to column id.

- Add `m_col_id_width` (lazy variable), `m_col_id_hidden`. This demultiplexes the hidden state from the width preference, when a column is hidden.

- Rename and refactor `m_selected_col` (single column number) -> `m_sort_col_id` (vector of column ids). This simplifies the maintainance of the sorting variables when columns are moved.

- Rename and refactor `m_asc` -> `m_sort_asc` (vector).

- Rename `m_ColumnHeaderNbr` -> `m_col_nr`.

- Remove `m_selected_row` (moved to derived classes), `m_default_sort_column` (not needed).

- Move all access to user settings into `savePreferences()` and `loadPreferences()`. This is also a preparation to refine the settings format.

- Add several helper functions.

### DIFF
--- a/src/assetspanel.cpp
+++ b/src/assetspanel.cpp
@@ -28,55 +28,48 @@
 #include <wx/srchctrl.h>
 
 wxBEGIN_EVENT_TABLE(mmAssetsListCtrl, mmListCtrl)
-    EVT_LIST_ITEM_ACTIVATED(wxID_ANY,   mmAssetsListCtrl::OnListItemActivated)
-    EVT_LIST_ITEM_SELECTED(wxID_ANY,    mmAssetsListCtrl::OnListItemSelected)
-    EVT_LIST_END_LABEL_EDIT(wxID_ANY,   mmAssetsListCtrl::OnEndLabelEdit)
     EVT_RIGHT_DOWN(mmAssetsListCtrl::OnMouseRightClick)
     EVT_LEFT_DOWN(mmAssetsListCtrl::OnListLeftClick)
 
-    EVT_MENU(MENU_TREEPOPUP_NEW,    mmAssetsListCtrl::OnNewAsset)
-    EVT_MENU(MENU_TREEPOPUP_EDIT,   mmAssetsListCtrl::OnEditAsset)
-    EVT_MENU(MENU_TREEPOPUP_ADDTRANS, mmAssetsListCtrl::OnAddAssetTrans)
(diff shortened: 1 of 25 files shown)


georgeef:
@whalley 
Better keep this on hold, until the Crowdin translations are restored.

whalley:
> "Move column left/right" appears only if applicable.

This does not work if the leftmost or rightmost columns are hidden. E.g. I have the ICON hidden but am still presented with the menu option

georgeef:
Yes, the check for spare hidden columns at the end would add more code for little value.

Do you think it is important?

whalley:
> Yes, the check for spare hidden columns at the end would add more code for little value.
> 
> Do you think it is important?

No, not an important issue for me.

PR #7365, opened 2025-03-22
Asset: unify initial val and market val calculation

guanlisheng:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7365)


guanlisheng:
fix #7365

Issue #6771, opened 2024-05-24
Avoiding duplicate entries on CSV Import


renato-mmex:
# Description of the feature

See: https://forum.moneymanagerex.org/viewtopic.php?t=11108

For example, I would import the csv files on the 15th of the month, but I would always need to set the date to overlap the date of my last import, so that the transactions included in my new download would go back to the 8th or even earlier.

The reason for this is because we often have transactions that come through late by certain vendors. I.e., I may have made a purchase on the 9th but the transaction will not appear until the 12th,or it will be listed as “pending” and is therefor not included when I export my transactions. We have this kind of thing happen regularly on all our credit card accounts. The same is true if I import on the 15th, but some purchases I made on the 12th may not be posted until the 18th or later, so I would have to catch those next time.

So, in order to catch everything I always overlap the dates of the imports, and in doing so I will catch the late postings but unfortunately also import others that have already been imported previously.

It would therefor be helpful if I could weed out those duplicates either at the time of import or else after wards. But I can’t seem to figure out how to do that with MMEX. 

Thanks for your help.

@n-stein I would support such a feature

jensb:
The way iFinance does this (on Mac) is to have a "IMPORT_IDENTIFIER" column for transactions which consists of date + "#" + amount. It seems this combination is "unique enough" to detect duplicates. 
MMEX could do something similar - maybe additionally add the payee to be really sure.

PR #7368, opened 2025-03-23
Feature: allow skipping or flagging duplicate transactions (QIF imports)

oboxodo:
This is my first attempt to contribute to this project. C++ is not my main language so I'm very open to feedback.

I tend to import QIF/CSV files from my bank a few times per month and I can't really use the date filtering to avoid duplicates because sometimes my bank would take a few days to include some transactions. Pretty much the same issue as described in https://github.com/moneymanagerex/moneymanagerex/issues/6771.

So I decided to try and add a few opt-in filtering features. Default behavior is exactly the same as before.

image

The options are:
* Duplicates:
  - By transaction number
  - By amount and exact date
  - By amount and nearby date
* Action:
  - Skip
  - Flag as duplicate

The options are pretty much self-explanatory but, just in case, the "By amount and nearby date" option will try to match the incoming transaction with any still-unmatched transaction having the exact same amount but in a time window determined from 4 days before the incoming date and 2 days after. These values might need some fine-tuning but I think they're good defaults.

I'm open to implement similar features for the CSV importing, but I'd like to get feedback first and iterate over this feature in any case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7368)

### DIFF
--- a/src/import_export/qif_import_gui.cpp
+++ b/src/import_export/qif_import_gui.cpp
@@ -181,27 +181,34 @@ void mmQIFImportDialog::CreateControls()
     //Filtering Details --------------------------------------------
     wxStaticBox* static_box = new wxStaticBox(this, wxID_ANY, _t("Filtering Details:"));
     wxStaticBoxSizer* filter_sizer = new wxStaticBoxSizer(static_box, wxVERTICAL);
-    wxFlexGridSizer* flex_sizer2 = new wxFlexGridSizer(0, 2, 0, 0);
-    filter_sizer->Add(flex_sizer2, g_flagsExpand);
+
+    // Create a horizontal sizer to hold the date and duplicate columns
+    wxBoxSizer* filter_grid_sizer = new wxBoxSizer(wxHORIZONTAL);
+    filter_sizer->Add(filter_grid_sizer, g_flagsExpand);
+
+    // Left column for dates
+    wxFlexGridSizer* dates_sizer = new wxFlexGridSizer(0, 2, 0, 0);
 
     // From Date
     dateFromCheckBox_ = new wxCheckBox(static_box, wxID_FILE8, _t("From Date")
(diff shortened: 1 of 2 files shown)


georgeef:
I didn't look into the implementation. I have a few comments on the functionality.

The selection between matching criteria is a good idea. You can also add "By all fields", which is the most strict.

The selection of the action is also a good idea.

The probability for two different transactions to have the same amount (and same or similar date) is quite high.
For loose matching criteria (like the last two), it is better to always add the potential matches as duplicates (and later the user can cleanup), rather than to skip them (which has the risk that they are not noticed by the user and they are lost). This means if a loose matching is selected, then only the "Flag as duplicate" action should be offered.

It is good to think about the possible use cases (and later to document them in the user manual). For example:
- Transactions are imported from bank data and are not modified after the import. Importing "By all fields" with "Skip", is almost safe and doesn't need any post-processing, but it has the risk that two different transactions may look identical. More safe is to import with "Flag as duplicate", at the cost of additional post-processing.
- The user has a way to edit the Number field, and devises a custom scheme to identify transactions using this field (maybe in combination with date, month, etc.). Then importing "By transaction number" with "Skip" can be safe, as long as the user assigns a unique Number (or combination) to each transaction.
- Transactions are imported from bank data and they may be modified after import. But it is rare that the same transactions are imported again. Loose matches (amount plus date) with "Flag as duplicate" offer a heuristic which may work well this use case, and it is safe at the cost of small post-processing.

AKJUS:
Akjus updates this new comments..

whalley:
@oboxodo Diego, thanks for the contribution and apologies for not processing this earlier. I have assigned https://github.com/moneymanagerex/moneymanagerex/issues/6771, for which this PR satisfies, to v1.9.2. I want to get v1.9.1 out ASAP (it is long overdue) so don't want any more feature changes at present.

PR #7388, opened 2025-03-28
Checking panel: fix date sorter

guanlisheng:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7388)

### DIFF
--- a/src/mmchecking_list.cpp
+++ b/src/mmchecking_list.cpp
@@ -443,7 +443,10 @@ void TransactionListCtrl::sortTransactions(int col_id, bool ascend)
         sortBy(SorterByNOTES(), ascend);
         break;
     case TransactionListCtrl::LIST_ID_DATE:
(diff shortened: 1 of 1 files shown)


georgeef (commented):
Sort by `LIST_ID_DATE` considers only the date part in TRANSDATE, because this is what is shown under column Date. If Time is enabled, the time part is shown in another column. Users expect to sort what they see, so when they sort by Date they expect a sort of the date part only.

This fix would generate confusing results in the following use case:
- Time is enabled  (both Date and Time columns appear)
- Users want to sort my Date/Payee (i.e., they want to see transactions sorted by payee within the same date; time is not relevant).

The generated list after this fix will be a sort by Date/Time/Payee, which is different than Date/Payee, and most probably not what the users want.

Time is merged with Date in TRANSDATE (probably to avoid a DB change when this change was introduced), from users point of view, Date and Time are two different fields and they are shown in two different columns.

PR #7456, opened 2025-04-23
make REFTYPE_CHOICES common

### LINKED ISSUE #1628: Model_Attachment::REFTYPE_CHOICES
Actually we have all main entities in Model_Attachment::REFTYPE_CHOICES vector and they are used to identify reference type for generic functions (such as attachments, customfields, assets).
I think they should be moved outside Model_Attachment, maybe in a dedicated .h file or in generic Model.h, what do you think?

guanlisheng:
to fix #1628 
- #1628 

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7456)

### DIFF
--- a/src/accountdialog.cpp
+++ b/src/accountdialog.cpp
@@ -422,7 +422,7 @@ void mmNewAcctDialog::OnCurrency(wxCommandEvent& /*event*/)
 
 void mmNewAcctDialog::OnAttachments(wxCommandEvent& /*event*/)
 {
-    wxString RefType = Model_Account::refName();
+    wxString RefType = Model_Account::refTypeName;
     mmAttachmentDialog dlg(this, RefType, m_account->ACCOUNTID);
     dlg.ShowModal();
 }
--- a/src/accountdialog.cpp
+++ b/src/accountdialog.cpp
@@ -33,7 +33,6 @@
 #include "model/Model_Infotable.h"
 #include "model/Model_Currency.h"
 #include "model/Model_Stock.h"
-#include "model/Model_Attachment.h"
 
 #include <wx/valnum.h>
 
@@ -423,7 +422,7 @@ void mmNewAcctDialog::OnCurrency(wxCommandEvent& /*event*/)
 
 void mmNewAcctDialog::OnAttachments(wxCommandEvent& /*event*/)
 {
-    wxString RefType = Model_Attachment::REFTYPE_NAME_BANKACCOUNT;
+    wxString RefType = Model_Account::REFTYPE_NAME_BANKACCOUNT;
     mmAttachmentDialog dlg(this, RefType, m_account->ACCOUNTID);
     dlg.ShowModal();
 }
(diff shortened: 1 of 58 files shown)


georgeef:
I don't understand the logic of this change. Looking into the source code, it seems that the choices are re-defined in each Model. I think it is better to define them only once, outside of any specific model.

For example
> Model_Account::REFTYPE_NAME_BANKACCOUNT
sould be something like
> Model_general::REFTYPE_NAME_BANKACCOUNT
or just
> REFTYPE_NAME_BANKACCOUNT

Another comment: `REFTYPE_*` is used in many places and currently it is prefixed by `Model_Attachment::`. All these instances should get a new prefix. If we keep the old prefix in most cases, what is the need to change it only for `Model_Account`?

georgeef (commented):
My review: see previous comment.

gabriele-v:
> I don't understand the logic of this change. Looking into the source code, it seems that the choices are re-defined in each Model. I think it is better to define them only once, outside of any specific model.
> 
> For example
> 
> > Model_Account::REFTYPE_NAME_BANKACCOUNT
> > sould be something like
> > Model_general::REFTYPE_NAME_BANKACCOUNT
> > or just
> > REFTYPE_NAME_BANKACCOUNT
> 
> Another comment: `REFTYPE_*` is used in many places and currently it is prefixed by `Model_Attachment::`. All these instances should get a new prefix. If we keep the old prefix in most cases, what is the need to change it only for `Model_Account`?

I agree with above, issue was opened because I think it's misleading having them in Model_Attachment while they are used for example for tags as well. They should be moved in a more generic model and then changed everywhere.

guanlisheng:
It's defined in the base class `Model` and is available for each derived model (instead of re-defined)

Model_Accout is just an example to show how to use it.

georgeef:
Ok, I see, but if `REFTYPE_*` can have multiple `Model_*` prefixes, this defeats the purpose of moving it out of `Model_Attachment`.  Even after a cleanup, in the future there can be instances of `REFTYPE_*` which compile, but with a prefix which doesn't make much sense (like in the source code of this commit, in which one instance has `Model_Account` and all other `Model_Attachment`).

georgeef (commented):
In `REFTYPE_ID_size` the last word was on purpose in small letters. This is in order to (1) distinguish meta-cases (small letters) from normal cases (capital letters), and (2) to avoid confusion in the rare case that "SIZE", "COUNT" or another meta-name happens to be a normal name. Several other enum lists are similar. See for example enum `PERIOD_ID` in `src/daterange2.h`, where all meta-cases have small letters.

Note: "size" and "count" are good words for the meta-case that follows all normal cases, because its value is equal to the number of normal cases. "MAX" is confusing, because the value of the last normal case is one smaller that the value of "_MAX" meta-case. I have changed several "_MAX" to "_size" in the past ("_count" is also ok). In other programs I have seen also "_sup" (from the mathematical "supreme", here it means "above all cases"), which is better than "_max", but not so common.

gabriele-v:
Cannot test it at the moment, but code side is what I was expecting, thanks!

gabriele-v (approved):

Issue #7433, opened 2025-04-17
C++ standard upgrade


guanlisheng:
https://github.com/AnthonyCalandra/modern-cpp-features

# C++ Standard Version Support by Compiler

This table lists the minimum version of each major compiler that supports various C++ standards.

| C++ Standard | GCC     | Clang   | MSVC     |
|--------------|---------|---------|----------|
| C++11        | 4.8+    | 3.3+    | 2013+    |
| C++14        | 5.0+    | 3.4+    | 2015+    |
| C++17        | 7.0+    | 5.0+    | 2017+    |
| C++20        | 10.0+   | 11.0+   | 2019+    |
| C++23        | 13.0+   | 16.0+   | 2022+    |

## Notes

- For best results, use the **latest compiler version** available.
- Partial support may exist in earlier versions but could be incomplete or buggy.
- MSVC versions refer to Visual Studio releases (e.g., MSVC 2019 = Visual Studio 2019).

guanlisheng:
How about C++20?  We have all compilers available and ready .


@whalley @vomikan @georgeef @n-stein and @renato-mmex

guanlisheng:
Just so you know, we are one C++20 now.  The next step is #7441

n-stein:
C++20 causes fmt to throw compile warnings which have been fixed in more recent versions. Bumping fmt to the latest release (11.1.4) resolves them.

guanlisheng:
Thank you, @n-stein.

PMaff:
```
# --------- Set compiler parameters ---------
set(CMAKE_CXX_STANDARD 20)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```

in CMakeLists.txt kicks out older systems like Ubuntu 18.04. MMEX is not compileable there anymore!
:-(

Commenting this out in CMakeLists.txt lets MMEX configure
```
-- MMEX configuration summary
-- ==========================
-- Version        : 1.9.1-Beta.1
-- Commit         : 35f152115 (2025-05-02)
-- Branch         : master
-- Host system    : Linux x86_64
-- Target system  : Linux x86_64
-- Build types    : 
-- Generator      : Unix Makefiles
-- Install prefix : /usr/local
-- DB encryption  : ON
-- 
-- Versions
-- --========--
-- Linux 5.4.0-215-generic
-- Ubuntu.18.04.bionic
-- Ubuntu.18.04.6.LTS
-- CMake 3.30.2
[...]

PR #7458, opened 2025-04-24
fix(#7433): update fmt submodule for C++20

n-stein:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7458)


guanlisheng:
hi @n-stein , wondering whether we can use `std::fmt` as c++20 upgrade done?

@georgeef , @whalley and @vomikan as FYI

Issue #7468, opened 2025-04-29
Last 1.9.1 Beta doesn't run on Linux


froux07:
MMEX version:
 - [ ] 1.7.0
 - [ ] 1.8.0
 - [ ] 1.8.1
 - [ ] 1.8.2
 - [ ] 1.9.0
 - [x] 1.9.1 Beta

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [ ] Windows
 - [ ] Mac OSX
 - [x] Linux : LinuxMint Xia 22.1 (based on Ubuntu Noble) + Ubuntu Noble

# Description of the bug

Maybe due the CPP20 integration but the last mmex 1.9.1 Beta doesn't run.
**_Deb version:_** 
/home/runner/work/moneymanagerex/moneymanagerex/src/choices.cpp(33): assert "id >= 0 && id < static_cast<int>(array.size())" failed in getName().
Erreur de segmentation (core dumped)

**_Flatpak :_** 
/run/build/moneymanagerex/src/choices.cpp(33): assert "id >= 0 && id < static_cast<int>(array.size())" failed in getName().
/usr/include/c++/14.2.0/bits/stl_vector.h:1149: constexpr std::vector<_Tp, _Alloc>::const_reference std::vector<_Tp, _Alloc>::operator[](size_type) const [with _Tp = ChoicesName::Item; _Alloc = std::allocator<ChoicesName::Item>; const_reference = const ChoicesName::Item&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

latest working version : mmex-2025-04-23-12a02bf

guanlisheng:
Thanks @froux07, can you share the detailed version info and build time?


It is similar to 
- #7406

froux07:
Hi @guanlisheng 
It's not similar, because in my case, MMEX does not launch.
These are not mmex built by myself.
I have no more logs than I gave above.
Whether with flatpak or deb, I run mmex in a terminal, and I get these errors.
**_These are mmex downloaded from pipelines:_**
Flatpak: https://app.circleci.com/pipelines/github/moneymanagerex/moneymanagerex
Deb: https://github.com/moneymanagerex/moneymanagerex/actions/workflows/ubuntu-ci.yml

guanlisheng:
[...]

PR #7475, opened 2025-05-02
Cleanups, labeling unused variables

gerstrong:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7475)


src/choices.cpp
```diff
@@ -28,8 +28,12 @@ ChoicesName::~ChoicesName()
 {
 }
 
-const wxString ChoicesName::getName(int id) const
+const wxString ChoicesName::getName(const int id)
 {
+    // Small hack helping to avoid crashes of inline definitions that make program crash here
+    if(id >= static_cast<int>(array.size()))
+        array.resize(id+1);
+
```
guanlisheng, line 36: Seems to fix 
- #7468 

But I don't know why. Do you have more insight?

@georgeef

georgeef:
In general I don't like this resize. It forces a vector access at an index that is known to be out of range.

Did you manage to debug in linux?  To me it seems to be either a bug linux compiler, or an undefined behavior according to cpp standard, which is implemented differently in linux.

I cannot debug, so I can only try to guess. One possible explanation is wrong optimization of `size()` for a const vector. `array` is declared as `const` with the idea that it is fixed (it never changes) after initialization. Maybe the compiler thinks that `array.size()` is known to be zero, which is true before the initialization of `ChoicesName` (but not after, because in all variable declarations `array` is non-empty). Does it work if you remove the `const`, but without the `resize()`?

Another possibility is that `array` is indeed empty. It doesn't seem to be a logical error in the program, because it has different behavior in other platforms. Can you check what are the actual values of `id` and `array.size()` when the assertion fails?

gerstrong:
Yes, it is a quick fix, also not so happy about it. I built this on Fedora Linux 41 (g++ 14.2). I can look deeper into it, but I fear we need to redesign some more files around it.

The array is empty, because the constructor ChoicesName::ChoicesName is called way later. There are static variables that are loaded before main() is executed and require that "array" already set up. I think ChoicesName::getName(const int id) should be called later.

georgeef:
I am not sure whether it is a program bug or a compiler bug/behavior. The fact that it appears only in linux hints towards the latter. `getName()` is anyway called after the initialization of a `ChoicesName` variable. It seems that the compiler transposes the initialization order, so when it calls `getName` the variable is still uninitialized (or perhaps it is initialized in parallel).

@gerstrong 
If you are in linux, could you check the behavior of C++17?

I don't know if we should analyze more and fix the source code (i.e., if the problem is due to undefiled behavior), or if we should revert back to C++17 (in case it is a compiler bug in the linux compiler).

georgeef:
Do you know which `ChoicesName` variable gives an assertion error?

I searched for `PERIOD_CHOICES`, which is easy to find. It appears only in `src/model/Model_Budget.{h,cpp}`. `getName` is called inside the inline function `Model_Budget::period_name`, which appears also in `src/budgetingpanel.cpp`, `src/budgetentrydialog.cpp`.

Could you remove the `inline` and check again?  Better do it for the variable that gives an assertion error.

gerstrong:
Bildschirmfoto_20250503_134830

gerstrong:
The compiler returns with a lot of errors like:  " in-class initialization of static data member ‘const wxString Model<DB_Table_USAGE_V1>::REFTYPE_NAME_TRANSACTIONSPLIT’ of non-literal type"

georgeef:
Ok, it is `REFTYPE_CHOICES`. Thank you for the debugging. I now see that `REFTYPE_CHOICES` is defined in `src/model/Model.h` as `inline static ChoicesName`, which is a mistake.

You need to create `src/model/Model.cpp` and to move the initialization there.

@guanlisheng
`REFTYPE_CHOICES` was moved in a recent commit in order to share it between different models. Does this fit to the last working commit reported in #7468?

guanlisheng:
The behavior on Linux/GCC is weird, according to C++17.

> an inline static data member can be defined in the class definition and may specify an initializer. It does not need an out-of-class definition:

@gerstrong, can you continue to verify whether moving the initializer into `Model.cpp` works?

georgeef:
According to this specification `REFTYPE_CHOICES` can be defined and initialized as inline (there is no syntax error), however it is not clear if the compiler creates one copy or multiple (and independent) copies each time `Model.h` is included.

In any case it is more clear to have a single copy of `REFTYPE_CHOICES` as static variable in class Model, which is initialized only once, even if this won't fix the assertion error.

PMaff:
```
# --------- Set compiler parameters ---------
set(CMAKE_CXX_STANDARD 20)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```

in CMakeLists.txt kicks out older systems like Ubuntu 18.04. MMEX is not compileable there anymore!
:-(

georgeef:
Making `REFTYPE_CHOICES` a function maybe makes its declaration less ugly, but it also makes its usage more ugly because of the extra `()`. I would prefer an ugly implementation than an ugly usage. Anyway if this fixes the problem it is at least a temporary solution.

@guanlisheng
Can you try to move the initialization to `Model.cpp`?  If it compiles and works on macOS I guess it will also work on linux.

georgeef:
Maybe it is better to move `REFTYPE_CHOICES` and all related declarations into class `ModelBase`, which is not a template. The initialization can be moved in the implementation of `ModelBase` in a new file `Model.cpp`.

georgeef:
I have sent a fix in #7478.

georgeef:
I tried to initialize `REFTYPE_NAME_*` at compile time using constexpr, but it gives an error (wxString is not a literal type).

@gerstrong's solution (make `REFTYPE_CHOICES` a function) may work. But it is not very clean and still some care is needed to avoid dependencies between static variables in different files.

Another solution (more clean, although less convenient) is to remove all `Model_*::refTypeName` variables and to use `REFTYPE_NAME_*` instead (as we had in the past, except that `REFTYPE_CHOICES` and `REFTYPE_NAME_*` can be moved to class `ModelBase`.

georgeef:
I reopened #7478. Moves all initializations of `Model_*::refTypeName` to `Model.cpp`. Tested on macOS, the initializations look ok.

@gerstrong 
Could you test on linux?

PR #7495, opened 2025-05-18
Some cleanups from past changes and making emplace work with wxvector that is not wrapped by stl vector

gerstrong:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7495)

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -1315,7 +1315,7 @@ CURLcode getYahooFinanceQuotes(const wxString& URL, wxString& output) {
 
     struct curl_slist* headers = nullptr;
     headers = curl_slist_append(headers, "Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8");
-    headers = curl_slist_append(headers, "User-Agent: Mozilla/5.0 (Windows NT 10.0; rv:109.0) Gecko/20100101 Firefox/117.0");
+    headers = curl_slist_append(headers, "user-agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.0.0 Safari/537.36");
     if (!savedCookie.IsEmpty()) headers = curl_slist_append(headers, static_cast<const char*>(("Cookie: " + savedCookie).mb_str()));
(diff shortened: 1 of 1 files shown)


gerstrong:
hi @guanlisheng 
yes, I read that sentences as well, but I don't think it applies everywhere. In my case it did not translate to stl vector. I am using Fedora 41 and version 3.2.6 of wxwidgets

guanlisheng:
wired. 
How do you compile your wxWidgets 3.2.6 here?

gerstrong:
I don't compile it. I am just using the fedora packages:

`
dnf list wx*
Updating and loading repositories:
Repositories loaded.
Installed packages
wxBase.x86_64          3.2.6-1.fc41        updates
wxBase-devel.x86_64    3.2.6-1.fc41        updates
wxGTK.x86_64           3.2.6-1.fc41        updates
wxGTK-devel.x86_64     3.2.6-1.fc41        updates
wxGTK-gl.x86_64        3.2.6-1.fc41        updates
wxGTK-i18n.noarch      3.2.6-1.fc41        updates
wxGTK-media.x86_64     3.2.6-1.fc41        updates
wxGTK-webview.x86_64   3.2.6-1.fc41        updates
wxsqlite3.x86_64       4.10.6-1.fc41       updates
wxsqlite3-devel.x86_64 4.10.6-1.fc41       updates
`

guanlisheng:
Got it. 
By default. wxWidgets v3.2 is compiled with STL disabled.  Is it possible to build your wxWidgets instead?

gerstrong:
It surely is possible, but why should I do that? It needs to work with standard packages

guanlisheng:
It is still standard para from the wxWidgets POV., and MMEX source code depends on it.

Here is the spec for wxWidgets 3.2.6 on fc41. We can see that `--enable-std_containers` or (`--enable-stl`) is missing.
https://src.fedoraproject.org/rpms/wxGTK/blob/f41/f/wxGTK.spec

gerstrong:
Yes, and we should not break compatibility to distro packages. I also don't like my changes, but they fix that issue.

guanlisheng:
No worry.  

> starting from version 3.3.0, these custom container classes are implemented in terms of the standard library classes by default

Issue #7558, opened 2025-07-22
Feature: Show stock list instead of help page in Stock portfolios


KWich:
If "Stock portfolios" is selected in the navigator, the help page is displayed, which is not very helpful here.  With this feature a stock summary over all portfolios is shown instead, which gives a quick overview.

### Existing view:
 
<img width="1224" height="512" alt="Image" src="https://github.com/user-attachments/assets/b1273c14-1f58-48b5-ba15-d2814d758c74" />

### New view:

<img width="1410" height="591" alt="Image" src="https://github.com/user-attachments/assets/fc317cce-e9b7-498d-b66d-90ed6ed0ffab" />

(Code will be provided)

guanlisheng:
hi @KWich, 

Would we show the Investment Account Summary list here instead of their stocks?  very close to the stock portion on the homepage.

KWich:
> Would we show the Investment Account Summary list here instead of their stocks?

I do not know if I understand this correctly. The existing reports (Stocks Summary Report)  and list in the dashboard either show an summary per portfolio or a list of stocks per portfolio.  There is no overview yet, over all stocks independent of there  portfolio. So if I own the same stock in several portfolios (which I in fact do for tax reasons), I will not see the overall total. This is what I changed with the proposed  new overview, which gives a quick overview independent of a the actual portfolios, and the means to directly perform updates, which Is not possible within a report.

guanlisheng:
your understanding is correct.  

I also got your use case now and it is to track the stock level tax by joining them with name from different portfolios.  it is a trick.  

so far, there is no good way to assoicate them while we can track it well within the sample portfolio(investment account). A better way is to assoicate them.

here are the typical hierarchy, and `MyStock2.B` in multi portfolios is your case.
```
Portfolios/
├── MyPortfolio1
│   ├── MyStock1.A
│   └── MyStock1.B
└── MyPortfolio2
    ├── MyStock2.B
    └── MyStock2.C
```

PR #7565, opened 2025-07-23
Feature#7558: Show stock list instead of help page in Stock portfolios

KWich:
Code for Feature #7558: Show stock list instead of help page in Stock portfolios
+ changes after copilot review
See issue description for details.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7565)

### DIFF
--- a/po/de_DE.po
+++ b/po/de_DE.po
@@ -8186,10 +8186,14 @@ msgstr "Los"
 msgid "Stock Portfolio: %s"
 msgstr "Aktien Portfolio: %s"
 
+#: stockspanel.cpp:409
+msgid "Stock Portfolios Overview"
+msgstr "Wertpapierübersicht"
+
 #: stockspanel.cpp:417
 #, c-format
 msgid "Cash Balance: %s"
-msgstr "Zahlungsbilanz: %s"
+msgstr "Kontostand: %s"
 
 #: stockspanel.cpp:418
 #, c-format
@@ -8859,4 +8863,3 @@ msgstr "Fehler "
 #: wizard_update.cpp:325 wizard_update.cpp:340
 msgid "MMEX Update Check"
 msgstr "MMEX Update Überprüfung"
-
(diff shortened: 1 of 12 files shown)


guanlisheng:
Thanks @KWich for making this.  

A help page might be good for the current stock hierarchy. 

btw, can you replace raw query with orm?

KWich:
> btw, can you replace raw query with orm?

What do you mean with "orm" ??

guanlisheng:
Sorry for the late reply.  

ORM means to use the Model_xxx (and DB_Table_xxx) to access database instead of your raw sql here.

KWich:
> can you replace raw query with orm?

I looked at the model and table implementation. From my understanding it is not possible to perform extended queries with the existing implementation like it is done with the "raw SQL". It would be a major effort to replace the SQL queries with an own logic, and do not see any advantage of doing it. From my understanding all queries performed directly in the SQL database are faster than first retrieving the data and processing them in the code level, and also cause less errors, because the SQL can be tested offline before.

guanlisheng:
It would be fast from an execution point of view, while keeping the logic consistent would be the first ask here.  It is one of the best practices here.  

Can you review the current homepage implementation and reconsider if any outstanding issues?

- https://github.com/moneymanagerex/moneymanagerex/blob/master/src/mmhomepage.cpp
- https://github.com/moneymanagerex/moneymanagerex/blob/master/src/model/Model_Account.cpp#L237
- https://github.com/moneymanagerex/moneymanagerex/blob/master/src/model/Model_Stock.cpp#L81

whalley:
@KWich Klaus, what is the status of this now? I don't actively use the stocks/shares capability so can't really test this in anger.  Is it worth merging for wider testing?

KWich:
I made one last change to polish the stock dialog handling.  From my point of view the pull request can now be merged, I'm not planning any more changes. I hope that @guanlisheng  will agree, too.

whalley:
Great. I saw the SQL changes, which look fine now. Will merge.

Issue #5124, opened 2022-09-23
Update Windoes wxWebView to use wxWEBVIEW_BACKEND_EDGE backend


whalley:
In order to support modern web interfaces the Windows backend should be updated as IE is no longer maintained and does not support latest HTML.

https://docs.wxwidgets.org/3.2/classwx_web_view.html

The following limitation may add complexity.

https://docs.wxwidgets.org/3.2/classwx_web_view.html
wxWEBVIEW_BACKEND_EDGE (MSW)
The Edge backend uses Microsoft's [Edge WebView2](https://docs.microsoft.com/en-us/microsoft-edge/hosting/webview2). It is available for Windows 7 and newer.

**This backend does not support custom schemes and virtual file systems.**.

_Originally posted by @whalley in https://github.com/moneymanagerex/moneymanagerex/issues/4545#issuecomment-1256006448_

whalley:
Dark Mode help (https://github.com/moneymanagerex/moneymanagerex/issues/4545) should work once this is implemented.

n-stein:
[745296d](https://github.com/n-stein/moneymanagerex/commit/745296d73c98f8ed836863eaca472df4eb6344ab) adds Edge compatibility, but only with wxWidgets 3.3.0 (current development branch). I'm still working on getting the VFS assets to load using Edge in the 3.2.x builds.

n-stein:
For those interested, a test version of MMEX built with wxWidgets 3.3 using the Edge backend can be found in builds on my `depends-wx3.3` branch. I try to keep this generally in sync with the current master branch, but it shouldn't be used for general beta testing as there may be UI bugs specific to the new wxWidgets version which may not affect the current flavor of wxWidgets 3.2.x we are using.

See for example: https://ci.appveyor.com/project/n-stein/moneymanagerex/builds/49155143

rmelillo76:
> For those interested, a test version of MMEX built with wxWidgets 3.3 using the Edge backend can be found in builds on my `depends-wx3.3` branch. I try to keep this generally in sync with the current master branch, but it shouldn't be used for general beta testing as there may be UI bugs specific to the new wxWidgets version which may not affect the current flavor of wxWidgets 3.2.x we are using.
> 
> See for example: https://ci.appveyor.com/project/n-stein/moneymanagerex/builds/49155143

Hey @n-stein  - this is very cool, wanted to give it a try but this build still seems to use the IE renderer. Any ideas?

[...]

PR #7617, opened 2025-08-13
fix(#5124): use Edge backend for WebView in MSW

n-stein:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7617)


n-stein:
I think the code analysis errors can be ignored. It is complaining about missing headers, but they clearly exist as it compiles fine...

Issue #7577, opened 2025-07-30
Feature: Unified date range handling


KWich:
In the current implementation date selection is combined with the transaction attribute filtering in one control. Either one of the predefined date ranges can be selected or a new filter request request has to be created to  display a specific date range. In addition all filters including the selected date are stored per account. 

<img width="583" height="261" alt="Image" src="https://github.com/user-attachments/assets/6f50cfa0-d5d6-4b49-a01e-a6e96e617949" />

This implementation has a few drawbacks:
- While switching between accounts for comparison, the same filter has to be selected again for each account.
-  A quick extension of a the selected range is not possible, first a a new range filter has be configured.

# Description of the feature
Several new functions are added:
## New date controls
Separate the date selection from the general transaction filters. A date range is selected with a new set of controls with represent a set of predefined ranges, but also allow to define or extend the currently selected 
date range on the fly.

<img width="1000" height="229" alt="Image" src="https://github.com/user-attachments/assets/25900637-b1f5-4ba3-8b2f-5c0b8269e51e" />

The same date setting is general, which means it is applied to all accounts and reports where feasible.
This allows flexibility in the date selection, a selected predefined range can be extended by changing the date in the date pickers.

## Date range editing
In addition an edit function is provided, which allows to change the predefined date ranges in the control:
- changing the sequence
- update a existing entry
- create new entries
- delete unused entries

<img width="852" height="931" alt="Image" src="https://github.com/user-attachments/assets/08ec1204-cffe-47cb-b229-2389e75da559" />

## Switch between existing and new handling
The feature is implemented additionally to the existing handling. With a new settings option the behavior can be switched back to the old behavior:

<img width="591" height="462" alt="Image" src="https://github.com/user-attachments/assets/ef722da9-d50d-4c55-b16f-0e1fafd6d99d" />

(Code will be provided)

whalley:
Sounds good but not sure we need the complexity of adding 'Enable combined transaction filter' option.

KWich:
> Sounds good but not sure we need the complexity of adding 'Enable combined transaction filter' option.

It's easy to remove the option and to purge the "old" code.  The questions is if all users are happy  with the change, the option provides more flexibility and allows a choice for the moment.

renato-mmex:
I would like to test these and other corrections, but https://github.com/moneymanagerex/moneymanagerex/issues/7567 is not fixed.

whalley:
[...]

PR #7655, opened 2025-09-02
Feature#7577 unified data update 2

KWich:
Like discussed the change for #7577:
- remove the old date range handling
- add a switch to store the the date range either individually per account/report or globally

In addition three more related changes:
- I noticed that the report id is always set to -1 during filter storage, so there was no individual filter storage. 
- User manual update for the settings description. Is not complete but reflects the current structure
- German translation of the user manual update

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7655)

### DIFF
--- a/src/mmreportspanel.cpp
+++ b/src/mmreportspanel.cpp
@@ -89,12 +89,12 @@ bool mmReportsPanel::Create(wxWindow *parent, wxWindowID winid
     SetExtraStyle(GetExtraStyle() | wxWS_EX_BLOCK_EVENTS);
     wxPanel::Create(parent, winid, pos, size, style, name);
 
-    m_use_dedicated_filter = !Option::instance().getUseCombinedTransactionFilter();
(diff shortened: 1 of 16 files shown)


src/mmreportspanel.cpp
```diff
@@ -120,52 +120,16 @@ bool mmReportsPanel::saveReportText(bool initial)
...
-                    wxDateTime end_date = m_end_date->GetValue();
-                    date_range->start_date(begin_date);
-                    date_range->end_date(end_date);
-                    m_start_date->Enable();
-                    m_end_date->Enable();
-                }
-                rb_->date_range(date_range, selectedItem);
-            }
-
-            if (rp & (rb_->RepParams::BUDGET_DATES | rb_->RepParams::ONLY_YEARS))
```
whalley, line 157: @KWich What happens with these date options?

PR #7793, opened 2025-10-13
Fixed sorting of investment accounts in stocks widget

i5an-i5ano5:
The investment accounts are not sorted alphabetically by account name in the Stocks widget. They are shown in the order they were created, because the sorting function is not called.
stocks

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7793)


whalley:
@i5an-i5ano5 Thanks for the fix.

PR #7818, opened 2025-10-24
Bugfix and transaction dialog enhancements

KWich:
This pull request consists of multiple parts:

1. Fixes to remove compiler warnings about unused vars

1. Fixes in the GRM editor:
     - ask for unsaved changes before a context menu is opened to prevent data loss
     - align font size of fixed font, which was to large

1. Fix a problem with the calculator popup and refactor: 
       - If a value was provided in the text filed and the calculator is opened, the value was deleted as soon as a calculator button was clicked, because the value in the calculator was selected.
       - refactored code
       
1. Small refactoring in the category dialog: 
     - Use the existing event table for all events instead a combination of event table and connect calls

1. Enhancement: Adapt transaction dialog
   - Additional accelerator key(s): "Ctrl+S" for Save and "Ctrl+N" for Save and New
   - New button to exchange to and from
   - New button to set date to today
   - Refactored control sequence
   - Hide unused controls depending on state
   
Transaction dialog 2

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7818)

PR #7829, opened 2025-10-27
Derive accelerator key from (translated) tooltip text (Open issue from  #7818)

KWich:
One last change for the accelerator key discussion in #7818 with @ovari :

To accommodate different languages and layouts, the used accelerator key for the reset today is retrieved from the tool tip text.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7829)

### DIFF
--- a/3rd/CMakeLists.txt
+++ b/3rd/CMakeLists.txt
@@ -225,7 +225,7 @@ if(NOT wxSQLite3_FOUND OR NOT (WXSQLITE3_HAVE_CODEC OR MMEX_ENCRYPTION_OPTIONAL)
     endif()
 
     if(NOT MSVC) # wxSQLite3 security ext doesn't require C 99 for MSVC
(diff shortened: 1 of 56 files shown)


ghost:
Thank you for implementing.

If a computer is set to a locale that is unknown in MMEX. Is the keyboard shortcut, and the one shown in the tooltip, the English source string?

Thank you once again.

KWich:
> If a computer is set to a locale that is unknown in MMEX. Is the keyboard shortcut, and the one shown in the tooltip, the English source string?

Yes.
It is not controlled by the locale but by the po file translation. If the string is not translated in the corresponding po file, it defaults to the default, which is us-English: (`"Set date to today (Ctrl+;)")) `  If the string is translated in the po file without defining a CTRL Key sequence (or an invalid one), the accelerator key is set to `Ctrl+.`, which should match most keyboards.

Issue #7710, opened 2025-09-16
v1.9.1 Crash on opening when it auto attempts to import WebApp transactions


tccadam:
MMEX version:
 - [ ] 1.7.0
 - [ ] 1.8.0
 - [ ] 1.8.1
 - [ ] 1.8.2
 - [ ] 1.9.0
 - [x] 1.9.1
 - [ ] Other (please specify)

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [x] Windows
 - [ ] Mac OSX
 - [ ] Linux 

# Description of the bug

With the new version 1.9.1 it crashes when it gets to the import webapp transactions before it lists them.  I've been able to get around this for the time being by using the portable version of 1.9.0. open the old version, import the webapp transactions. close old version, then open v1.9.1 and it gets past the import webapp transactions without crashing.  I'm assuming because there are no transactions left to be listed and downloaded.

# Reproduction

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:

Expected result:

Actual result:

# Additional information

If you have a backtrace for a crash or a warning, paste it here.

whalley:
Confirmed. I can reproduce.

whalley:
@gerstrong  Have reverted much of this as it was causing the issue - https://github.com/moneymanagerex/moneymanagerex/commit/34c4a4b2d9cd7656685e759008c6d5a7c7722ed8.

Just left the STL change for the vector.

whalley:
@tccadam Can you please confirm if this build fixes the issue. https://ci.appveyor.com/project/moneymanagerex/moneymanagerex/builds/52757296/job/n7hockt7wdwgnow9/artifacts

tccadam:
@whalley YES! mmex-1.9.2-Beta.1-win64-portable works fine, no more crashing on webapp importing. Thanks!

sharkyiwc:
@whalley same issue, it's work with mmex-1.9.2-Beta.1-win64-portable
Thanks

PR #7926, opened 2025-12-07
STL Vector to WxVector conversion added.

gerstrong:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7926)

### DIFF
--- a/po/mmex.pot
+++ b/po/mmex.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: MMEX\n"
 "Report-Msgid-Bugs-To: developer@moneymanagerex.org\n"
-"POT-Creation-Date: 2025-12-02 11:07+0000\n"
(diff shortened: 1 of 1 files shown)


whalley:
@gerstrong This change was previously reverted... https://github.com/moneymanagerex/moneymanagerex/issues/7710 as it caused a crash.

gerstrong:
Oh, I see, maybe it makes sense to detect which version of wx (or flavor) is used and with #define set case 1 or 2. I don't understand why Fedora 42 is that strange dealing with wxVector wxWidgets 3.x.

Let me have a second look and try to make this an exception. Without my change, I cannot compile MMEX on my system. Maybe you have an idea, what needs to be changed. Maybe something with the CMake setup?

whalley:
I've really no idea on this.

gerstrong:
Okay, I added another detection for such cases. Let me know if it improves the situation.

PR #7932, opened 2025-12-09
Update Backup Documentation (EN & IT) with Correct Backup Behavior

frankieorabona:
This PR updates the Backup documentation for both English and Italian languages.

Changes include:

- Corrected the explanation of when backups are created:
  * Backups are generated only when the database has been modified
  * Both startup and exit backups follow this rule
- Documented the correct daily backup behavior:
  * Only one _start and one _update backup file per day
  * MMEX never overwrites or updates existing .bak files
- Added detailed explanation of backup filename structure
- Improved the restore procedure to avoid filename conflicts
- Clarified historical snapshot behavior with scheduled transactions
- Updated sticky notes, tips, and safety recommendations
- Preserved existing document structure, layout, images, and formatting

These changes ensure the documentation accurately reflects MMEX’s real backup behavior and help users understand how to safely manage and restore backups.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7932)

### DIFF
--- a/docs/en_US/index.html
+++ b/docs/en_US/index.html
@@ -511,155 +511,200 @@
 
 	<h3>Database Backup</h3>
 
-
-	<p>
-    MMEX generates backup copies (<code>.bak</code> files) to protect your data from errors,
-    crashes, corruption, or incompatibilities caused by newer database versions.
-	</p>
-
-	<img src="../en_US/images/create_db_07.png" alt="Database backup options" class="shadow" />
-	<!-- The image should show Tools → Options → Others with Backup options visible. -->
-
-	<h4>Backup Settings</h4>
-
-	<p>
-		Backup settings are available at:
-	</p>
-
-	<p>
-		<kbd><samp>Tools</samp> → <samp>Settings</samp> → <samp>Other</samp></kbd>
-	</p>
-
-	<p>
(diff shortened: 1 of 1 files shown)


frankieorabona:
Just to add a few notes about this update.

The goal of the PR is simply to correct the Backup section so that it reflects how MMEX really behaves today.
I kept the original layout, images, notes and formatting, and replaced only the part that needed an update.

Main points now clarified:
-backups are created only if the database has been modified
-both startup and exit backups follow this rule
-on the same day MMEX creates at most one _start and one _update file
-existing .bak files are never overwritten or updated
-only the main .mmb file changes when multiple edits are made in the same day
-the restore procedure is now clearer and avoids filename conflicts
-the part about scheduled transactions and historical snapshots has been expanded and corrected

If anything needs to be adjusted or refined, I’m available to update the PR.

As context, since I'm currently working on the documentation in both English and Italian, I updated the Backup section in both languages in a consistent way.
The goal is to avoid having mismatched or outdated descriptions across translations, especially for parts of the program where the behavior has changed over time.

Both versions now follow the same structure and explain the same concepts, so future updates can stay aligned more easily.

Thanks for checking it.

PR #7940, opened 2025-12-12
New Features: Navigator expand/collapse all popup menu  -  Reconcile dialog

KWich:
This pull request contains two new features:

### Expand/Collapse all popup 
 A new popup menu for the navigator which allows to expand and collapse the whole tree. the popup is invoked with a right mouse click either in the empty navigator space or on the main item.

CollapePopup

### A  reconcile dialog, which can be invoked for each account:
reconciledialog

See online manual for details.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7940)

### DIFF
--- a/src/generic/genericFocusBitmapButton.h
+++ b/src/generic/genericFocusBitmapButton.h
@@ -18,7 +18,7 @@
 
 #pragma once
 
-#include <wx/wx.h>
+#include "defs.h"
 
 class genFocusBitmapButton : public wxBitmapButton
 {
--- a/src/generic/genericFocusButton.h
+++ b/src/generic/genericFocusButton.h
@@ -18,7 +18,7 @@
 
 #pragma once
 
-#include <wx/wx.h>
+#include "defs.h"
 
 class genFocusButton : public wxButton
 {
--- a/src/uicontrols/reconciledialog.h
+++ b/src/uicontrols/reconciledialog.h
@@ -85,9 +85,6 @@ private:
     void OnRightFocusKill(wxFocusEvent& event);
     void handleListFocusKill(wxListCtrl* list);
 
-    //void OnButtonFocus(wxFocusEvent& event);
-    //void OnButtonFocusKill(wxFocusEvent& event);
-
     void setListItemData(const Model_Checking::Data* trx, wxListCtrl* list, long item);
     void processRightClick(wxListCtrl* list, long item);
     void processLeftClick(wxListCtrl* list, wxPoint pt);


frankieorabona:
wow! The reconcile dialog is a really nice addition: having a dedicated UI with clear entry points, transaction selection, difference calculation and the Reconcile / Reconcile later workflow makes the reconciliation process much more intuitive and user-friendly.

Great work, this is a very useful feature!

po/mmex.pot
```diff
@@ -7926,10 +7975,12 @@ msgid "Cash Flow - Transactions"
 msgstr ""
 
 #: reports/categexp.cpp:206
+# mmframe.cpp:1716
 msgid "Collapse All"
 msgstr ""
 
 #: reports/categexp.cpp:207
+# mmframe.cpp:1715
 msgid "Expand All"
```
ovari: Please add a keyboard accelerator, say `&Expand All`. Thank you

ghost:
In the reconcile window, if a transaction is selected:
1. If the transaction is unreconciled, can pressing the `Spacebar` toggle the transaction to reconciled?
2. If the transaction is reconciled, can pressing the `Spacebar` toggle the transaction to unreconciled?
3. Once the transaction status is updated, automatically move to the next transaction, unless at the end of the transaction list.

Thank you

KWich:
> * If the transaction is unreconciled, can pressing the `Spacebar` toggle the transaction to reconciled?
> 
> * If the transaction is reconciled, can pressing the `Spacebar` toggle the transaction to unreconciled?
> 
Space bar handling is implemented

> * Once the transaction status is updated, automatically move to the next transaction, unless at the end of the transaction list.

From my point of view this behavior is not desirable, since already the arrow keys can be used to move to the next (or the previous) item. I extended the dialog in such a way, that in addition to the up and down arrows, the left and right arrows will switch between the lists (see provided documentation)

ghost:
> > * Once the transaction status is updated, automatically move to the next transaction, unless at the end of the transaction list.
> 
> From my point of view this behavior is not desirable, since already the arrow keys can be used to move to the next (or the previous) item. I extended the dialog in such a way, that in addition to the up and down arrows, the left and right arrows will switch between the lists (see provided documentation)

The move automatically to the next transaction after the `Spacebar` pressed is released to toggle a reconcile/unreconcile a transaction is a feature that is implemented in Quicken and saves time when reconciling statements.

rmelillo76:
The Import and Export menus are displayed twice:

image

rmelillo76:
Hi, @KWich great work here! Just a few suggestions to consider, would welcome other's opinions as well:
1. Add the Number column in addition to Date, Payee and Amount. I can imagine many users would be reconciling uncleared checks so that could be a useful field to have.
2. Widen or autofit the Date column. When using MM/DD/YYYY format it is truncated with "..."
3. Change the headers from "Payments / Transfers outgoing" to simply "Withdrawals" and "Deposits / Transfers outgoing" to simply "Deposits" to match the terms used in the main checking panel.
4. Use alternating row colors like the checking panel grids.
5. Edit and New buttons are a different size than Toggle All, make them consistent
6. Open the dialog centered over the main window.

KWich:
> * Add the Number column in addition to Date, Payee and Amount. I can imagine many users would be reconciling uncleared checks so that could be a useful field to have.

Done
 
> * Widen or autofit the Date column. When using MM/DD/YYYY format it is truncated with "..."

Set to autofit, should work now.

> * Change the headers from "Payments / Transfers outgoing" to simply "Withdrawals" and "Deposits / Transfers outgoing" to simply "Deposits" to match the terms used in the main checking panel.

Done

> * Use alternating row colors like the checking panel grids.

This has to implemented as a new feature

 
> * Edit and New buttons are a different size than Toggle All, make them consistent

Done
 
> * Open the dialog centered over the main window.

 In my setup (Linux) this is the case.

KWich:
> The Import and Export menus are displayed twice:

I fixed it

rmelillo76:
Hi @KWich - while using this some more I noticed that 3 (very old) transactions from 2020 and 2021 that were in a **Void** status were not showing in the new Reconcile dialog, but others were. 

This also got me thinking that perhaps Void transactions shouldn't be included because (at least in my case), these were checks I started to issue but made an error, but I like to keep them in my register for historical purposes.

I opened my .mmb in DB Browser to see if I could determine why those were not included and I observed that the DELETEDTIME for those 3 records was NULL (not a zero-length string). And in fact, most of my records prior to sometime in 2023 are NULL, but of course most are reconciled. 

image





#### I'd like thoughts on this:

1. Should Void status be included, or just Unreconciled and Follow-up? To me Void and Duplicate, although un-reconciled may be confusing to include for this reason. When clicking Reconcile later changing them to Follow-up (in my case) also affected my running and ending balance because of the way that MMEX normally treats Void transactions that have a non 0.00 amount.  
2. Do we need a DB Script to update all unreconciled items DELETEDTIME from NULL to '', or can this code be modified to handle both?  (Note, if you edit and save those older records, the field does get updated to a zero-length string). 

https://github.com/moneymanagerex/moneymanagerex/blob/9343366c05e0e9e40ceefdae809ac96f89f15f13/src/uicontrols/reconciledialog.cpp#L248


As always, thanks for your great and innovative work here.

KWich:
> * Should Void status be included, or just Unreconciled and Follow-up? To me Void and Duplicate, although un-reconciled may be confusing to include for this reason. When clicking Reconcile later changing them to Follow-up (in my case) also affected my running and ending balance because of the way that MMEX normally treats Void transactions that have a non 0.00 amount.

In the way I use it I like to see all transactions which are not reconciled in the dialog. But hat's mainly because I do not use the void or duplicate states. I agree that this may be confusing, since the reconcile dialog does not indicate the void and duplicated states.  Possible improvements could be to add the state indicator for  void and duplicated to the dialog and/or to provide a setting, which allows to restrict the selection to unreconciled and followup. 

> * Do we need a DB Script to update all unreconciled items DELETEDTIME from NULL to '', or can this code be modified to handle both?  (Note, if you edit and save those older records, the field does get updated to a zero-length string).

How do you get transactions with a DELETEDTIME of NULL in the first place? I looked through the code and my understanding is that that the DELETEDTIME is always supposed to be an empty string, if the transaction is valid  (at least this is only checking criteria). If NULL should be considered, this would affect other places in the code too, since in most case there are only checks for an empty string. 
So if NULL was a valid option before it would be good idea to update NULL to Empty string during a database update.

rmelillo76:
>In the way I use it I like to see all transactions which are not reconciled in the dialog. But hat's mainly because I do not use the void or duplicate states. I agree that this may be confusing, since the reconcile dialog does not indicate the void and duplicated states. Possible improvements could be to add the state indicator for void and duplicated to the dialog and/or to provide a setting, which allows to restrict the selection to unreconciled and followup.

I think for simplicity adding the current state either as a columnn, or even just applying the icons that are in the main checking panel would give enough of a visual indicator, e.g.:

image



> How do you get transactions with a DELETEDTIME of NULL in the first place? I looked through the code and my understanding is that that the DELETEDTIME is always supposed to be an empty string, if the transaction is valid (at least this is only checking criteria). If NULL should be considered, this would affect other places in the code too, since in most case there are only checks for an empty string.
> So if NULL was a valid option before it would be good idea to update NULL to Empty string during a database update.

It was added back in March 2023 so transactions older than that are NULL. So with the next release I agree it would be good to add something along the lines of
`UPDATE CHECKINGACCOUNT_V1 SET DELETEDTIME = '' WHERE DELETEDTIME IS NULL;`

https://github.com/moneymanagerex/database/blob/5a77461d932541cbcd10f2c776dd33398b335964/incremental_upgrade/database_version_16.sql#L5

Spoker:
> > * Should Void status be included, or just Unreconciled and Follow-up? To me Void and Duplicate, although un-reconciled may be confusing to include for this reason. When clicking Reconcile later changing them to Follow-up (in my case) also affected my running and ending balance because of the way that MMEX normally treats Void transactions that have a non 0.00 amount.
> 
> In the way I use it I like to see all transactions which are not reconciled in the dialog. But hat's mainly because I do not use the void or duplicate states. I agree that this may be confusing, since the reconcile dialog does not indicate the void and duplicated states. Possible improvements could be to add the state indicator for void and duplicated to the dialog and/or to provide a setting, which allows to restrict the selection to unreconciled and followup.

I tried the 1.9.2 RC and very much like the reconcile window. I however do make use (a lot) of void transactions and the inclusion of these transactions in the window makes it unusable. I would love to have a way to only include unreconciled or follow-up transaction. I would also love the inclusion of the indicator as shown bij @rmelillo76 . Thanks you for your great work already!

KWich:
I see that a few more options are preferable. 
For the status indication my idea is to add an additional column which indicates the current status with the letters. Due to the limitations of wxWidgets, only one column with symbols is possible, and I do not want to mix the current status with the new selection indication for reconcile.  

In addition there will be options in the settings panel, which control if the current status and/or number columns are displayed, since not everybody needs them.
 
Another option in the settings would control, if the void and duplicated transactions are included or not, or should there be separate switches?

Spoker:
Maybe a checkbox in the reconcile panel "include void transactions" as a direct way to toggle the behaviour?

KWich:
See #8004 for the proposed changes.

PR #7962, opened 2025-12-21
Added the option to export stocks and added the option to limit the exported transactions/stocks between 2 dates (issue#7810)

Argyro-Mpanioti:
I changed the export menus so that they include 2 more checkboxes and 2 more date pickers. The first checkbox replaces the transactions with the stocks when it is checked. The second checkbox and the date pickers make it so that when the checkbox is checked the only transactions/stocks that are exported are those whose dates are between the dates that have been picked by the user. When the second checkbox isn't checked all the transactions/stocks appear regardless of date and the date pickers are disabled. I saw the fields that appear in the stock portfolio and added fields with the same names in the list of fields that the user can add in the exported report. I also altered the encoding text and button so that they look good on both the importing menu and the exporting menu.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/7962)


whalley:
@Argyro-Mpanioti Build fails

Argyro-Mpanioti:
I fixed the issue, the build is successful now.

whalley:
Thanks for contribution. Going to leave this until after v1.9.2 is released as need to stabilise the current build before release laster this month.

Argyro-Mpanioti:
Great! Thanks for letting me know.

PR #8096, opened 2026-01-27
Fix dashboard Income vs Expenses ignoring selected date range

### LINKED ISSUE #8087: Dashboard does not show Last N days
MMEX version:
 - [ ] 1.7.0
 - [ ] 1.8.0
 - [ ] 1.8.1
 - [ ] 1.8.2
 - [ ] 1.9.0
 - [ ] 1.9.1
 - [x] 1.9.2
 - [ ] Other (please specify)

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [x] Windows
 - [ ] Mac OSX
 - [ ] Linux 

# Description of the bug

When you use the last N days option it doesn't work properly, forcing you to switch to the last xx days option

# Reproduction

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:

Expected result:
 show N days chosen

Actual result:
Show  1 Year Ago to Date

Image

# Additional information

If you have a backtrace for a crash or a warning, paste it here.

frankieorabona:
This PR fixes an issue where the Income vs Expenses widget on the dashboard
did not respect the selected date range, especially “Last N days”.

The root cause was that OptionSettingsHome cached the mmDateRange at construction
time, so changes to the selected period were saved but not reflected on the
dashboard until application restart.

The date range is now rebuilt dynamically in
OptionSettingsHome::get_inc_vs_exp_date_range(), ensuring the dashboard always
uses the current settings.

Fixes #8087

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/8096)

### DIFF
--- a/src/optionsettingshome.cpp
+++ b/src/optionsettingshome.cpp
@@ -43,7 +43,7 @@ OptionSettingsHome::OptionSettingsHome()
     m_all_date_ranges.push_back(wxSharedPtr<mmDateRange>(new mmLastFinancialYear()));
     m_all_date_ranges.push_back(wxSharedPtr<mmDateRange>(new mmAllTime()));
     m_all_date_ranges.push_back(wxSharedPtr<mmDateRange>(new mmLast365Days()));
-    m_all_date_ranges.push_back(wxSharedPtr<mmDateRange>(new mmLastNDays(Model_Infotable::instance().getInt("HOMEPAGE_INCEXP_DAYS", 14))));
+    m_all_date_ranges.push_back(wxSharedPtr<mmDateRange>(new mmLastNDays(0)));
 
     int sel_id = Option::instance().getHomePageIncExpRange();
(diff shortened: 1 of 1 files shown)


whalley:
@frankieorabona Having looked at this it has more to do with the fact that the "Last N Days" range is index 16 into the table rather than 15. Changing to 16 fixes the issue and enables the "1 year ago to date" option which is currently masked. However having a hardcoded index reference is not ideal so I might look to recognise the need for the 'N days' another way - perhaps an attribute of mmDateRange.

frankieorabona:
My patch is intentionally minimal, just to restore the correct behavior without changing the existing structure.

I agree that a hardcoded index isn’t ideal — it would be better to identify “N days” explicitly (via a dedicated type/attribute).
Happy to help with a follow-up refactor if needed.

PR #8104, opened 2026-01-31
refactor includes

georgeef:
This PR normalizes the include directives in source files under `src/`.

Symbols:
* `$SRC_DIR`: the directory of the source file; equal to `src` or a directory under `src/` (e.g., `src/model`)
* `$SRC_FILE`: the source file name, without a directory path.
* `$DST_DIR`: the directory of the target file; it can be under `src/` (e.g., `src/model`), or out of `src/` (e.g., `3rd`).
* `$DST_FILE`: the target file name, without a directory path.

Normalization:
```
if $DST_DIR is out of src/:
  #include <$DST_SUBDIR/$DST_FILE>
  example:
    <wx/log.h> is found in (wxWidgets)/include/
    <wx/wxsqlite3.h> is found in 3rd/wxsqlite3/include/
else if $DST_DIR is equal to $SRC_DIR:
  #include "$DST_FILE"
  example in source file src/mmhomepage.cpp:
    "constants.h" is found in src/
  example in source file src/model/allmodels.h:
    "Model_Account.h" is found in src/model/
else if $DST_DIR is equal to src/
  #include "$DST_FILE"
  example in source file src/model/Model.h
    "choices.h" is found in src/
else if $DST_DIR is equal to "src/$DST_SUBDIR"
  #include "$DST_SUBDIR/$DST_FILE"
  example in source file src/accountdialog.h:
    "model/Model_Account.h" is found in src/model/
  example in source file src/reports/cashflow.cpp:
    "model/Model_Account.h" is found in src/model/
```

Changes:
* add `util/check_includes`; this utility shows warnings and recommends a fix in include directives
* in src/CMakeLists.txt: remove `model`, `db` from `target_include_directories`
* fix warnings from the utility above
* fix `util/sqlite2cpp.py`

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/8104)


georgeef:
@whalley 
This PR compiles and runs on macOS, also the Ubuntu checks are ok. However Codacy Static Code Analysis fails. Could you check why this happens? Are the include paths in 3rd/ configured as system paths in Codacy?

It is strange that there were similar includes before, which did not fail. E.g., `src/util.h` in master contains
```
#include <rapidjson/document.h>
```

But the same include fails in `src/wizard_update.h` or in `src/constants.cpp` in this PR.

whalley:
> @whalley This PR compiles and runs on macOS, also the Ubuntu checks are ok. However Codacy Static Code Analysis fails. Could you check why this happens? Are the include paths in 3rd/ configured as system paths in Codacy?
> 
> It is strange that there were similar includes before, which did not fail. E.g., `src/util.h` in master contains
> 
> ```
> #include <rapidjson/document.h>
> ```
> 
> But the same include fails in `src/wizard_update.h` or in `src/constants.cpp` in this PR.

@guanlisheng Think you are the admin for this. Can you please check.

PR #8143, opened 2026-02-10
rename files and classes (part 1)

georgeef:
This PR organizes files in subdirectories and renames several files, as proposed in #7093.

The complete list of changes is stored in `util/filenames.log` for future reference, and also copied below. Developers can consult this file to find the previous filenames.

```
  choices                            -> util/choices
  dateday                            -> util/DateDay
  daterange2                         -> util/DateRange2
  primitive                          -> util/primitive
  reports/mmDateRange                -> util/DateRange
  singleton                          -> util/singleton
  util                               -> util/util

  model/Model                        -> model/_Model
  model/Model_Account                -> model/AccountModel
  model/Model_Asset                  -> model/AssetModel
  model/Model_Attachment             -> model/AttachmentModel
  model/Model_Billsdeposits          -> model/ScheduledModel
  model/Model_Budget                 -> model/BudgetModel
  model/Model_Budgetsplittransaction -> model/ScheduledSplitModel
  model/Model_Budgetyear             -> model/BudgetPeriodModel
  model/Model_Category               -> model/CategoryModel
  model/Model_Checking               -> model/TransactionModel
  model/Model_Currency               -> model/CurrencyModel
  model/Model_CurrencyHistory        -> model/CurrencyHistoryModel
  model/Model_CustomField            -> model/FieldModel
  model/Model_CustomFieldData        -> model/FieldValueModel
  model/Model_Infotable              -> model/InfotableModel
  model/Model_Payee                  -> model/PayeeModel
  model/Model_Report                 -> model/ReportModel
  model/Model_Setting                -> model/SettingModel
  model/Model_Shareinfo              -> model/TransactionShareModel
  model/Model_Splittransaction       -> model/TransactionSplitModel
  model/Model_Stock                  -> model/StockModel
  model/Model_StockHistory           -> model/StockHistoryModel
  model/Model_Tag                    -> model/TagModel
  model/Model_Taglink                -> model/TagLinkModel
  model/Model_Translink              -> model/TransactionLinkModel
  model/Model_Usage                  -> model/UsageModel
  model/allmodel.h                   -> model/_all.h
  model/readme.txt                   -> model/_readme.txt
  option                             -> model/PreferencesModel

  assetspanel                        -> panel/AssetPanel
  billsdepositspanel                 -> panel/ScheduledPanel
  budgetingpanel                     -> panel/BudgetPanel
  mmchecking_list                    -> panel/JournalList
  mmcheckingpanel                    -> panel/JournalPanel
[...]

PR #8147, opened 2026-02-12
rename files and classes (part 2)

georgeef:
This PR is a continuation of #8143. The complete list of changes is stored in `util/filenames.log` for future reference.

Naming conventions
* Where it is possible, files have the same name as the main class inside them.
* The common suffices are: Model, Panel, Manager, Dialog, Preferences, Report.
* Names ending in one of the common suffices are organized in folders and do not have a prefix.
* Files and classes under `util/` but do not have a common suffix. To avoid collisions with other libraries they have prefix `mm`.
* Files containing a collection of definitions, but without a main class, are lowercase or start with `_`.
* Base files in model, panel, preferences start with `_`, in order to be listed separately from other files.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/8147)


georgeef:
@whalley
This is the second and last massive file and class renaming, which affects other developers. I plan to refactor `util/sqlite2cpp.py` and to rename also the files under `db/`, but this will not interfere with the work of other developers.

Most files under `src/` are now organized in a folder. Only a few spare files remain at the top level of `src/`.

PR #8163, opened 2026-02-15
refactor and refine db tables

georgeef:
### Changes
* refactor and refine `util/sqlite2cpp.py`
* refine `util/sqliteupgrade2cpp.py`
* move generated files from `db/` to `table/`
* refactor `table/_TableBase` out of `util/sqlite2cpp.py`
* rename generated patch files from `*.mmdbg` to `patch_*.sql`
* move implementation code from .h to .cpp files
* simplify `*Table::Data` to make it primitive (not depending on `*Table`)
* lower (move) `ModelBase::db_` to `TableBase::m_db` and simplify table functions
* move `SorterBy*` inside `*Table` and add types
* rename `DB_Column` -> `TableOpV` and interchange op and value for better readability

### More changes
* refactor `*Col`, `*Row` outside of `*Table`
* create `table/_TableFactory.tpp`; this is standard practice for template classes in C++
* move shared logic from `util/sqlite2cpp.py` to `TableFactory<>`; the generated files mostly contain lists of names and conversion functions
* create type alias `Data` = `Row` in `Model<>`; `Data` will be refined in the future

### Naming conventions
* the names under `table/` are based on database terminology (column, row, table)
* these names are not application-friendly, but they should not be used directly in the GUI (at the moment they serve as a reminder that these functions are low-level)
* in the future `Data` will be an application-friendly structure for `Row`, and `Model` will handle the conversions between `Data` and `Row`; GUI interacts with `Data` and `Model`; only `Model` interacts with `Table`

### Next steps
* refactor `Model<>`
* gradually refine `*Data` with stronger types and more application-friendly fields

@guanlisheng @whalley as FYI

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/8163)


whalley:
@georgeef I think the change from get() to cache_id() and cache_key() introduces complexity from a programming/readability perspective. The caching should be hidden from those actually writing application code and left to the Model/DB code.

georgeef:
I'll change it. The name "get" is too much overloaded, I need to think of a better name.

Do not merge yet anyway. This commit crashes the application under some conditions.

georgeef:
Status:
* fixed the crash issue (it was overloading of "save()" in TransactionModel - overloading creates problems in general).
* renamed `get_id` -> `get_record` (this was the previous name; not a nice name, but it is used in very few lines).
* renamed `cache_id` -> `get_id`.
* renamed `cache_key` -> `get_key`.

I did more tests, seems to be stable now.

georgeef:
@whalley 
I found one more bug after renaming: `remove()` is also overloaded and creates infinite loops, like with `save()`. I fix it in the next PR that I prepare.

Issue #8191, opened 2026-03-02
[regression] Show buttons in the user interface language


ovari:
# Description of the bug

Buttons are shown in English when the language is set to Hungarian (magyar).

<img width="739" height="406" alt="Image" src="https://github.com/user-attachments/assets/7038f392-ea91-4e7f-9bcf-0d596ec6eb4d" />

# Reproduction

Is the bug reproducible? Always

## Reproduction steps

1. Change the user interface language to Hungarian.
2. Restart MMEX.
3. Click the **New Account** button.

## Expected result

The buttons are shown in the set user interface language.

## Actual result

The currencies are shown incorrectly in English.

Thank you

```
Money Manager Ex
• Verzió: 1.9.3-Beta.1 64 bites
• Készítési időpont: márc. 2 2026, 03:16:30
• Adatbázis verziója: 21 (aes256cbc-hmac-sha512)
• Git-végrehajtás: 284d96df3 (2026-03-02)
• Git-ág: master
MMEX a következő támogatási termékeket használja:
• wxWidgets 3.3.1 (wxGTK 3.24)
• wxSQLite3 4.9.1 (SQLite 3.40.0)
• RapidJSON 1.1.0
• Lua 5.3.5
• libcurl/8.15.0 OpenSSL/3.3.5 zlib/1.3.1 libidn2/2.3.8 libpsl/0.21.5 nghttp2/1.66.0
• gettext 0.22.5
• apexcharts.js
Összeállítás a következőkkel:
• CMake 3.31.5
• GNU Make 4.4.1
• GCC 14.2.0
Fut a következőn:
• GNOME.47 (Flatpak runtime)
• GNOME 47 (Flatpak runtime) ""
• Linux 6.12.73+deb13-amd64 x86_64
• X-CINNAMON (X11)
• Rendszer területi beállítása: en_AU.UTF-8
• Felhasználói felület nyelve: hu_HU.UTF-8
• 1280x720 24-bit 96x96ppi
```

ghost:
1. Right-click on an account.
2. Click **Delete account** option.
3. **No** and **Yes** should show in Hungarian.

<img width="230" height="124" alt="Image" src="https://github.com/user-attachments/assets/8ece1c85-9e71-42a4-bc60-c6db6d932769" />

ghost:
@georgeef, can you please fix? Thank you

georgeef:
This is automatic as long as there are translations for the button labels.

[...]

PR #8195, opened 2026-03-02
fix(#8191): package wxWidgets translations

georgeef:
### Changes
* create `po_wxstd/`
* copy wxWidgets locale/*.po files into `po_wxstd/`
* `po_wxstd/CMakeLists.txt` packages the compiled wxstd.mo files together with the mmex.mo files

### Other changes
* in the build directory, create subdirectory `grm/` and move .grm files there

### Notes

wxWidgets translations are processed in a separate source and build directory (`po_wxstd`) in order to avoid conflicts with MMEX translation files. wxWidgets translations can be disabled by commenting out the `add_subdirectory(po_wxstd)` command in `CMakeLists.txt`.

The MMEX translation files are of the form `<LANG>_<COUNTRY>.po`, while the wxWidgets translation files are mostly of the form `<LANG>.po` (with some exceptions). A mapping from MMEX basename to wxWidsgets basename is defined in `po_wxstd/CMakeLists.txt` using commands of the form `set(WXSTD_MAP_<MMEX_BASENAME> "<WXSTD_BASENAME>")`.

Most languages available in MMEX are also available in wxWidgets, but a few are missing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/8195)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -472,6 +472,7 @@ endif()
 ## Subdirectories with targets ##
 add_subdirectory(3rd)
 add_subdirectory(po)
+add_subdirectory(po_wxstd)
 add_subdirectory(src)
 
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -507,6 +507,7 @@ ENDFOREACH()
 add_custom_target(generate_theme_files ALL DEPENDS ${THEMEFILES})
 
 # Generate Reports
+MAKE_DIRECTORY("${CMAKE_CURRENT_BINARY_DIR}/grm")
 GETSUBDIRS(GRMGROUPDIR "${PROJECT_SOURCE_DIR}/general-reports/packages/")
 FOREACH(GRMGROUP ${GRMGROUPDIR})
     GETSUBDIRS(GRMDIR "${PROJECT_SOURCE_DIR}/general-reports/packages/${GRMGROUP}/")
@@ -516,8 +517,8 @@ FOREACH(GRMGROUP ${GRMGROUPDIR})
                                 "${PROJECT_SOURCE_DIR}/general-reports/packages/${GRMGROUP}/${GRMNAME}/sqlcontent.sql"
                                 "${PROJECT_SOURCE_DIR}/general-reports/packages/${GRMGROUP}/${GRMNAME}/template.htt"
         )
-        create_zip("${CMAKE_CURRENT_BINARY_DIR}/${GRMGROUP}-${GRMNAME}.grm" "${GRM_DEFAULT_FILES}" "${PROJECT_SOURCE_DIR}/general-reports/packages/${GRMGROUP}/${GRMNAME}/")
-        list(APPEND GRMFILES "${CMAKE_CURRENT_BINARY_DIR}/${GRMGROUP}-${GRMNAME}.grm")
+        create_zip("${CMAKE_CURRENT_BINARY_DIR}/grm/${GRMGROUP}-${GRMNAME}.grm" "${GRM_DEFAULT_FILES}" "${PROJECT_SOURCE_DIR}/general-reports/packages/${GRMGROUP}/${GRMNAME}/")
+        list(APPEND GRMFILES "${CMAKE_CURRENT_BINARY_DIR}/grm/${GRMGROUP}-${GRMNAME}.grm")
     ENDFOREACH()
 ENDFOREACH()
 add_custom_target(generate_grm_files ALL DEPENDS ${GRMFILES})
--- a/po_wxstd/CMakeLists.txt
+++ b/po_wxstd/CMakeLists.txt
@@ -58,18 +58,10 @@ set(WXSTD_MAP_vi_VN "vi")
 set(WXSTD_MAP_zh_CN "zh_CN")
 set(WXSTD_MAP_zh_TW "zh_TW")
 
-execute_process(
-    COMMAND sh "${wxWidgets_CONFIG_EXECUTABLE}" --prefix
-    OUTPUT_VARIABLE wxWidgets_prefix
-    RESULT_VARIABLE wxWidgets_prefix_result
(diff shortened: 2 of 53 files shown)


georgeef:
No, these are the exact labels used by default in wxWidgets. I did a trial, I added greek translations for these labels in po/el_GR.po (as if they were translated in crowdin) and after rebuild I could see the translations in the two windows.

georgeef:
@whalley 
The labels above are already translated in wxWidgets source code, under `locale/`. Any idea how we can reuse the wxWidgets built-in translations?

whalley:
> @whalley The labels above are already translated in wxWidgets source code, under `locale/`. Any idea how we can reuse the wxWidgets built-in translations?

To be honest I thought the wxWidgets built-in translations were in use via this:
https://github.com/moneymanagerex/moneymanagerex/blob/da831d48dbf572ce45ff7afe6823de11fdfa9c68/src/mmex.cpp#L88

...though I'm no expert with the internationalisation elements.

georgeef:
According to wxWidgets manual, `AddStdCatalog()` should load the "wxstd" domain for the selected language, but looking into the debugger it doesn't load anything. It looks for "wxstd.mo" in a predefined search path (set of prefixes), e.g., on macOS it tries to find the "wxstd" english translations in
```
/Volumes/Money\ Manager\ Ex/mmex.app/Contents/Resources/en_US.lproj/wxstd.mo
```

This file does not exist because it is not packaged. If wxWidgets is pre-installed independently of MMEX, probably adding a standard prefix to the search path could solve the problem. However in my setup I did not install wxWidgets, and most users don't. So the only solution would be to package the built-in wxWidgets `.mo` files together with MMEX.

Another solution is this PR: add the wanted messages in a dummy table, in order to be translated again in the MMEX project. It is easy to implement but shifts some effort to translators (no idea if there is a tool to copy translations from wxWidgets files to MMEX crowdin).

In this PR we talk about 5 messages only, so not a big deal.

@ovari 
Do you think it makes sense to re-translate the wxWidgets messages?  Overall there are about 2,000 messages in wxWidgets source code, but I think we use very few of them.

@n-stein as FYI

ghost:
> Do you think it makes sense to re-translate the wxWidgets messages? Overall there are about 2,000 messages in wxWidgets source code, but I think we use very few of them.

Please add them. I'll translate them into Hungarian.

whalley:
>So the only solution would be to package the built-in wxWidgets .mo files together with MMEX.

This is the best option, we should look at building this into the makefile alongside the other .mo files. It does not make sense to duplicate work.

whalley:
> > Do you think it makes sense to re-translate the wxWidgets messages? Overall there are about 2,000 messages in wxWidgets source code, but I think we use very few of them.
> 
> Please add them. I'll translate them into Hungarian.

Don't think this is the right way to go.. See my other comment regarding including the wstd.mo in the build.

georgeef:
I agree, the correct way is to package the wxWidgets translations. I suppose there is no licensing issue.

@whalley 
I keep this PR open (labeled at Work in Progress in title), in order to implement a change in `po/CMakeLists.txt`.

georgeef:
@whalley 
This implementation uses the `wx-config` script to find the wxWidgets translation files. I tested on my development environment on macOS, but it may fail in the app build environment.

Looking into the linux and windows build logs, it didn't find the wxWidgets source root. Any idea how to find it?

@ovari 
~Please check the Hungarian translations in the build executable of this PR.~

whalley:
@georgeef It works great, good work on the CMakeList changes. 

I presume we can remove the constants.cpp/.h changes now.

If any languages are missing I guess people should be pointed at https://wxwidgets.org/about/translations/ to contribute changes that would assist the wider wxWidgets community.

georgeef:
It doesn't work on the github builds. On macOS development environment, the `wx-config --prefix` command returns the source root of wxWidgets, because following the installation instructions we build but do not install wxWidgets. On linux github builds, the same command returns the installation prefix (not the source root), which contains the compiled libraries but not the source .po files. On windows github build I don't understand how wxWidgets is installed.

One solution is to pass the wxWidgets source root as an environment variable. This assumes that wxWidgets sources are available in all development and build environments. It also needs an update to installation instructions.

Another solution is to use the compiled .mo files from the wxWidgets installation directory. This solution is quite complicated. If wxWidgets is not installed, then these files cannot be found, at least not with the same names. The .po files are not available, only the .mo files are installed.

Another solution is to copy the wxWidgets .po files under `po_wxstd`. This makes MMEX build simpler on either a development or a build environment. It requires a periodic update (copy) from the wxWidgets sources, like the periodic updates from crowdin.

georgeef:
We can keep the table in constants.{cpp,h} as a placeholder for the future. The sample entries are commented out.

whalley:
> Another solution is to copy the wxWidgets .po files under `po_wxstd`. This makes MMEX build simpler on either a development or a build environment. It requires a periodic update (copy) from the wxWidgets sources, like the periodic updates from crowdin.

This sounds a pragmatic and workable solution.

georgeef:
I'll make a manual copy. We can update manually every few months until three is a more automated update. The main labels do not change often.

georgeef:
@ovari
Please check the Hungarian translations in the build executable of this PR.

ghost:
> Please check the Hungarian translations in the build executable of this PR.

Sorry, don't know how to download the build executable of this PR for Linux Mint Debian Edition.

Are you able to provide a link to the flatpak executable file?

Thank you

georgeef:
@whalley 
Is the linux build uploaded?  In the CI log I see that upload is skipped. Is there a way to test before merge?

whalley:
> @whalley Is the linux build uploaded? In the CI log I see that upload is skipped. Is there a way to test before merge?

Not quite sure why the builds are only uploaded on merge but I've merged anyway now.

PR #8238, opened 2026-03-14
Update _DataEnum.cpp

ovari:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/8238)


georgeef:
@whalley 
Please reject this PR.

@ovari 
Several strings with `_n` (like those in this PR) are stored in the database or used internally, and dot not appear in the GUI. If you change a single character in them you break database compatibility. Please do not change strings unless you are sure that they are GUI strings only.

In particular, DO NOT change any string in `_DataEnum.cpp`. See also the warning in the beginning of this file. Strings like `SingleChoice` do not appear in GUI and should not be listed with `_n`, because they don't need translation. Some cleanup is needed.

georgeef:
Added a related proposal in #8239.

Issue #8268, opened 2026-03-24
Context menu right‑click causes crash in Currency Manager


habal575:
MMEX version:
 - [ ] 1.7.0
 - [ ] 1.8.0
 - [ ] 1.8.1
 - [ ] 1.8.2
 - [ ] 1.9.0
 - [ ] 1.9.1
 - [ ] 1.9.2
 - [x] 1.9.3-Beta.2 64-bit / Git commit: eaf794b (2026-03-23)

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [x] Windows
 - [ ] Mac OSX
 - [ ] Linux 

# Description of the bug
In the Currency Manager, the context menu crashes when you right-click in the area marked in yellow.

<img width="1142" height="508" alt="Image" src="https://github.com/user-attachments/assets/8307c59f-b157-40d4-a772-c9be3f1ebfef" />

# Reproduction

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:
See description of the bug

Expected result:
Context menu

Actual result:
Debug report "Money Manager EX"

# Additional information

If you have a backtrace for a crash or a warning, paste it here.

renato-mmex:
I tested the latest Beta 2 and didn't encounter any issues.

<img width="240" height="68" alt="Image" src="https://github.com/user-attachments/assets/bc2ffd26-b107-4b11-9217-36a022cbe909" />

KWich:
I can reproduce it and will provide a fix

habal575:
Thank you very much; the context menu functions are working in the version I just installed (70e6cb7, March 24, 2026).

However, I noticed that the “date” and “value” fields are no longer in the correct position. This might be because I tested it too early, in which case I apologize.

<img width="937" height="606" alt="Image" src="https://github.com/user-attachments/assets/f26c09f3-2072-4d0f-a896-a618179edee4" />

KWich:
> However, I noticed that the “date” and “value” fields are no longer in the correct position.

Sorry this is an error I accidentally introduced, it should be fixed now.

habal575:
Thank you very much for your work. The issue has been resolved in version 6a5ea9e (2026-03-24).

<img width="926" height="596" alt="Image" src="https://github.com/user-attachments/assets/52b137e0-cb7d-49fe-bda5-18c8b8e54c7c" />

PR #8270, opened 2026-03-24
Compiler warnings removed and fix for double entry in Nav State

KWich:
This pull request contains these  commits:
- removal of  multiple compiler warnings, displayed under Linux compile.
- Fixes the problem with the navigator tree state, that if a subbranch has the same name as another (sub)branch, the wrong state (collapsed/expanded) is applied: With a refactoring the whole path is stored instead of only the branch name to make the entries unique.
- Fix for crash in  Currency Manager on right mouse click (#8268)

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/8270)

### DIFF
--- a/src/dialog/CurrencyChoiceDialog.cpp
+++ b/src/dialog/CurrencyChoiceDialog.cpp
@@ -269,7 +269,7 @@ void CurrencyChoiceDialog::CreateControls()
     wxStaticBoxSizer* historyStaticBox_Sizer = new wxStaticBoxSizer(w_hostory_box, wxVERTICAL);
     rightBoxSizer->Add(historyStaticBox_Sizer, g_flagsExpand);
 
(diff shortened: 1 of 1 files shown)


georgeef:
@KWich 
The changes in `mmGUIFrame::~mmGUIFrame()` create a crash at startup -> Exit. Please fix.

KWich:
> @KWich The changes in `mmGUIFrame::~mmGUIFrame()` create a crash at startup -> Exit. Please fix.

Do you have any details? What kind of crash, OS, Trace/Log?  Thanks

georgeef:
The `mmGUIFrame` destructor calls `navTreeStateToJson()`, which saves into `InfoModel`. If you click Exit on the startup window, `m_db` is NULL and this save crashes the application. You need to add a check for NULL somewhere.

After the fix check that there is no other crash (test the exit on startup windows, or after startup and change of language).

Issue #8276, opened 2026-03-30
Crash on startup with existing databases in 1.9.3.0 on Windows


frankieorabona:
MMEX version:
 - [ ] 1.7.0
 - [ ] 1.8.0
 - [ ] 1.8.1
 - [ ] 1.8.2
 - [ ] 1.9.0
 - [ ] 1.9.1
 - [ ] 1.9.2
 - [ x ] Other (please specify) 1.9.3 BETA 2   (downloaded by appveyor 4720)

Operating System:
 - [x] Windows
 - [ ] Mac OSX
 - [ ] Linux 

# Description of the bug

MMEX crashes on startup when opening existing/real databases.

The issue does not happen with new test databases, but it does happen with multiple existing databases, including an older real database from 2023.

I collected multiple crash reports and they all point to the same failure pattern:
ACCESS_VIOLATION (0xc0000005), then wxString::wxString / operator+, then wxDataSchemeFSHandler::OpenFile.

In one trace the UI path also includes wxTreeCtrl::MSWOnNotify, while in other traces it comes from wxTimerImpl::SendEvent, but the crash lands in the same code path.

This suggests the problem is not random corruption of a single DB file, but a reproducible startup/rendering regression triggered by existing databases.

# Reproduction

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:
1. Start MMEX 1.9.3.0 on Windows 11
2. Open an existing/real database
3. MMEX crashes during startup

Expected result:
The database should open normally.

Actual result:
MMEX crashes during startup.

# Additional information

- New empty/test databases do not crash
- Existing real databases do crash
- Also reproducible with an older real database from 2023
- Crash reports attached

Backtrace summary:
ACCESS_VIOLATION (0xc0000005)
wxString::wxString
operator+
wxDataSchemeFSHandler::OpenFile

In another trace:
wxTreeCtrl::MSWOnNotify
wxDataSchemeFSHandler::OpenFile
wxString::wxString
operator+

If needed, I can also provide one of the affected database files privately for testing, since I prefer not to attach it publicly here.

frankieorabona:
Some debug report:

[MoneyManagerEx_dbgrpt-33532-20260330T211841.zip](https://github.com/user-attachments/files/26359780/MoneyManagerEx_dbgrpt-33532-20260330T211841.zip)

[...]

PR #8281, opened 2026-03-31
fix(#8276): temporary fix for corrupted CATEGORY_V1

georgeef:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/8281)


georgeef:
As discussed in #8276, this is not a proper fix.

Issue #8283, opened 2026-04-02
Inserting a tag into an existing schedule transaction causes an error message


habal575:
MMEX version:
 - [ ] 1.7.0
 - [ ] 1.8.0
 - [ ] 1.8.1
 - [ ] 1.8.2
 - [ ] 1.9.0
 - [ ] 1.9.1
 - [ ] 1.9.2
 - [ x ] 1.9.3 / Git commit: c6d3c12 (2026-03-31)

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [x] Windows
 - [ ] Mac OSX
 - [ ] Linux 

# Description of the bug

Please describe your issue with details.
Inserting a tag into an existing scheduled transaction causes the system an error message.

<img width="834" height="310" alt="Image" src="https://github.com/user-attachments/assets/3216ac9f-2d3f-4d40-a4c5-67c326cceab4" />

When I create a new scheduled transaction with a tag, everything works fine.

# Reproduction
Inserting a tag into an existing scheduled transaction

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:
1.) Edit an existing scheduled transaction
<img width="1400" height="907" alt="Image" src="https://github.com/user-attachments/assets/324ad221-439f-431f-bc44-63d8a450b307" />

2.) Add an tag and save
<img width="1398" height="910" alt="Image" src="https://github.com/user-attachments/assets/e565e114-9b88-4f27-a91a-5701eccc2ac7" />

Expected result:
Add the tag to the scheduled transaction

Actual result:
3.) The result is the error message shown here; sometimes no error message appears, but the tag is not saved in the record.
<img width="834" height="310" alt="Image" src="https://github.com/user-attachments/assets/5ded5f75-06d6-4c52-9ed8-ac4bf65edeea" />

# Additional information

If you have a backtrace for a crash or a warning, paste it here.

georgeef:
Thank you for the bug report. Should be fixed in #8286

habal575:
Thank you very much for your excellent work.

With the version listed below, everything works perfectly again—both the scheduled transaction and the scheduled split transaction.

• Version: 1.9.3-Beta.2 64-bit
• Built on: Apr 2 2026 11:44:52
• Database version: 21 (aes256cbc-hmac-sha512)
• Git commit: 375999d (2026-04-02)

PR #8286, opened 2026-04-02
fix(#8283): use key() instead of name() in model

georgeef:
Fix bug introduced in #8277.

The name of several choices In `data/_DataEnum` has been duplicated without translation, for use by model functions. Currently the following choices have different key (not translated) and name (translated):
* FieldTypeN
* RefTypeN

This PR replaces `name()` with `key()` in several model functions.

### Impact
The bug reported in #8283 is due to `RefTypeN` -> `e_sched`. There are more entries which create similar bugs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/8286)


georgeef:
@whalley 
This is a critical fix. Please prioritize.

PR #8302, opened 2026-04-09
Stock Panel updates

KWich:
This pull request contains these updates:
- Added a name filter to the panel to select one or more stocks by name
- Colors a loss in red 
- Stores the Selection value for "All" or "Non-Zero values" in the settings
- Automatic adjustment of the column sizes when the window is resized. 

StockPanelChanges

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/8302)

### DIFF
--- a/src/panel/StockList.cpp
+++ b/src/panel/StockList.cpp
@@ -511,21 +511,38 @@ int StockList::initVirtualListControl(int64 trx_id)
 {
     /* Clear all the records */
     DeleteAllItems();
-
-    // TODO
-    if (w_panel->m_account_id > -1 ) {
-        m_stock_a = StockModel::instance().find(
-            StockCol::HELDAT(w_panel->m_account_id),
-            StockCol::NUMSHARES(w_panel->getFilter() ? OP_GT : OP_GE, 0.0)
-        );
+    if (w_panel->m_name_filter_value.IsEmpty()) {
+        // TODO
+        if (w_panel->m_account_id > -1 ) {
+            m_stock_a = StockModel::instance().find(
+                StockCol::HELDAT(w_panel->m_account_id),
+                StockCol::NUMSHARES(w_panel->getFilter() ? OP_GT : OP_GE, 0.0)
+            );
+        }
+        else { // create summary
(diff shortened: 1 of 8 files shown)


georgeef:
@KWich 
This PR added a new case `OP_LK` in struct `OP` in `src/table/_TableClause.h`, which is very nice to have (I also wanted to add it). Please check for impact and consistency with the rest of the code. E.g, how it behaves when used with `TrxModel::WHERE_DATE`.

You need to search for variables or function arguments of type `OP`, where the value can be any `OP` case, and assess if the new case `OP_LK` works as expected. You may need to fix inconsistencies or add a comment where `OP_LK` is not allowed (no need for formal check and debug error). Probably everything is fine, but you need to check thoroughly for consistency (if you did already just confirm). 

@whalley 
Changes in `base`, `data`, `table` (these are the lowest-level layers in `src/_readme.txt`) may create inconsistencies and new bugs, because they are used in multiple places. I propose that changes in these layers are (1) submitted separately and not as part of new features, (2) reviewed before they are merged. The author of the changes should somehow confirm that there was an impact and consistency check. The argument that a new addition does not affect the existing code is not sufficient. E.g., programmers in the future will see that `OP_LK` is one of the available op codes and will use it without any restriction (because no restrictions are documented anywhere). This addition may work fine for the new feature for which it was introduced, but may break in other contexts.

The next layer (`model`, `pref`, `util`) is also used in multiple places in GUI. I propose to have a review (or confirmation from the author that consistency has been checked) before merge also for this layer, but changes can be submitted together with features in GUI.

In the last weeks I did many changes and re-organization of the lower layers, and I can only tell that it is a lot of effort to see the whole picture (search for each and every instance where a low-level structure is used) and to keep the whole code consistent. And it is a mess if low-level changes are done without checking for consistency. I am not a fan of "bug driven design" (add something quickly, and fix the rest of the code if and when a bug is reported).

Th e`OP_LK` introduced in this PR seems to be fine, but it is a good example of where a thorough check is needed in order to prevent future bugs.

Issue #7137, opened 2025-01-18
More options for Summary of Accounts report


JoshuaLampert:
# Description of the feature

I really like the "Summary of Accounts" report, but I would like to have more customizable options. Especially, I would like to see

- [ ]  A weekly overview in addition to the monthly and yearly overview. Ideally, there would even be an "every N days" overview, where N is customizable.
- [ ] An option to set the time period, as, e.g., in the Income vs. Expenses report.
- [ ] An option to filter the accounts, as, e.g., in the Income vs. Expenses report.
- [ ] An option to show different graphs for different accounts in one diagram.

# Use cases

Having more options to filter the report, allows for a better overview of the development of the account balance.

georgeef:
Range controls will be revised in a next version (maybe 1.9.1), first for transaction panels and then for reports.

For transaction panels, the current plan is for both ends of the range to have a period type (Year, Quarter, Month, Week, Day) and a numeric value. E.g., it will be possible to specify a range from the beginning of the second previous month to the end of next week. Future (scheduled) transactions in panels have been implemented in the upcoming version (1.9.0).

For reports, the range will be like in panels and the reporting periods will either have the same type as the end of the range, or it will be specified separately. E.g., for the range above, the reporting periods will be the end of each week. Like in panels, future periods will include scheduled transactions.

JoshuaLampert:
Not stale.

PR #8320, opened 2026-04-15
Add selection between accounts and account types in summary of accounts report

JoshuaLampert:
This essentially implements the fourth check box in #7137 by adding a dropdown selection menu to the summary of accounts report, which allows to switch between displaying _account types_ (current version) or the _accounts_ (new) as the columns in the table and bars in the bar chart.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/8320)

### DIFF
--- a/src/report/BalanceReport.cpp
+++ b/src/report/BalanceReport.cpp
@@ -38,7 +38,7 @@ BalanceReport::BalanceReport(BalanceReport::PERIOD_ID period_id) :
         : REPORT_ID::YearlySummaryofAccounts
     );
     m_selection_map["name"] = _t("View");
-    m_selection_map["values"] = _t("Account Types,Accounts");
+    m_selection_map["values"] = _t("Account Types") + "," + _t("Accounts");
     m_selection_map["default"] = _t("Account Types");
 }
 
--- a/src/report/BalanceReport.cpp
+++ b/src/report/BalanceReport.cpp
@@ -37,6 +37,9 @@ BalanceReport::BalanceReport(BalanceReport::PERIOD_ID period_id) :
         ? REPORT_ID::MonthlySummaryofAccounts
         : REPORT_ID::YearlySummaryofAccounts
     );
+    m_selection_map["name"] = _t("View");
+    m_selection_map["values"] = _t("Account Types,Accounts");
+    m_selection_map["default"] = _t("Account Types");
 }
 
 std::map<mmDate, double> BalanceReport::loadAccountBalance_mDate(const AccountData& account_d)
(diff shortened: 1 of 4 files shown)


src/report/BalanceReport.cpp
```diff
@@ -37,6 +37,9 @@ BalanceReport::BalanceReport(BalanceReport::PERIOD_ID period_id) :
         ? REPORT_ID::MonthlySummaryofAccounts
         : REPORT_ID::YearlySummaryofAccounts
     );
+    m_selection_map["name"] = _t("View");
+    m_selection_map["values"] = _t("Account Types,Accounts");
```
whalley: @JoshuaLampert I would suggest building this up from "Account Types" and "Accounts" rather than a combined string as it will lead to clearer translations and ensure that "Account Types" is only needs a single translation.

whalley: @JoshuaLampert I would suggest that when "Accounts" is selected that the global setting of "Accounts Visible" is adhered to, e.g don't show "Closed" accounts if "Accounts Visible" = "Open".

JoshuaLampert: > @JoshuaLampert I would suggest building this up from "Account Types" and "Accounts" rather than a combined string as it will lead to clearer translations and ensure that "Account Types" is only needs a single translation.

Thanks for the suggestion @whalley. I agree and did that in [ca689e9](https://github.com/moneymanagerex/moneymanagerex/pull/8320/commits/ca689e93b48df0039fe8f30a3ad919d22998cb03).

> @JoshuaLampert I would suggest that when "Accounts" is selected that the global setting of "Accounts Visible" is adhered to, e.g don't show "Closed" accounts if "Accounts Visible" = "Open".

We could do that, but I am not sure if it leads to more confusion because:
1. There already is an accounts filter in the summary of accounts report to select specific accounts. What should be shown if a closed account is selected with the "Specific Accounts..." selection option? Should the global "Accounts Visible" selection be respected (such that the account is not shown) or should it still be shown because it was selected? It is unclear to me and both solution seem like they can lead to confusion.
2. As far as I can see, the same applies to the Income vs. Expenses report, where I think this setting is also not respected. In my opinion it would be inconsistent to only have it either of the report.
3. Even if I have the "Accounts Visible" option set to "Open" to hide them in the list of accounts in the navigator, I would usually still want to include them in the summary of accounts report. Consider the case where I had some money on an account, which I transfer to another (still open) account before a date `x` and then close the previous account at date `x`. If closed accounts are not shown in the summary of accounts report (say if all accounts are selected in the accounts selection menu) I would see a lower balance before that date `x` and an unnatural increase around date `x`. I would find that confusing.

whalley: Understood, good points.

PR #8323, opened 2026-04-16
refine data and model (part 10)

georgeef:
### Changes
* switch to new database API; remove older API
* add virtual `TableFactory::find_id_isUsed`
* refine virtual `TableFactory::purge_id`
* refine various `purge_*` methods
* refactor `mmAttachmentManage` -> `util/mmAttachment`

### Notes

In general the new database API is a superset of the older API. There are only minor differences, described below.

The older `find_all()` adds "COLLATE NOCASE" after "ORDER BY ${COLUMN}", while the new `find_data_a()` and `find_id_a()` do not; however, this is the default SQLite behavior if "${COLUMN}" is already defined with "COLLATE NOCASE".

The older `find()` and `find_all()` return an immutable `const DataA`, while the new `find_data_a()` returns a mutable `DataA`. The reason for this change is that quite often the application code needs to overwrite and update the returned records in `DataA`.

The older `struct ${COLUMN} : TableOpV` provides a constructor of the form `${COLUMN}(op, value)` and a simplified constructor of the form `${COLUMN}(value)` where `op=OP_EQ` is implied. The new API provides methods of the form `WHERE_${COLUMN}(op, value)` but does not provide a simplified form `WHERE_${COLUMN}(value)` where `op=OP_EQ` is implied. The presense of `OP_EQ`, as well as the prefix `WHERE_`, make the application code a little more verbose (and a little more readable). The simplified form `WHERE_${COLUMN}(value)` could be added in `util/sqlite2cpp.py` in the future, if it is desirable.

The new API gives access to SQLite sorting (ORDER BY), i.e., the results are sorted as part of `find_data_a()`, and no additional sorting is needed in most cases. However, sorting of transactions (`CHECKINGACCOUNT_V1`) by date-only (the 10-character prefix of `TRANSDATE`) is still inefficient, and it is better done outside of `find_data_a()`. The SQLite engine can sort by date-only with
```
ORDER BY substr(TRANSDATE, 1, 10)
```
however this is slow, unsless there is an index for this prefix, like
```
CREATE INDEX idx_date ON CHECKINGACCOUNT_V1(substr(TRANSDATE, 1, 10))
```
Instead of adding prefix indexes for date, I propose that in the next schema change, the date and time fields of `TRANSDATE` are stored in two separate fields. In general application code becomes simpler and more robust if the database schema becomes simpler and clean.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/8323)


georgeef:
@whalley 
~The windows build failed at the uploading phase. Could you check?~
It is ok now.

PR #8345, opened 2026-05-02
Fixes for template with C++20 and wxUSE_STD_CONTAINERS can now be false

gerstrong:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/8345)


general-reports
whalley, line 1: Think you may have an older version of the general-reports

Issue #8364, opened 2026-05-19
Create a scheduled transaction from an existing transaction (category and tags are missing)


habal575:
MMEX version:
 - [ ] 1.7.0
 - [ ] 1.8.0
 - [ ] 1.8.1
 - [ ] 1.8.2
 - [ ] 1.9.0
 - [ ] 1.9.1
 - [ ] 1.9.2
 - [x] 1.9.3 Git commit: 6ee3a11 (2026-05-17)

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [x] Windows
 - [ ] Mac OSX
 - [ ] Linux 

# Description of the bug

When I create a scheduled transaction from an existing transaction (category and tags are missing)

<img width="1997" height="836" alt="Image" src="https://github.com/user-attachments/assets/af33b61a-b2fd-421f-9e99-347f5bab8aef" />

<img width="1400" height="911" alt="Image" src="https://github.com/user-attachments/assets/09a16df4-e64c-48a1-a4e0-c3484117d0ee" />

# Reproduction

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

Reproduction steps:
Create a scheduled transaction from an existing transaction

Expected result:
That the category and tags are also copied into the new transaction

Actual result:
<img width="1400" height="911" alt="Image" src="https://github.com/user-attachments/assets/6e83854d-f513-412f-b8ad-4f343f91c4c5" />

# Additional information

• Version: 1.9.3-Beta.2 64-bit
• Built on: May 17 2026 12:27:57
• Database version: 21 (aes256cbc-hmac-sha512)
• Git commit: 6ee3a11 (2026-05-17)
MMEX is using the following support products:
• wxWidgets 3.3.2 (wxMSW 10.0)
• wxSQLite3 4.9.1 (SQLite 3.40.0)
• RapidJSON 1.1.0
• Lua 5.3.5
• libcurl/8.5.0-DEV Schannel
• gettext 0.22.5
• apexcharts.js

If you have a backtrace for a crash or a warning, paste it here.

whalley:
I can see that the Tag is not copied. For the category it seems to us the Default category logic as set in the Transaction settings, which seems wrong.

habal575:
Thank you for looking into this issue.

[...]

PR #8371, opened 2026-06-02
Multiple changes and fixes

KWich:
This pull request contains these enhancements/fixes:

- Prevent crash during print due to invalid pointer (reported in #8309):
  Note: This only removes the crash but does not fixes the printing issues.
- An implementation for #8364, which ensures that tags, category and color information are correctly copied when a scheduled transaction from an existing transaction is created 
- SchedDialog: correct sizer layout to remove warnings about wrong allocation, ensure that the dialog is centered on MMEX window and not on screen like all other dialogs.
- TrxDialog: If a split transaction is displayed ensure that in the split dialog the tags are displayed (regression introduced due to redesign)
- Journal panel: Added missing implementation to display the split tags when "View Split Categories" is called.
- Journal panel: Align right click behavior with other panels - if a right click is performed on an un-selected  transaction, it is selected before showing the context menu.
- Journal panel: Disable the whole submenu "Mark as" instead of its entries if no transaction is selected.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/8371)

### DIFF
--- a/src/app/mmFrame.cpp
+++ b/src/app/mmFrame.cpp
@@ -3544,7 +3544,12 @@ void mmFrame::OnAbout(wxCommandEvent& WXUNUSED(event))
 
 void mmFrame::OnPrintPage(wxCommandEvent& WXUNUSED(event))
 {
-    panelCurrent_->printPage();
+    if (panelCurrent_) {
+        panelCurrent_->printPage();
+    }
+    else {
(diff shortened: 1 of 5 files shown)


KWich:
Is the problem that the Ubuntu CI Build can not install the wxWidgets known?

Issue #8311, opened 2026-04-12
Column width resizing is not saved


ramshake:
MMEX version:
 - [ ] 1.7.0
 - [ ] 1.8.0
 - [ ] 1.8.1
 - [ ] 1.8.2
 - [ ] 1.9.0
 - [ ] 1.9.1
 - [x] 1.9.2
 - [x ] Other 1.9.3 beta 2 build 4794

Note: bug reporters are expected to have verified the bug still exists
either in the last stable version of MMEX or on updated development code
(master branch).

Operating System:
 - [x] Windows
 - [ ] Mac OSX
 - [ ] Linux 

# Description of the bug

I updated to last build from v1.9.2 because i saw Last N days report was fixed in 1.9.3. But, The resizing of columns in transactions is not saved. It is getting reset to default view after I open the application again. 

# Reproduction

Columns are reset to default after upgrade. I resized some columns and hid 2 columns. I colsed the app and started again. The columns are reset to default.

Is the bug reproducible? 
 - [x] Always 
 - [ ] Randomly 
 - [ ] Happened only once

grumpy235:
I also have this problem on Windows using locally compiled master  at commit 12372ee98
I deleted my current mmexini from C:\Users\xxxx\AppData\Roaming so as to force a refresh and opened my existing database.
All columns are displayed as expected with a rogue vertical line in number as follows...

<img width="639" height="161" alt="Image" src="https://github.com/user-attachments/assets/2fdb21a5-17f4-4baa-818b-ab5448b3fc1c" />

This line is also seen in Notes column

I hide columns that I do not require (SN, ID, Tag) - display looks fine apart from the odd spurious vertical line.
Close/restart the application and get the following..

<img width="430" height="162" alt="Image" src="https://github.com/user-attachments/assets/530f45e8-dd83-4809-a1df-1e0c5775e374" />

Have now got a vertical line in Date and Number and the Payee column and subsequent columns are now extremely wide

grumpy235:
Just to ensure what I am seeing has nothing to do with my compile I downloaded portable Appveyor build with an empty mmexini.db3

• Version: 1.9.3-Beta.2 64-bit
• Built on: Apr 17 2026 02:46:50
• Database version: 21 (aes256cbc-hmac-sha512)
• Git commit: 07a3768 (2026-04-17)


 Opened a very small test database and displayed the only Account in it and this gave...
 
[...]

PR #8397, opened 2026-07-06
fix(#8311): Resize option added & wx3.3.3 & v1.9.3RC2

whalley:

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/8397)


src/pref/ViewPref.cpp
KWich, line 1: _t("Resize columns to fit transaction panel")):  
Better name it to  _t("Automatic resize columns to fit panel"))  , as it affects not only the  transaction view bit also the stock view.
The option is also better show as general view option and not transaction related. 

PO File entry is needed !

whalley: > _t("Resize columns to fit transaction panel")): Better name it to _t("Automatic resize columns to fit panel")) , as it affects not only the transaction view bit also the stock view. The option is also better show as general view option and not transaction related.
> 
> PO File entry is needed !

Have renamed and moved, the PO will get automatically generated when merged.

whalley:
WxWidgets packaging issue... https://github.com/wxWidgets/wxWidgets/issues/26664

PR #8413, opened 2026-07-14
Column handling fixes and refactor

KWich:
This pull request makes these changes:

Fixes:
- Fix segmentation fault if no resizable columns are displayed
- Fix to prevent that last column can be hidden ( If the last column was hidden, a reset in the database is necessary to show the columns again.)

Refactor:
- Refactor automatic column handling to generic solution in ListBase, which includes scheduled transactions, Budgetlist, Asset, ...
- Refactor Journallist & _Listbase: Unnecessary disabled column handling removed

Enhancements:
- Automatically resize list after column is hidden or shown 
- Set default tor automatically resize to true

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/8413)


src/model/PrefModel.cpp
```diff
@@ -541,7 +541,7 @@ void PrefModel::saveDoSpecialColorReconciled(const bool value)
 
 void PrefModel::loadDoPanelResize()
 {
-    m_do_panel_resize = SettingModel::instance().getBool("RESIZE_TRANSACTION_PANEL", false);
```
whalley, line 544: This was set to false to mimic existing behaviour. That way users are not 'surprised' and can continue to set the widths themselves without automatic resizing.

KWich: It is better if new feature which improves the behavior is per default activated, instead of having to search through a list of settings (which are not documented). 
From my understanding in this case the settings switch is only intended as a fall back.

whalley: There was discussion here: https://github.com/moneymanagerex/moneymanagerex/issues/8311#issuecomment-4885104675 I'm happy either way, I actually prefer the auto-resize but I can see it annoyimg/confusing for those who have set up the sizing themselves to match screen sizes, etc.

KWich: Ok, I missed the point that the default value was already discussed, I changed it back and updated the docu.

PR #8453, opened 2026-08-08
Financial planner: segmented budgets, derived amounts and long-term plan

golish:
## What this adds

Structured financial planning on top of the existing budget, scheduled-transaction and stock models, for the case where a plan currently lives in a spreadsheet or a notebook: a month planned against paycheck-aligned halves, amounts that are derived rather than retyped, and a multi-year view of what is expected versus what is already spoken for.

It is additive. Database version goes 21 → 22; existing budgets are untouched.

### Budget segments (`BUDGETSEGMENT_V1`)

A budget period can be split into intra-period windows, so a month can be planned as two paycheck halves instead of one lump sum. `BUDGETTABLE_V1.SEGMENTID` is nullable — existing entries keep applying to the whole period and are unaffected by the upgrade.

### Derived and estimated amounts

`BUDGETTABLE_V1` gains `AMOUNTTYPE` (Fixed/Estimated/Auto) and `AUTOSOURCE`. An Auto entry is computed instead of typed:

| Source | Meaning |
|---|---|
| `CCBALANCE:<account>` | what a credit card currently owes |
| `AVG:<n>M` | trailing n-month mean of actual amounts |
| `LAST` | most recent actual |
| `SCHED:<id>` | amount of a scheduled transaction |

This removes the placeholder amounts that otherwise have to be looked up by hand each period. An unresolvable source is flagged, never silently `0`.

### Long-term plan (`PLANGROUP_V1`, `PLANITEM_V1`)

Plan items are grouped into projects (a trip, a collection) and carry a status. Only Planned/Committed count as obligations; Wishlist entries are tracked without reducing free assets. Unit-based items model share vesting as units × price, less tax.

### Assumptions (`PLANASSUMPTION_V1`)

A long-term plan rests on guesses — an assumed share price, a tax rate. These are stored once and referenced by plan items, so changing an assumption updates every dependent figure. Resolution order is assumption → per-item override → live market price. Reports mark assumption-derived values and show how far the outcome moves if the guesses are wrong. Deleting an assumption freezes the value onto its dependants rather than silently zeroing them.

### `PlanEngine`

Merges opening balances, budget entries expanded per segment, scheduled transactions and plan items into a projected timeline with a running balance, and computes free assets as `liquid assets + expected income − committed obligations`. It has no UI dependencies.

### Carrying across periods

A budget period is an accounting convenience, not something the money respects. Bills cluster around due dates, so a month that is net positive can be deeply negative in its first half — and for anyone paid weekly it is worse. Two things follow from that.

The timeline already threaded a running balance and never showed it. Each period now reports the balance it opens with as well as the one it closes on, and the report names the low point of the whole horizon: a plan that ends well can still run out of money in the middle, and that trough is the number worth acting on.

Beyond cash, `BUDGETTABLE_V1.ROLLOVER` lets a category carry rather than reset. `Surplus` turns an entry into a sinking fund — the contribution still leaves the balance, but it accumulates, and a plan item in the same category is met from that fund before it touches the period's own cash, so an annual bill saved for over twelve months stops landing as a shock. A fund never covers more than it holds; anything beyond it is a real cost in the period it falls. Existing rows are backfilled to `None` and behave exactly as before.

There is deliberately no deficit-carry option, though it is the obvious companion. Carrying a deficit is a budget-versus-actual idea — you overspent an envelope, so the next one is smaller — and a forward projection has no actuals to overspend against. Absorbing an unfunded bill into the fund drops a payment that genuinely leaves the account out of the cash flow; keeping the cash honest instead counts the repaying contributions twice. Rather than ship a setting that quietly understates a bill, it is left out.

### UI and reports

`BudgetSegmentDialog`, `PlanAssumptionDialog` and `PlanManagerDialog`, reached from Tools. Two reports (Long-Term Plan, Planned Cash Flow) are registered in the navigation tree, gated on the new tables being populated.

## Notes for reviewers

- **The `database` submodule pointer is deliberately left at upstream.** The v22 schema belongs in `moneymanagerex/database` and is up for review as [moneymanagerex/database#61](https://github.com/moneymanagerex/database/pull/61). Pointing the parent at a commit that exists only on a fork would make this branch impossible to check out recursively, since `.gitmodules` names the upstream repository. The generated sources in `src/table` carry the schema, so nothing here needs the pointer moved; it should be bumped in one step once #61 lands. **Note that #61 is a hard prerequisite for Windows**, because AppVeyor regenerates `src/table/*` from the submodule during `before_build` — Linux and macOS compile the committed sources and so are unaffected.
- **A sign fix worth flagging.** `PlanEngine::build_timeline` treated a positive budget amount as an expense. MMEX stores the amount signed by direction — `BudgetEntryDialog` negates an expense on save and `BudgetPanel` reads the sign back the same way — so the projection had income and expenses swapped throughout. Anyone comparing against an earlier build of this branch will see different, and correct, figures.
- **One drive-by fix:** `mmHTMLBuilder::addTableRowBold` was declared in `htmlbuilder.h` but never defined. It had no callers, so it never failed to link until now. Implemented alongside `addTableRow`. Happy to split this into its own PR if preferred.
- The new `table/`, `data/` and `_DataEnum` code follows the existing generated-source conventions, and `util/sqlite2cpp.py` is updated with the new table-to-class mappings.

## Testing

- `Ubuntu CI Build` passes on both `ubuntu-22.04` and `ubuntu-24.04`; a macOS build (universal, statically linked wx) also passes on a fork branch.
- The 21 → 22 migration was run against a real production database **through MMEX's own statement-splitting path**, not by executing the script in one go — the two behave differently, which matters because `dbUpgrade::SplitQueries` splits on `;` and only discards single-line comment-only fragments. All existing budget rows preserved, new tables and indexes created, `AMOUNTTYPE` and `ROLLOVER` backfilled with no NULLs, and every existing entry defaulting to whole-period.
- The vesting/free-asset arithmetic was checked against a hand-computed plan and matches.

Feedback very welcome on the schema shape in particular, since that is the part that is expensive to change later.

- - -
This change is [Reviewable](https://reviewable.io/reviews/moneymanagerex/moneymanagerex/8453)


golish:
A note on the failing Codacy check, since it is entirely mechanical:

All 22 findings are the same cppcheck rule, `useInitializationList`, and every one of them is in a **generated** file (`BudgetSegmentTable.cpp`, `PlanGroupTable.cpp`, `PlanItemTable.cpp`, `PlanAssumptionTable.cpp`, `BudgetTable.cpp`) on the `*Row` default constructor.

That is the existing output shape of `util/sqlite2cpp.py`. For comparison, the generated `BudgetPeriodRow` already on master is identical in form:

```cpp
BudgetPeriodRow::BudgetPeriodRow()
{
    BUDGETYEARID = -1;
}
```

Codacy only flags the new tables because the files are new, not because they differ from the convention. I have deliberately not hand-edited them: they are generator output, and editing them would make the checked-in files diverge from what `sqlite2cpp.py` produces on the next regeneration.

If you would like this silenced, the fix belongs in the generator template so that every table benefits at once. Happy to do that as a separate PR if you want it — it would touch every generated table file, so it seemed wrong to bundle into this one.

golish:
The AppVeyor failure here is a cross-repository ordering problem rather than a defect in the C++ — flagging it since the errors look alarming out of context.

AppVeyor regenerates the table sources from the `database` submodule before building:

```
before_build:
- cd src\table &&
  python ..\..\util\sqlite2cpp.py ..\..\database\tables.sql &&
  python ..\..\util\sqliteupgrade2cpp.py ..\..\database\incremental_upgrade
```

Because the submodule is still pinned at v21, that step **overwrites** the generated v22 sources committed here with v21 ones, stripping `SEGMENTID` / `AMOUNTTYPE` / `AUTOSOURCE` and the new tables back out again. The build then fails exactly where you would expect:

```
src\data\BudgetData.cpp(45,9): error C2039: 'SEGMENTID': is not a member of 'BudgetRow'
src\model\BudgetSegmentModel.cpp(108,20): error C2039: 'WHERE_SEGMENTID': is not a member of 'BudgetCol'
```

The Ubuntu GitHub Actions workflow does not regenerate, which is why it passes and AppVeyor does not.

So the schema genuinely has to land in the `database` repository first. I have opened that PR: **moneymanagerex/database#61**.

I verified the two are consistent by regenerating rather than by inspection: running both generators against that branch reproduces the generated sources committed here **exactly** — zero differences across `BudgetSegmentTable`, `PlanGroupTable`, `PlanItemTable`, `PlanAssumptionTable`, `BudgetTable`, `tables_en.sql` and `_TableUpgrade.h`, ignoring only the `AUTO GENERATED at` timestamp line.

Once #61 is merged I will push the submodule pointer update here and AppVeyor should go green. Happy to reorder or restructure if you would prefer this split differently.

golish:
Force-pushed one fix after testing the migration on a real database.

Opening a v21 file with a build of this branch failed with `Statement not accessible` and rolled the upgrade back. The schema was fine — the problem was a semicolon inside a `--` comment in the v22 upgrade script.

`dbUpgrade::SplitQueries` splits on `;` and only discards *single-line* comment-only tokens, so splitting mid-comment left a multi-line comment-only token that reached `PrepareStatement` and produced a null statement. The tail of the comment then corrupted the following `CREATE TABLE`.

The only change here is the regenerated `src/table/_TableUpgrade.h` — one comment line. The matching source change is in moneymanagerex/database#61, where I have written up the mechanism in more detail.

Verified by reproducing MMEX's own execution path rather than using `executescript`: build a v21 database, split on `;`, apply the same comment filter, then prepare and execute each token individually. The upgrade now applies all 14 statements, preserves existing budget rows, creates the four new tables, adds the three `BUDGETTABLE_V1` columns, backfills `AMOUNTTYPE` with no NULLs, leaves `SEGMENTID` NULL so existing entries continue to apply to the whole period, and ends at `user_version = 22`.

Ubuntu CI and a macOS build are both green on the updated branch. AppVeyor still needs #61 to merge first, for the reason described earlier in this thread.

golish:
Pushed a further round of work on this branch, driven by testing the build. Summarising since the diff has grown.

**Assumption groups.** A single assumption answers a question with one number, which makes it awkward to say "MSFT might be 350, 400 or 450" and move the plan between those cases. A group is a typed, scoped set of alternatives with one member active; plan items reference the group, so changing the active member in one place moves every dependent figure. Typing the group is what makes the link safe: a share-price group scoped to a ticker only offers itself to rows about that ticker, so a tax rate can no longer be selected where a price is meant. The resolver enforces the same rule, so a link that stops applying is ignored rather than silently used.

**A correctness fix worth calling out.** Items with no target date were excluded from every total, so an undated obligation did not reduce free assets. Once a target date became optional that was a real error, not a presentation issue. `find_between_a` can now include them and the summary and sensitivity calculations do.

**Presentation.** The plan report rendered nested groups flattened, so a project's structure was lost, and expenses in no group were invisible while still counted in the totals below - which made the figures look wrong. Both are fixed, with a subtotal per level. Amounts, values and units now go through currency/percent formatting rather than being printed raw.

**Reach.** Budget segments appear under their period in the navigation tree and are reachable from the Budget Editor where periods are created, rather than only from a separate Tools entry. A transaction or scheduled transaction can be added to the budget from its context menu; a scheduled one can be linked via `SCHED:<id>` so the budget follows it instead of copying a number that will drift.

The schema PR (moneymanagerex/database#61) has been updated in step and still needs to land first, for the `before_build` regeneration reason described earlier in this thread. I amended v22 rather than adding a v23 since it is unmerged.

Both Ubuntu and a macOS build are green on this branch. Happy to split any of this out if it would be easier to review as separate PRs - the assumption-group work in particular stands alone.

whalley:
@golish I'm looking to get a v1.9.4 release out next week to mop up some bugs found after releasing v1.9.3. Will then look to review and integrate your changes as part of v1.9.5

golish:
> @golish I'm looking to get a v1.9.4 release out next week to mop up some bugs found after releasing v1.9.3. Will then look to review and integrate your changes as part of v1.9.5

Sounds good, let me know if you need anything from me!

